PR #103, opened 2021-03-17
Integrated Kuberarmor with CICD

prithivkrishna:


Dockerfile.kubearmor
```diff
@@ -0,0 +1,31 @@
+### Builder
```
nyrahul, line 1: There already exists this file in the repo in `KubeArmor/build` folder. Better to create a symbolic link or best would be to directly use that file instead of creating new one.

helm/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
nyrahul, line 1: Is it possible to use the existing YAML defined without creating a new one?

helm/templates/crd.yaml
```diff
@@ -0,0 +1,326 @@
+apiVersion: apiextensions.k8s.io/v1
```
nyrahul, line 1: This file too is a duplicate of an existing CRD YAML and must not be copied. This will surely cause problems in the near future.

helm/templates/daemonset.yaml
```diff
@@ -0,0 +1,191 @@
+apiVersion: apps/v1
```
nyrahul: Duplicated file. Avoid this.

nyrahul (commented):
General comments:
1. It would be better to remove docker build steps from here. Better to use the same publicly shared docker images on `docker.io accuknox/kubearmor`.
2. Remove duplicated files.

Dockerfile.kubearmor
gowtham-dharsan: We cannot use symbolic link, Reason when building Dockerfiles ( it need to copy multiple folders from parent directory ) but it would try to copy from KubeArmor/build/ only if we use sym link. Also we cannot use that directly still it fails to copy folder which is in parent directory.

gowtham-dharsan: Possibility: We can remove Dockerfile.kubearmor from KubeArmor/build and place it in parent directory instead. That's what has been done now, we may have to remove it from KubeArmor/build if you are alright with that.

helm/templates/clusterrolebinding.yaml
gowtham-dharsan: Rahul, Right now the deployment strategy we are using is, deploying services in ArgoCD **directly from Helm artifactory**, So helm chart is mandatory for CD.

Jenkinsfile
```diff
@@ -0,0 +1,7 @@
+@Library('KloudOne-Jenkins-CI-library@master') _
```
nyrahul, line 1: Please remove this deployment specific file.

deployments/GKE/kubearmor.yaml
```diff
@@ -60,7 +60,8 @@ spec:
         imagePullPolicy: Always
         securityContext:
           privileged: true
-        args: ["-port=32767", "-output=/tmp/kubearmor.log"]
+        args: ["-gRPC=32767", "-logPath=/tmp/kubearmor.log"]
```
nyrahul, line 63: Why is this change?

nyrahul: @nam-jaehyun , can you please check if this is ok?

nyrahul (changes requested):

Dockerfile.kubearmor
```diff
@@ -0,0 +1,31 @@
...
+### Make executable image
+
+FROM alpine:3.12
+
+RUN apk update
+RUN echo "@edge http://dl-cdn.alpinelinux.org/alpine/edge/testing" | tee -a /etc/apk/repositories
+
+RUN apk update
+RUN apk add bash curl procps bcc bcc-dev
+RUN apk add apparmor@edge
```
nam-jaehyun, line 25: Before:
RUN apk add apparmor@edge

Need to change:
RUN apk add apparmor@edge apparmor-utils@edge

Jenkinsfile
nam-jaehyun: This file will be removed later, right?

deployments/GKE/kubearmor.yaml
nam-jaehyun: This change was right, but this file was already updated in the master branch.

nam-jaehyun:
Close this PR. If needed, please create a new PR based on the latest code.

PR #105, opened 2021-03-23
multiple protocols/caps handling in the same rule

### LINKED ISSUE #91: specify multiple protocols, capabilities in the same rule
KubeArmor currently allows enabling/disabling network protocols and system capabilities, one protocol/cap at a time.
It should be possible to specify multiple protocols or system capabilities in the same rule.

For e.g, if the user wants to enable TCP and UDP protocols (and not ICMP), then the user has to specify two different rules i.e, one enabling TCP and the other enabling UDP. Following is the policy specification in the context:
```yaml
network:
    matchProtocols:
    - protocol: [TCP|UDP|ICMP]       #---> Only TCP or UDP or ICMP could be specified.
      fromSource:
      - path: [absolute file path]
      - dir: [absolute directory path]
        recursive: [true:false]
```

It should be possible to specify multiple protocols in the same rule using the following (prospective) spec:
```yaml
network:
    matchProtocols:
    - protocol: [TCP,UDP,ICMP]    #---> Use comma separated protocols to specify multiple protocols.
      fromSource:
      - path: [absolute file path]
      - dir: [absolute directory path]
        recursive: [true:false]
```

The same is true for specifying multiple system capabilities.

chkhalt:
KubeArmor/common/common.go
KubeArmor/core/kubeUpdate.go
pkg/k8s/api/v1/kubearmorpolicy_types.go

The code added is able to expand any struct whose first field is a comma separated string to a list of structs (expanding the first field and duplicating the others).

The following struct, for example:

```
{"one, two, three", <anything>}
```

will be translated to:
```
{"one",  <anything>}, { "two",  <anything>}, {"three", <anything>}
```

Fixes: #91

### DIFF
--- a/KubeArmor/common/common.go
+++ b/KubeArmor/common/common.go
@@ -41,6 +41,63 @@ func ContainsElement(slice interface{}, element interface{}) bool {
 	return false
 }
 
+// ObjCommaCanBeExpanded Function
+func ObjCommaCanBeExpanded(objptr interface{}) bool {
+	ovptr := reflect.ValueOf(objptr)
+	if ovptr.Kind() != reflect.Ptr {
+		return false
+	}
+
+	ov := ovptr.Elem()
+	if ov.Kind() != reflect.Slice {
+		return false
+	}
+
+	if ov.Len() == 0 {
+		return false
+	}
+
+	ovelm := ov.Index(0)
+	if ovelm.Kind() != reflect.Struct {
+		return false
+	}
+
+	field0 := ovelm.Field(0)
+	if field0.Kind() != reflect.String {
+		return false
+	}
+
+	value := field0.Interface().(string)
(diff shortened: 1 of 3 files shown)


nam-jaehyun:
@johngv2 Could you also provide sample policies used for self-verification and some screenshots (e.g., policy application and policy verification with some inputs)?

chkhalt:
@nam-jaehyun I used some rules from examples/multiubuntu/security-policies/

chkhalt:
Here are some screenshots... now I'm testing some variations of protocols and capabilities.

Screen Shot 2021-03-23 at 20 54 20
Screen Shot 2021-03-23 at 21 06 54

nam-jaehyun (approved):
it looks good to take the code

PR #116, opened 2021-04-06
Use eBPF instead Auditd for logging

nam-jaehyun:
- Use eBPF only for logging
  - Auditd has generated a large number of raw logs since we supported host security
  - Auditd has consumed a large amount of the host disk volume
  - For those reasons, KubeArmor now uses eBPF only, not along with Auditd
  - However, we still need to keep eyes on whether there are any logs that eBPF is missing or not
- Support host security logs in the Log Client
- Add a document for testing KubeArmor


KubeArmor/monitor/logUpdate.go
```diff
@@ -142,9 +143,10 @@ func (mon *SystemMonitor) UpdateLogs() {
 					log = execLogMap[log.HostPID]
 					delete(execLogMap, log.HostPID)
 
-					if msg.ContextSys.Retval == -13 {
```
asifalix, line 145: Just checking what does the hardwired value of -13 does here.. Is there a better way to represent this with a constant?

nam-jaehyun: -13 is the error code defined in the Linux kernel headers for "Permission denied".

PR #131, opened 2021-05-17
GoLang latest stable release

ravkishu:
Auto fetch GoLang latest stable binary release


nam-jaehyun:
@ravkishu When we keep having the latest Golang, I'm wondering if there could be dependency issues on the packages used by the existing code.

PR #154, opened 2021-06-18
vagrant: dev-env based on latest kernel

### LINKED ISSUE #153: supporting netnext/latest kernels
- [x] supporting netnext for dev env setup
- [x] fixing compilation issues for kubearmor for latest kernel (>5.10)
- [x] Updating dev guide document
- [x] speed up compilation

nyrahul:
A dev-env based on the latest kernel is required to test/dev corresponding latest features from the kernel.

Fixes: #153


contribution/vagrant/Vagrantfile
```diff
@@ -1,18 +1,27 @@
...
+	# Using Cilium's Ubuntu Next image https://app.vagrantup.com/cilium/boxes/ubuntu-next
+	VM_IMG = "cilium/ubuntu-next"
+	VM_NAME = "kubearmor-dev-next"
+else
+	VM_IMG = "ubuntu/bionic64"
+	VM_NAME = "kubearmor-dev"
+end
+
 Vagrant.configure("2") do |config|
   config.vm.define "KubeArmor-Dev" do |cfg|
```
geyslan: I think that `config.vm.define` should be the same as `VM_NAME`, so the user can up and connect to both VMs interchangeably.

```diff
 Vagrant.configure("2") do |config|
-  config.vm.define "KubeArmor-Dev" do |cfg|
+  config.vm.define VM_NAME do |cfg|
```

If it is set statically, the right `VM_IMG` won't be up in despite of `NETNEXT` when other is already created.

geyslan (commented):
I made a single suggestion about `VM_NAME`.

contribution/vagrant/Vagrantfile
nyrahul: Handled, thanks

contribution/vagrant/Vagrantfile
```diff
@@ -35,4 +43,9 @@ Vagrant.configure("2") do |config|
 
   # Initialize Kubernetes
   config.vm.provision :shell, path: kubearmor_home + "/contribution/self-managed-k8s/k8s/initialize_kubernetes.sh"
+
+  # Reboot to use new kernel
+  if ENV['NETNEXT'] == "1" then
+	  config.vm.provision :reload
```
geyslan, line 49:
```bash
vm:
* The 'reload' provisioner could not be found.
```

It's necessary to install the plugin `reload`.

```diff
@@ -9,6 +9,7 @@ else
 end
 
 Vagrant.configure("2") do |config|
+  config.vagrant.plugins = "vagrant-reload"
   config.vm.define VM_NAME do |cfg|
     cfg.vm.box = VM_IMG
     cfg.vm.provider "virtualbox" do |vb|
@@ -46,6 +47,6 @@ Vagrant.configure("2") do |config|
 
   # Reboot to use new kernel
   if ENV['NETNEXT'] == "1" then
-     config.vm.provision :reload
+    config.vm.provision :reload
   end
 end
```

But reloading is causing ownership loss.

```bash
vagrant@kubearmor-dev-next:~$ ls -l ~/
total 8
drwxr-xr-x 1 root    root    4096 Jun 19 14:41 KubeArmor
drwxr-xr-x 4 vagrant vagrant 4096 Jun 21 23:23 go
```

And it seems tricky to solve: https://stackoverflow.com/questions/23894739/synced-folders-lost-when-rebooting-a-vagrant-machine-using-the-ansible-provision

nyrahul: Yes, reloading is necessary because we are updating the kernel and we need to switch to the new one.

Reloading (even if you do vagrant halt/up) is causing ownership loss if you have an old VirtualBox-guest. If you try the `NETNEXT=1 vagrant plugin install vagrant-vbguest` this will update the VirtualBox-guest image and things work after that. The other option is for the user to use sudo when running kubearmor.

There is no other easy way to solve this problem. Using NFS to mount the synced_folder would be another option but it requires some other configuration.

nyrahul: I have updated the documentation to explain these points.

geyslan: This looks great. One other way would be to declare the needed plugins inside the Vagrantfile.

```diff
@@ -9,6 +9,7 @@ else
 end
 
 Vagrant.configure("2") do |config|
+  config.vagrant.plugins = ["vagrant-vbguest", "vagrant-reload"]
   config.vm.define VM_NAME do |cfg|
```

It will ask one to install plugins locally and to run the requested command again.

```bash
❯ NETNEXT=1 vagrant up
Vagrant has detected project local plugins configured for this
project which are not installed.

  vagrant-vbguest
Install local plugins (Y/N) [N]: y
Installing the 'vagrant-vbguest' plugin. This can take a few minutes...
Fetching micromachine-3.0.0.gem
Fetching vagrant-vbguest-0.30.0.gem
Installed the plugin 'vagrant-vbguest (0.30.0)'!


Vagrant has completed installing local plugins for the current Vagrant
project directory. Please run the requested command again.
```

nyrahul: This is even better. Will try to handle this as part of another PR. Thanks

PR #152, opened 2021-06-16
feeder: policyMatcher: Refactoring SecurityPolices functions and adding matchPatterns logic

geyslan:
This refactors the `UpdateSecurityPolicies()` and  `UpdateHostSecurityPolicies()` providing the same logic
with less code through helper functions, reducing checking and execution.

This also includes logic for `matchPatterns` and adds the `ksp-ubuntu-1-file-pattern-block.yaml` security policy example.

As an additional implementation, this proposes changes to the `KubeArmor/types/types.go:MatchPolicy` structure by adding a `Regexp` pointer field to it.

Fixes #143

PR #155, opened 2021-06-21
Add policy match support for native AppArmor policies

### LINKED ISSUE #151: Add policy match support for native AppArmor policies
KubeArmor allows native apparmor policies to be added as it is in YAML policies. These native policies are being enforced but do not generate any logs upon policy violations. This issue can be reproduced by simply applying any native apparmor policy and violating it in the selected pods. This should have ideally generated logs in the kubearmor.log file. But the log file will contain no such events.
```
apiVersion: security.accuknox.com/v1
kind: KubeArmorPolicy
metadata:
  name: native-apparmor-policy-1
  namespace: multiubuntu
spec:
  severity: 5
  message: "native read-only etc"
  selector:
    matchLabels:
      group: group-1
  apparmor: |
        deny /etc/{*,**} w,
```
After applying this policy we can check in one of the ubuntu pods with group=group-1 by trying to violate the policy. In this example it makes /etc/ directory read-only. 
```
vagrant@kubearmor-dev:/etc/apparmor.d$ kubectl exec -ti -n multiubuntu ubuntu-1-deployment-5c475fb4fb-rfttn -- /bin/bash
root@ubuntu-1-deployment-5c475fb4fb-rfttn:/# cd /etc
root@ubuntu-1-deployment-5c475fb4fb-rfttn:/etc# echo "data" >> hosts
bash: hosts: Permission denied
``` 
This violation should generate logs. We can check it by looking the contents of /tmp/kubearmor.log inside the kubearmor container.
```
vagrant@kubearmor-dev:/etc/apparmor.d$ kubectl exec -ti -n kube-system kubearmor-kp6kr -- /bin/bash
bash-5.0# cat /tmp/kubearmor.log

```
We can see its empty and does not contain any log for violation.
This could be the result of feeder not tracking native policy entries in its security policy map.[1]https://github.com/accuknox/KubeArmor/blob/cc3b39c19ba6f23cd418d042db7eac4e0b145e06/KubeArmor/feeder/policyMatcher.go#L446
Possible fix would be to modify it to add native apparmor policies to [1].
We might need to modify struct MatchPolicy[2] as well adding by adding field(s) for native policies.
https://github.com/accuknox/KubeArmor/blob/cc3b39c19ba6f23cd418d042db7eac4e0b145e06/KubeArmor/types/types.go#L179

oneiro-naut:
This makes it possible for native AppArmor policy logs to be dumped to kubearmor logs.
The logs are only weakly matched so unrelated denied events might be dumped marked as MatchedNativePolicy.

Fixes: #151

### DIFF
--- a/KubeArmor/feeder/feeder.go
+++ b/KubeArmor/feeder/feeder.go
@@ -473,7 +473,7 @@ func (fd *Feeder) PushLog(log tp.Log) error {
 
 	// gRPC output
 
-	if log.Type == "MatchedPolicy" || log.Type == "MatchedHostPolicy" {
+	if log.Type == "MatchedPolicy" || log.Type == "MatchedNativePolicy" || log.Type == "MatchedHostPolicy" {
 		pbAlert := pb.Alert{}
 
 		pbAlert.Timestamp = log.Timestamp
--- a/KubeArmor/feeder/policyMatcher.go
+++ b/KubeArmor/feeder/policyMatcher.go
@@ -20,6 +20,18 @@ func (fd *Feeder) UpdateSecurityPolicies(action string, conGroup tp.ContainerGro
 		matches := tp.MatchPolicies{}
 
 		for _, secPolicy := range conGroup.SecurityPolicies {
+			if len(secPolicy.Spec.Apparmor) > 0 {
+				match := tp.MatchPolicy{}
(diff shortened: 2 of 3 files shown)


geyslan:
This PR changes `feeder/policyMatcher.go` that is also modified by PR #152. The last to be merged must rebase. So let's keep following them. Thanks.

PR #190, opened 2021-07-07
policyMatcher: fix bug introduced by PR #188 (len is always true)

geyslan:
PR #188 introduced the `ResourceType` field and its length check that always returns true.

This moves the else branch to be the default switch clause.

This also simplifies the boolean comparison to silent the `go-staticcheck` (S1002).


nam-jaehyun:
Thanks for the PR, but I updated the code with some other changes.

PR #196, opened 2021-07-09
Added security policy validator in custom controller by [VenkataReddy]

### LINKED ISSUE #104: Need a security policy validator and supporting per-field/rule action/severity
When we apply security policies based on KubeArmor's CRD, we need to check two things.

1. Syntax

CRD basically contains syntax validation logic for each field, but it can only check input types or patterns.

2. Semantics

CRD doesn't contain any validations for semantics, meaning that a user needs to properly define security policies without semantics errors.

Here are some examples.
- 'ownerOnly' works with the Allow action.
- 'recursive' is only effective with directories, not paths.
- NodeSelector cannot be defined with selector.

As a result, we need a security policy validator at the Kubernetes side in order to check syntax and semantics errors
as soon as some policies are applied into Kubernetes and before they are detected by KubeArmor.

The validator might be implemented as a custom controller along with the CRD.

CBVenkataReddy:
Inside the controllers/kubearmorpolicy_controller.go and controllers/kubearmorhostpolicy_controller.go file we have the function ”Reconcile”, Here the function “Reconcile” act as a security policy validator at the Kubernetes side in order to check syntax and semantics errors. So business logic (below 1 & 2 points) for validating the syntax and semantics errors will be part of “Reconcile” function.
1.check policies already applied in kubernetes at the controller.
2. Update the policy status for valid(OK)/invalid(Not OK), 
 delete the invalid policy and return the fail code when a new policy is applied and there are some issues in the policy.

Fixes: #104

### DIFF
--- a/KubeArmor/core/kubeUpdate.go
+++ b/KubeArmor/core/kubeUpdate.go
@@ -589,7 +589,7 @@ func (dm *KubeArmorDaemon) GetSecurityPolicies(identities []string) []tp.Securit
 }
 
 // UpdateSecurityPolicy Function
-func (dm *KubeArmorDaemon) UpdateSecurityPolicy(action string, secPolicy tp.SecurityPolicy) {
+func (dm *KubeArmorDaemon) UpdateSecurityPolicy(action string, secPolicy tp.SecurityPolicy, status string) {
 	dm.ContainerGroupsLock.Lock()
 	defer dm.ContainerGroupsLock.Unlock()
 
@@ -626,7 +626,9 @@ func (dm *KubeArmorDaemon) UpdateSecurityPolicy(action string, secPolicy tp.Secu
 			dm.LogFeeder.UpdateSecurityPolicies("UPDATED", dm.ContainerGroups[idx])
 
 			// enforce security policies
-			dm.RuntimeEnforcer.UpdateSecurityPolicies(dm.ContainerGroups[idx])
+			if status == "OK" {
+				dm.RuntimeEnforcer.UpdateSecurityPolicies(dm.ContainerGroups[idx])
+			}
 		}
 	}
 }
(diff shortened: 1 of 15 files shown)


nam-jaehyun:
1. please update the code on the latest files.
after we moved out from accuknox, KubeArmor uses security.kubearmor.com instead of security.accuknox.com.
2. let's say we have controllers for KubeArmorPolicy and KubeArmorHostPolicy.
how can we deploy those controllers into Kubernetes?
3. It would be better if you have some screenshots that show how the controllers work.

nam-jaehyun:
@CBVenkataReddy how did you generate the YAML files for controllers?

CBVenkataReddy:
> @CBVenkataReddy how did you generate the YAML files for controllers?

I have used "kustomize" for customizing the manifests & generating YAML files "kustomize build config/default"

KubeArmor/core/kubeUpdate.go
```diff
@@ -670,7 +672,7 @@ func (dm *KubeArmorDaemon) WatchSecurityPolicies() {
 						if policy.Metadata["policyName"] == secPolicy.Metadata["policyName"] &&
 							policy.Metadata["namespaceName"] == secPolicy.Metadata["namespaceName"] &&
 							policy.Metadata["generation"] == secPolicy.Metadata["generation"] {
-							exist = true
+							//exist = true
```
nam-jaehyun, line 675: why this line is disabled?

CBVenkataReddy: Here the condition basically checks weather policy exists or not, if it exists then it continue without applying security policies to containers (UpdateSecurityPolicy). 

Controller validates KubeArmrorPolicy schema based on the validation it updates the status, 
After controller status updation KubeArmror gets event (MODIFIED) as policy exit condition becomes true and sets "exist = true" and continue back to check another events in infinite loop.

PR #199, opened 2021-07-11
Enable already-deployed container monitoring in system_monitor

### LINKED ISSUE #163: Security policies not working if Pods are deployed before KubeArmor
Security policies can't be deleted, modified and added to the pods which were deployed prior to KubeArmor deployment. #162 is a result of this issue. The following steps were ran on the kubearmor-dev vagrant VM.
Steps to reproduce the bug:
1. Deploy multiubuntu pods.
`vagrant@kubearmor-dev:~$ kubectl apply -f KubeArmor/examples/multiubuntu/`
2. Deploy KubeArmor.
`vagrant@kubearmor-dev:~$ kubectl apply -f KubeArmor/deployments/docker/`
3. Apply any security policy to multiubuntu pods.
```
vagrant@kubearmor-dev:~$ kubectl apply -f KubeArmor/examples/multiubuntu/security-policies/ksp-group-1-proc-path-block.yaml 
kubearmorpolicy.security.accuknox.com/ksp-group-1-proc-path-block created
```
4. Check if policy is enforced. Its not being enforced (we don't get any permission denied in this example).
```
vagrant@kubearmor-dev:~$ kubectl exec -ti -n multiubuntu ubuntu-1-deployment-5c475fb4fb-sq6xv -- /bin/bash
root@ubuntu-1-deployment-5c475fb4fb-sq6xv:/# /bin/sleep 1
```
NOTE: 
- kubectl shows deleted and created events but the actual AppArmor profiles in `/etc/apparmor.d/` aren't being modified or enforced. 
- The same happens if KubeArmor **restarts** since these are essentially the same conditions, ie pods being up before KubeArmor itself is up.

oneiro-naut:
This makes it possible to have already-deployed container entries in pid_ns_map.
This will enable log generation for these containers.

Fixes: #163

### DIFF
--- a/KubeArmor/BPF/system_monitor.c
+++ b/KubeArmor/BPF/system_monitor.c
@@ -182,12 +182,8 @@ static __always_inline u32 add_pid_ns()
         return pid_ns;
     }
 
-    if (get_task_ns_pid(task) == 1) {
-        pid_ns_map.update(&pid_ns, &one);
-        return pid_ns;
-    }
-
-    return 0;
+    pid_ns_map.update(&pid_ns, &one);
+    return pid_ns;
 
 #endif /* MONITOR_HOST || MONITOR_CONTAINER */
 }


nam-jaehyun:
This modification allows the eBPF monitor to keep track of the processes in the already-deployed containers.
I'm wondering how we can map container IDs and PID namespaces at this point and how we can generate alerts.

oneiro-naut:
> I'm wondering how we can map container IDs and PID namespaces at this point and how we can generate alerts.

I think mapping is already working the way it worked before and we won't explicitly need to do that since I am able to get all the required logs. As soon as an entry is added the security policy logs and ContainerLogs start working for the container and are being dumped to feeder. I tested it with sample security policies in multiubuntu.

nam-jaehyun:
@oneiro-naut roll back to the previous code since it generates non-container related logs too.

oneiro-naut:
@nam-jaehyun Sure thing.

Issue #193, opened 2021-07-08
Lost events coming from eBPF-based monitor


nam-jaehyun:
Currently, there are some cases that KubeArmor misses some events coming from the eBPF monitor.
This usually happens when a host has limited resources.
In addition, KubeArmor cannot sometimes generate alerts due to such losses.
We may be able to reproduce such losses by running the automated testing framework several times
and see the failed test cases.

oneiro-naut:
@nam-jaehyun Is it true for ContainerLogs as well? 
https://github.com/kubearmor/KubeArmor/blob/f4fba7cf59b62145fffeff5c63d74bd89fbd96b2/KubeArmor/feeder/feeder.go#L553
I tried to get these logs using kubearmor-logclient and didn't get some of them.

dku-boanlab:
I think that's because you didn't add the 'kubearmor-visibility' annotation.

nam-jaehyun:
@oneiro-naut Can you take this issue?

oneiro-naut:
@nam-jaehyun Sure thing. Thanks.

nam-jaehyun:
For testing, I suggest this way.

1. Run KubeArmor
```
~/KubeArmor/KubeArmor$ sudo -E ./kubearmor -logPath=/tmp/kubearmor.log -logFilter=all -enableHostPolicy
```

2. Run the runtime test script
```
~/KubeArmor/tests$ ./test-scenarios-in-runtime.sh
```
- See if there are any failed cases

3. Repeatedly run the runtime test script even though there was no failed case in the previous execution.
```
~/KubeArmor/tests$ ./test-scenarios-in-runtime.sh
```
- See if there are any failed cases
- At this step, please do not re-run KubeArmor since we also want to see if KubeArmor works during repeated container deployments and deletions

4. After running the script multiple times, check the KubeArmor log (/tmp/kubearmor.log) and debug the issue.

nam-jaehyun:
Observation 1

```
[...]

PR #207, opened 2021-07-16
update the way to map containerIDs and PidNS/MntNS

nam-jaehyun:
- update the way to map containers to PID/MNT namespaces
  (related to #193)
- update test scripts

### DIFF
--- a/KubeArmor/common/common.go
+++ b/KubeArmor/common/common.go
@@ -244,6 +244,12 @@ func StrToFile(str, destFile string) {
 	if err != nil {
 		kg.Err(err.Error())
 	}
+
+	// sync the file
+	err = file.Sync()
+	if err != nil {
+		kg.Err(err.Error())
+	}
 }
 
 // ============= //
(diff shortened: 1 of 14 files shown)


oneiro-naut:
@nam-jaehyun The mnt_id we get in ebpf code(system_monitor.c) does not match with any actual mnt_id we get using readlink command. And so we are not getting any container log. Because this lookup returns empty containerID(because key used does not match with keys already present):https://github.com/kubearmor/KubeArmor/blob/155b35d3bfb0c8057b3aa102bf8f6d7a0cb42530/KubeArmor/monitor/systemMonitor.go#L454
And so all of them are skipped here:
https://github.com/kubearmor/KubeArmor/blob/155b35d3bfb0c8057b3aa102bf8f6d7a0cb42530/KubeArmor/monitor/systemMonitor.go#L467

nam-jaehyun:
@oneiro-naut could you show me some screenshots for better understanding?

oneiro-naut:
@nam-jaehyun I just compared these 2 MntID and they seem to be different for same containerID.
[1]https://github.com/kubearmor/KubeArmor/blob/155b35d3bfb0c8057b3aa102bf8f6d7a0cb42530/KubeArmor/monitor/systemMonitor.go#L454
and
[2]https://github.com/kubearmor/KubeArmor/blob/155b35d3bfb0c8057b3aa102bf8f6d7a0cb42530/KubeArmor/core/dockerHandler.go#L233

oneiro-naut:
@nam-jaehyun I found the cause. We need to incorporate [this]( https://github.com/iovisor/bcc/pull/3540) change in mnt_namespace struct defintion based on kernel versions. I tested it and now its working.

nam-jaehyun:
@oneiro-naut could you make a PR for this change?

oneiro-naut:
> @oneiro-naut could you make a PR for this change?

Sure thing.

PR #211, opened 2021-07-19
Fix missing logs and incorrect logs

### LINKED ISSUE #193: Lost events coming from eBPF-based monitor

oneiro-naut:
Fixes: #193

### DIFF
--- a/KubeArmor/core/containerdHandler.go
+++ b/KubeArmor/core/containerdHandler.go
@@ -246,8 +246,28 @@ func (dm *KubeArmorDaemon) UpdateContainerdContainer(ctx context.Context, contai
 		dm.ContainersLock.Lock()
 		if _, ok := dm.Containers[containerID]; !ok {
 			dm.Containers[containerID] = container
-		} else if dm.Containers[containerID].AppArmorProfile == "" && container.AppArmorProfile != "" {
-			dm.Containers[containerID] = container
+		} else if dm.Containers[container.ContainerID].PidNS == 0 && dm.Containers[container.ContainerID].MntNS == 0 {
+			// this entry was updated by kubernetes before docker detects it
+			// thus, we here use the info given by kubernetes instead of the info given by docker
+
+			container.NamespaceName = dm.Containers[container.ContainerID].NamespaceName
+			container.ContainerGroupName = dm.Containers[container.ContainerID].ContainerGroupName
+			container.ContainerName = dm.Containers[container.ContainerID].ContainerName
+
+			container.PolicyEnabled = dm.Containers[container.ContainerID].PolicyEnabled
+			container.ProcessVisibilityEnabled = dm.Containers[container.ContainerID].ProcessVisibilityEnabled
+			container.FileVisibilityEnabled = dm.Containers[container.ContainerID].FileVisibilityEnabled
+			container.NetworkVisibilityEnabled = dm.Containers[container.ContainerID].NetworkVisibilityEnabled
+			container.CapabilitiesVisibilityEnabled = dm.Containers[container.ContainerID].CapabilitiesVisibilityEnabled
+
+			dm.Containers[container.ContainerID] = container
+
+			for _, conGroup := range dm.ContainerGroups {
+				if conGroup.ContainerGroupName == container.ContainerGroupName && conGroup.AppArmorProfiles[container.ContainerID] == "" {
+					conGroup.AppArmorProfiles[container.ContainerID] = container.AppArmorProfile
+					break
+				}
+			}
 		} else {
 			dm.ContainersLock.Unlock()
 			return false
@@ -289,8 +309,6 @@ func (dm *KubeArmorDaemon) MonitorContainerdEvents() {
 
 	dm.LogFeeder.Print("Started to monitor Containerd events")
 
-	getContainersExecutedOnce := false
-
 	for {
 		select {
 		case <-StopChan:
@@ -325,20 +343,6 @@ func (dm *KubeArmorDaemon) MonitorContainerdEvents() {
 				for containerID, context := range deletedContainers {
 					dm.UpdateContainerdContainer(context, containerID, "destroy")
 				}
-			}
-
-			if !getContainersExecutedOnce {
-				for _, conGroup := range dm.ContainerGroups {
-					for _, containerID := range conGroup.Containers {
-						if container, ok := dm.Containers[containerID]; ok {
-							if conGroup.AppArmorProfiles[containerID] == "" {
-								conGroup.AppArmorProfiles[containerID] = container.AppArmorProfile
-							}
-						}
-					}
-				}
-
-				getContainersExecutedOnce = true
 			}
 		}
 
--- a/KubeArmor/core/dockerHandler.go
+++ b/KubeArmor/core/dockerHandler.go
@@ -176,8 +176,28 @@ func (dm *KubeArmorDaemon) GetAlreadyDeployedDockerContainers() {
 				dm.ContainersLock.Lock()
 				if _, ok := dm.Containers[container.ContainerID]; !ok {
 					dm.Containers[container.ContainerID] = container
-				} else if dm.Containers[container.ContainerID].AppArmorProfile == "" && container.AppArmorProfile != "" {
+				} else if dm.Containers[container.ContainerID].PidNS == 0 && dm.Containers[container.ContainerID].MntNS == 0 {
+					// this entry was updated by kubernetes before docker detects it
+					// thus, we here use the info given by kubernetes instead of the info given by docker
+
+					container.NamespaceName = dm.Containers[container.ContainerID].NamespaceName
+					container.ContainerGroupName = dm.Containers[container.ContainerID].ContainerGroupName
+					container.ContainerName = dm.Containers[container.ContainerID].ContainerName
+
+					container.PolicyEnabled = dm.Containers[container.ContainerID].PolicyEnabled
+					container.ProcessVisibilityEnabled = dm.Containers[container.ContainerID].ProcessVisibilityEnabled
+					container.FileVisibilityEnabled = dm.Containers[container.ContainerID].FileVisibilityEnabled
+					container.NetworkVisibilityEnabled = dm.Containers[container.ContainerID].NetworkVisibilityEnabled
+					container.CapabilitiesVisibilityEnabled = dm.Containers[container.ContainerID].CapabilitiesVisibilityEnabled
+
 					dm.Containers[container.ContainerID] = container
+
+					for _, conGroup := range dm.ContainerGroups {
+						if conGroup.ContainerGroupName == container.ContainerGroupName && conGroup.AppArmorProfiles[container.ContainerID] == "" {
+							conGroup.AppArmorProfiles[container.ContainerID] = container.AppArmorProfile
+							break
+						}
+					}
 				} else {
 					dm.ContainersLock.Unlock()
 					continue
@@ -192,15 +212,6 @@ func (dm *KubeArmorDaemon) GetAlreadyDeployedDockerContainers() {
 		}
 	}
 
-	for _, conGroup := range dm.ContainerGroups {
-		for _, containerID := range conGroup.Containers {
-			if container, ok := dm.Containers[containerID]; ok {
-				if conGroup.AppArmorProfiles[containerID] == "" {
-					conGroup.AppArmorProfiles[containerID] = container.AppArmorProfile
-				}
-			}
-		}
-	}
 }
 
 // UpdateDockerContainer Function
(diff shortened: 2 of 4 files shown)


KubeArmor/feeder/feeder.go
```diff
@@ -529,10 +532,9 @@ func (fd *Feeder) PushLog(log tp.Log) {
 
 	// standard output / file output
 
-	if fd.filter == "policy" {
-		if len(log.PolicyName) > 0 {
+	if fd.Filter == "policy" {
+		if len(log.PolicyName) > 0 && log.PolicyName != "HostLog" {
```
nam-jaehyun: why do we need to filter out HostLog?

KubeArmor/core/kubeUpdate.go
```diff
@@ -423,6 +428,7 @@ func (dm *KubeArmorDaemon) WatchK8sPods() {
 						}
 					}
 
+					pod.AlreadyDeployed = !updateAppArmor
```
nam-jaehyun: its coverage is too huge. KubeArmor runs on the environments with no LSM or some other LSMs except for AppArmor, which means that we will always see pod.AlreadyDeployed=true in such environments.

KubeArmor/core/kubeUpdate.go
```diff
@@ -507,6 +513,7 @@ func (dm *KubeArmorDaemon) WatchK8sPods() {
 						}
 					}
 
+					pod.AlreadyDeployed = !updateSELinux
```
nam-jaehyun: the same comment (refer to the above one).

KubeArmor/core/kubeUpdate.go
nam-jaehyun: When I tested the code, I think we don't need the AlreadyDeployed flag.
Without the flag, it works fine in the following environments.
- ubuntu 20.04 with docker, ubuntu 20.04 with containerd, ubuntu 18.04 with microk8s
- minikube with 4.19 kernel, ubuntu 18.04 (local), ubuntu 20.04 (local)

KubeArmor/core/containerdHandler.go
```diff
@@ -246,8 +246,21 @@ func (dm *KubeArmorDaemon) UpdateContainerdContainer(ctx context.Context, contai
 		dm.ContainersLock.Lock()
 		if _, ok := dm.Containers[containerID]; !ok {
 			dm.Containers[containerID] = container
-		} else if dm.Containers[containerID].AppArmorProfile == "" && container.AppArmorProfile != "" {
+		} else if dm.Containers[containerID].PidNS == 0 && dm.Containers[containerID].MntNS == 0 {
```
nam-jaehyun:
```
dm.ContainersLock.Lock()
if _, ok := dm.Containers[containerID]; !ok {
	dm.Containers[containerID] = container
} else if dm.Containers[container.ContainerID].PidNS == 0 && dm.Containers[container.ContainerID].MntNS == 0 {
	// this entry was updated by kubernetes before docker detects it
	// thus, we here use the info given by kubernetes instead of the info given by docker

	container.NamespaceName = dm.Containers[container.ContainerID].NamespaceName
	container.ContainerGroupName = dm.Containers[container.ContainerID].ContainerGroupName
	container.ContainerName = dm.Containers[container.ContainerID].ContainerName

	container.PolicyEnabled = dm.Containers[container.ContainerID].PolicyEnabled
	container.ProcessVisibilityEnabled = dm.Containers[container.ContainerID].ProcessVisibilityEnabled
	container.FileVisibilityEnabled = dm.Containers[container.ContainerID].FileVisibilityEnabled
	container.NetworkVisibilityEnabled = dm.Containers[container.ContainerID].NetworkVisibilityEnabled
	container.CapabilitiesVisibilityEnabled = dm.Containers[container.ContainerID].CapabilitiesVisibilityEnabled

	dm.Containers[container.ContainerID] = container

	for _, conGroup := range dm.ContainerGroups {
		if conGroup.ContainerGroupName == container.ContainerGroupName && conGroup.AppArmorProfiles[container.ContainerID] == "" {
			conGroup.AppArmorProfiles[container.ContainerID] = container.AppArmorProfile
			break
		}
	}
} else {
	dm.ContainersLock.Unlock()
	return false
}
dm.ContainersLock.Unlock()
```

KubeArmor/core/dockerHandler.go
```diff
@@ -176,8 +176,20 @@ func (dm *KubeArmorDaemon) GetAlreadyDeployedDockerContainers() {
 				dm.ContainersLock.Lock()
 				if _, ok := dm.Containers[container.ContainerID]; !ok {
 					dm.Containers[container.ContainerID] = container
-				} else if dm.Containers[container.ContainerID].AppArmorProfile == "" && container.AppArmorProfile != "" {
+				} else if dm.Containers[container.ContainerID].PidNS == 0 && dm.Containers[container.ContainerID].MntNS == 0 {
```
nam-jaehyun, line 179:
```
dm.ContainersLock.Lock()
if _, ok := dm.Containers[container.ContainerID]; !ok {
	dm.Containers[container.ContainerID] = container
} else if dm.Containers[container.ContainerID].PidNS == 0 && dm.Containers[container.ContainerID].MntNS == 0 {
	// this entry was updated by kubernetes before docker detects it
	// thus, we here use the info given by kubernetes instead of the info given by docker

	container.NamespaceName = dm.Containers[container.ContainerID].NamespaceName
	container.ContainerGroupName = dm.Containers[container.ContainerID].ContainerGroupName
	container.ContainerName = dm.Containers[container.ContainerID].ContainerName

	container.PolicyEnabled = dm.Containers[container.ContainerID].PolicyEnabled
	container.ProcessVisibilityEnabled = dm.Containers[container.ContainerID].ProcessVisibilityEnabled
	container.FileVisibilityEnabled = dm.Containers[container.ContainerID].FileVisibilityEnabled
	container.NetworkVisibilityEnabled = dm.Containers[container.ContainerID].NetworkVisibilityEnabled
	container.CapabilitiesVisibilityEnabled = dm.Containers[container.ContainerID].CapabilitiesVisibilityEnabled

	dm.Containers[container.ContainerID] = container

	for _, conGroup := range dm.ContainerGroups {
		if conGroup.ContainerGroupName == container.ContainerGroupName && conGroup.AppArmorProfiles[container.ContainerID] == "" {
			conGroup.AppArmorProfiles[container.ContainerID] = container.AppArmorProfile
			break
		}
	}
} else {
	dm.ContainersLock.Unlock()
	continue
}
dm.ContainersLock.Unlock()
```

KubeArmor/core/dockerHandler.go
```diff
@@ -223,6 +226,21 @@ func (dm *KubeArmorDaemon) UpdateDockerContainer(containerID, action string) {
 		dm.ContainersLock.Lock()
 		if _, ok := dm.Containers[containerID]; !ok {
 			dm.Containers[containerID] = container
+		} else if dm.Containers[containerID].PidNS == 0 && dm.Containers[containerID].MntNS == 0 {
```
nam-jaehyun, line 237:
```
dm.ContainersLock.Lock()
if _, ok := dm.Containers[containerID]; !ok {
	dm.Containers[containerID] = container
} else if dm.Containers[containerID].PidNS == 0 && dm.Containers[containerID].MntNS == 0 {
	// this entry was updated by kubernetes before docker detects it
	// thus, we here use the info given by kubernetes instead of the info given by docker

	container.NamespaceName = dm.Containers[containerID].NamespaceName
	container.ContainerGroupName = dm.Containers[containerID].ContainerGroupName
	container.ContainerName = dm.Containers[containerID].ContainerName

	container.PolicyEnabled = dm.Containers[containerID].PolicyEnabled
	container.ProcessVisibilityEnabled = dm.Containers[containerID].ProcessVisibilityEnabled
	container.FileVisibilityEnabled = dm.Containers[containerID].FileVisibilityEnabled
	container.NetworkVisibilityEnabled = dm.Containers[containerID].NetworkVisibilityEnabled
	container.CapabilitiesVisibilityEnabled = dm.Containers[containerID].CapabilitiesVisibilityEnabled

	dm.Containers[containerID] = container

	for _, conGroup := range dm.ContainerGroups {
		if conGroup.ContainerGroupName == container.ContainerGroupName && conGroup.AppArmorProfiles[containerID] == "" {
			conGroup.AppArmorProfiles[containerID] = container.AppArmorProfile
			break
		}
	}
} else {
	dm.ContainersLock.Unlock()
	return
}
dm.ContainersLock.Unlock()
```

KubeArmor/core/kubeUpdate.go
```diff
@@ -59,6 +59,28 @@ func (dm *KubeArmorDaemon) UpdateContainerGroupWithPod(action string, pod tp.K8s
 			}
 		}
 
+		// parse annotations and set flags
```
nam-jaehyun: just ordering...

```
// update flags
if pod.Annotations["kubearmor-policy"] == "enabled" {
	newGroup.PolicyEnabled = tp.KubeArmorPolicyEnabled
} else if pod.Annotations["kubearmor-policy"] == "audited" {
	newGroup.PolicyEnabled = tp.KubeArmorPolicyAudited
} else {
	newGroup.PolicyEnabled = tp.KubeArmorPolicyDisabled
}

// parse annotations and set flags
for _, visibility := range strings.Split(pod.Annotations["kubearmor-visibility"], ",") {
	if visibility == "process" {
		newGroup.ProcessVisibilityEnabled = true
	} else if visibility == "file" {
		newGroup.FileVisibilityEnabled = true
	} else if visibility == "network" {
		newGroup.NetworkVisibilityEnabled = true
	} else if visibility == "capabilities" {
		newGroup.CapabilitiesVisibilityEnabled = true
	}
}
```

KubeArmor/core/kubeUpdate.go
```diff
@@ -127,6 +127,11 @@ func (dm *KubeArmorDaemon) UpdateContainerGroupWithPod(action string, pod tp.K8s
 			dm.RuntimeEnforcer.UpdateSecurityProfiles(action, pod, true)
 		}
 
+		if pod.AlreadyDeployed {
```
nam-jaehyun: I think we don't need this

KubeArmor/core/kubeUpdate.go
```diff
@@ -170,6 +175,37 @@ func (dm *KubeArmorDaemon) UpdateContainerGroupWithPod(action string, pod tp.K8s
 			}
 		}
 
+		// parse annotations and set flags
```
nam-jaehyun: just ordering...

```
// update flags

prevPolicyEnabled := dm.ContainerGroups[conGroupIdx].PolicyEnabled

if pod.Annotations["kubearmor-policy"] == "enabled" {
	dm.ContainerGroups[conGroupIdx].PolicyEnabled = tp.KubeArmorPolicyEnabled
} else if pod.Annotations["kubearmor-policy"] == "audited" {
	dm.ContainerGroups[conGroupIdx].PolicyEnabled = tp.KubeArmorPolicyAudited
} else {
	dm.ContainerGroups[conGroupIdx].PolicyEnabled = tp.KubeArmorPolicyDisabled
}

// parse annotations and set flags

dm.ContainerGroups[conGroupIdx].ProcessVisibilityEnabled = false
dm.ContainerGroups[conGroupIdx].FileVisibilityEnabled = false
dm.ContainerGroups[conGroupIdx].NetworkVisibilityEnabled = false
dm.ContainerGroups[conGroupIdx].CapabilitiesVisibilityEnabled = false

for _, visibility := range strings.Split(pod.Annotations["kubearmor-visibility"], ",") {
	if visibility == "process" {
		dm.ContainerGroups[conGroupIdx].ProcessVisibilityEnabled = true
	} else if visibility == "file" {
		dm.ContainerGroups[conGroupIdx].FileVisibilityEnabled = true
	} else if visibility == "network" {
		dm.ContainerGroups[conGroupIdx].NetworkVisibilityEnabled = true
	} else if visibility == "capabilities" {
		dm.ContainerGroups[conGroupIdx].CapabilitiesVisibilityEnabled = true
	}
}
```

KubeArmor/core/kubeUpdate.go
nam-jaehyun: I think we don't need this

KubeArmor/core/kubeUpdate.go
nam-jaehyun: I think we don't need this too

KubeArmor/feeder/policyMatcher.go
```diff
@@ -912,7 +912,6 @@ func (fd *Feeder) UpdateMatchedPolicy(log tp.Log) tp.Log {
 				// use 'AllowWithAudit' to get the logs for allowed operations
 				return tp.Log{}
 			}
```
nam-jaehyun: empty line

KubeArmor/types/types.go
```diff
@@ -78,11 +78,12 @@ type ContainerGroup struct {
...
-	Annotations map[string]string
-	Labels      map[string]string
-	Containers  map[string]string
-	HostVolumes []HostMountedVolume
+	Metadata        map[string]string
+	Annotations     map[string]string
+	Labels          map[string]string
+	Containers      map[string]string
+	HostVolumes     []HostMountedVolume
+	AlreadyDeployed bool
```
nam-jaehyun: I think we don't need this

KubeArmor/feeder/feeder.go
oneiro-naut: Its a mistake. I will fix it. I confused Type with PolicyName. I think the conditions based on length are sufficient.

KubeArmor/core/kubeUpdate.go
oneiro-naut: I will remove this but then we need to call LogFeeder.UpdateSecurity irrespective of this flag. I added this to make sure that MatchPolicy struct was added for already deployed containers as well. Because for already deployed pods we don't get their modified event because we never patch them.

KubeArmor/core/kubeUpdate.go
oneiro-naut: We do need this one only without the condition that is. Its because if we remove this the already deployed container logs wont be able to get matched because their MatchPolicy struct wont get added if added policy event was received earlier than this function call. The reason is that for already deployed pods there is only ADDED event and no MODIFIED event. Our code today calls LogFeeder.UpdateSecurityPolicy only inside MODIFIED case.

KubeArmor/types/types.go
oneiro-naut: Sure thing. I will remove it.

KubeArmor/core/kubeUpdate.go
nam-jaehyun: It seems that it works without policy updates.
I ran like this.
start KubeArmor -> deploy multiubuntu -> stop KubeArmor -> start KubeArmor -> ./test-scenarios-in-runtime.sh

oneiro-naut: It does work but not always. At times it won't that happens when some pod added event is detected later than the added security policy event. In that case the containergroup for that pod is empty when the policy arrives so entry for them are not added but when the pod added event arrives it does not call the updatesecurity policy again so the logfeeder remains unaware of those containergroup matchpolicy structs. You can reproduce this behavior by repeating the above sequence. I added this line to cover that corner case. It will work only when struct containerGroup for a pod is updated before logfeeder.UpdateSecurityPolicy(action, conGroup) is called. So the call in ADDED makes sure that its added irrespective of the order. Btw I use the exact sequence to test already deployed containers. I think our test scripts cover all kinds of situations now with the addition of multiubuntu script.

PR #218, opened 2021-07-22
License Headers

daemon1024:


nam-jaehyun:
@daemon1024 can you also create a script to add the license headers if they do not exist?
there could be new files that someone creates. in this case, if we have a script, it would be nice to manage them I think.

daemon1024:
Sure @nam-jaehyun.

daemon1024:
I used https://github.com/google/addlicense to add the headers.
Should I commit the header file into some folder?

nam-jaehyun:
> I used https://github.com/google/addlicense to add the headers.
> Should I commit the header file into some folder?

What do you mean committing the header file into some folder?

daemon1024:
Like we have boilerplate in `pkg/KubeArmorPolicy/hack` Similarly the boilerplate header file for adding the headers via `addlicense`

nam-jaehyun:
@daemon1024 can you add the header file and a script file that simply executes addlicense with full arguments in "contribution/license"?

PR #219, opened 2021-07-22
Fix minor bug in RegisterAppArmorHostProfile

oneiro-naut:


KubeArmor/enforcer/appArmorEnforcer.go
```diff
@@ -468,7 +464,7 @@ func (ae *AppArmorEnforcer) UpdateAppArmorHostProfile(secPolicies []tp.HostSecur
 			return
 		}
 
-		if output, err := kl.GetCommandOutputWithErr("apparmor_parser", []string{"-r", "-W", "/tmp/kubearmor.host"}); err == nil {
+		if output, err := kl.GetCommandOutputWithErr("apparmor_parser", []string{"-r", "-W", "/etc/apparmor.d/kubearmor.host"}); err == nil {
```
nam-jaehyun, line 467: don't we need to remove line 472 too?
```
if err := ae.RemoveAppArmorHostProfile(); err != nil {
			ae.LogFeeder.Errf("Failed to remove the AppArmor host profile in %s (%s)", ae.HostName, err.Error())
			return
		}
```

oneiro-naut: Yes we do. I fixed that.

PR #220, opened 2021-07-22
eventAuditor: pattern-map, process-spec-map and process-filter-map management (userspace)

geyslan:
- [x] `Makefile` for compilation of a static `libbpf.a`
- [x] Create generic shared maps management logic
    - [x] Map loading/pinning (initialization)
    - [x] Map unloading/unpinning
- [x] Convert already commited code to use [kubearmor/libbpf](https://github.com/kubearmor/libbpf)
- [x] `pattern-map`, `process-spec-map` and `process-filter-map` tasks
  - [x] Implement the skeleton code for pattern-map (introduced by the last draft)
  - [x] Update pattern to the pattern-map from the userspace (introduced by the last draft)
  - [x] Implement the skeleton code for pattern-map, process-spec-map and process-filter-map management
  - [x] Update process spec to the process-spec-map from the userspace


KubeArmor/eventAuditor/eventAuditor.go
```diff
@@ -1,12 +1,14 @@
...
+import fd "github.com/kubearmor/KubeArmor/KubeArmor/feeder"
+
 // =================== //
 // == Event Auditor == //
 // =================== //
 
 // EventAuditor Structure
 type EventAuditor struct {
-	//
+	LogFeeder *fd.Feeder
```
nam-jaehyun, line 14: we will eventually use the log feeder later, but we don't need to think about it at this point.

KubeArmor/eventAuditor/sharedMapManager.go
```diff
@@ -1,8 +1,79 @@
...
+		if err != nil {
+			ea.LogFeeder.Printf(err.Error())
+			continue
+		}
+
+		bpfMod, err = lbpf.NewModuleFromFile(mapObjFilePath)
+		if err != nil {
+			ea.LogFeeder.Printf(err.Error())
+			continue
+		}
```
geyslan: Error Checking and Logging Logic.

It would be changed to:

```go
		if err != nil {
			ea.LogFeeder.Errf(err.Error())
			continue
		}
```
@nam-jaehyun, what do you think?

KubeArmor/eventAuditor/eventAuditor.go
geyslan: I see. I added it once we must stick to an error checking and logging logic, like the comment below. So I'll maintain this change for now. That's ok?

KubeArmor/eventAuditor/sharedMapManager.go
```diff
@@ -1,8 +1,146 @@
...
 // =========================== //
 
-// handle process-spec table
+// eventAuditorEBPFModule Structure
+type eventAuditorEBPFModule struct {
+	eaMod *lbpf.Module
+	eaMap *lbpf.BPFMap
+}
+
+var sharedMods map[string]*eventAuditorEBPFModule
```
geyslan: @nam-jaehyun 

BPFMap has a private pointer for the Module that contains it, but libbpfgo doesn't provide it for the consumer of the BPFMap. The solution that I found was to maintain the Map and its respective Module on our eventAuditorEBPFModule.

Why am I doing this? Well I think it's better to have an unique module for every single pinned map. This would make maintenance easier.

KubeArmor/eventAuditor/BPF/Makefile
```diff
@@ -0,0 +1,92 @@
+ifeq ($(V), 1)
```
nyrahul, line 18: If this Makefile is inspired by some existing Makefile/repo, please ensure we attribute that repo/project in the header. For e.g, 
```
This makefile is inspired by libbpf/bootstrap.
```
If there is license attribution needed, please add that too.

nyrahul (changes requested):
Attribution and license header check.

KubeArmor/eventAuditor/BPF/Makefile
geyslan: @nyrahul I don't really know how to determine the composition and creation of this Makefile. Let's see. Here we are using a libbpf submodule that needs to be compiled beforehand, so the Makefile will require rules for that. For the correct use of libbpf we'll need vmlinux.h, so we have to use bpftool and create some rules to create it and check the btf kernel file. Later, we'll need to compile the bpf objects that also require other rules. 

Has libbpf-bootstrap some similar steps? Yes it has (and it does have license).

https://github.com/libbpf/libbpf-bootstrap/blob/master/examples/c/Makefile

Has libbpfgo some similar steps? Yes it has (and it doesn't have license).

https://github.com/aquasecurity/libbpfgo/blob/main/Makefile

Has linux/tools/testing/selftests/bpf some similar steps? Yes it has (and it does have license).

https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/Makefile

I suppose that the steps **needed** to build the solution are not attached to the solution license, but let me know. Thanks.

nyrahul: The logic is plain and simple. If you did copy-paste verbatim from some existing Makefile then it is always best to attribute the source (whichever source you picked up from). It doesn't hurt to say, this Makefile was inspired from so and so Makefile.

geyslan: > It doesn't hurt to say, this Makefile was inspired from so and so Makefile.

Exactly. I totally agree. I just wasn't certain since this file was not copied-pasted verbatim, but inspired from various Makefiles (edit: and written from scratch). However, as you suggested I'll include the information about the inspirations and the related licenses to avoid further implications.

geyslan: Done.

KubeArmor/eventAuditor/sharedMapManager.go
geyslan: This workaround won't be necessary anymore. The https://github.com/aquasecurity/libbpfgo/pull/41 solved this question.

geyslan:
For testing:

`$ cd KubeArmor/eventAuditor/BPF`

`$ make run-tests`

PR #224, opened 2021-07-28
Change log action from Allow to Audit if its audit mode

### LINKED ISSUE #213: testcase #9 is always failed in minikube
With minikube, all test cases except for test case 9 are passed.
However, test case 9 is always failed; thus, we need to check the reason and fix this issue.

oneiro-naut:
This changes log action Allow to Audit (Allow) and Block to Audit (Block) if its audit mode.

Fixes: #213

### DIFF
--- a/KubeArmor/feeder/policyMatcher.go
+++ b/KubeArmor/feeder/policyMatcher.go
@@ -908,6 +908,14 @@ func (fd *Feeder) UpdateMatchedPolicy(log tp.Log) tp.Log {
 				return log
 			}
 		} else if log.Type == "MatchedPolicy" {
+			if log.PolicyEnabled == tp.KubeArmorPolicyAudited {
+				if log.Action == "Allow" {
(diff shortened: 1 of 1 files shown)


KubeArmor/feeder/policyMatcher.go
```diff
@@ -908,6 +908,14 @@ func (fd *Feeder) UpdateMatchedPolicy(log tp.Log) tp.Log {
 				return log
 			}
 		} else if log.Type == "MatchedPolicy" {
+			if log.PolicyEnabled == tp.KubeArmorPolicyAudited {
+				if log.Action == "Allow" {
+					log.Action = "Audit"
```
nam-jaehyun: let's keep the original action like the block action.
-> Audit (Allow)

oneiro-naut: Sure.

Issue #170, opened 2021-06-27
Github Action Tasks for KubeArmor


daemon1024:
- [x] Build
		`make build`
- [x] Code Coverage
	- codecov
- [x] CodeQL Analysis / GoSec
- [x] Go related Checks
	- [x] lint
	- [x] fmt
- [x] Testing
	- [x] Unit Test
			`make testall`
	- [x] Auto testing framework
			https://github.com/accuknox/KubeArmor/blob/master/contribution/testing_guide.md
		- [x] For different kernel versions
			Github Actions allows to specify ubuntu 18.04/ ubuntu 20.04 as such, I am not sure if we can configure kernel versions
- [x] File Headers

nam-jaehyun:
@daemon1024 Can you check ci-go.yml again?

daemon1024:
Hey @nam-jaehyun , I looked through the fails earlier. It's due to the missing go.sum file.
Since `go.sum` was in the .gitignore earlier there was no change introduced by `go mod tidy` but now since it's not there, due to the missing file, go mod tidy introduces a change and hence failing the test.

daemon1024:
I should detail out the error message more. I will send in a patch with go.sum and some refinements in ci, if it's fine with you?

nam-jaehyun:
I updated ci-go.yml a bit. Please check the file again and update it if some updates are needed

nam-jaehyun:
did you miss committing something?
![image](https://user-images.githubusercontent.com/67940721/127457989-2e3d9142-e3b8-4983-b696-07e44c8e87ad.png)

daemon1024:
Need `make build-test` before running the tests now.
This will generate the testable binary.

nam-jaehyun:
Is it possible to have a check logic or an option for that?
While we need the coverage in the GitHub action, we don't need that in a local development environment.

daemon1024:
Makes sense. I will modify the script accordingly.

PR #228, opened 2021-07-29
Add gosec, golint and gofmt workflows

daemon1024:
Ref #170


nam-jaehyun:
@daemon1024 will be merged as soon as we fix security things.

PR #236, opened 2021-08-04
Add automated test framework support for KubeArmorHostPolicy

### LINKED ISSUE #168: Support the automated test framework for KubeArmorHostPolicy
KubeArmor provides test scripts for KubeArmorPolicy.
Now, we need to support KubeArmorHostPolicy too.

oneiro-naut:
This adds KubeArmorHostPolicy support to the existing test framework. To run the host test scenarios hostpolicy must be enabled in kubearmor using the `-enableHostPolicy` option.

The commits also add skipped and reran testcase information to be shown in the test results and during the test script run.

Fixes: #168

### DIFF
--- a/tests/test-scenarios-in-runtime.sh
+++ b/tests/test-scenarios-in-runtime.sh
@@ -385,7 +385,14 @@ function run_test_scenario() {
     NATIVE=0
     HOST_POLICY=0
     NATIVE_HOST=0
-    if [[ $policy_type == "np" ]]; then
+
+    if [[ $policy_type == "ksp" ]]; then
+        if [ $SKIP_CONTAINER_POLICY == 1 ]; then
+            echo -e "${MAGENTA}[SKIP] Skipped $3${NC}"
+            skipped_testcases+=("$3")
+            return
+        fi
+    elif [[ $policy_type == "np" ]]; then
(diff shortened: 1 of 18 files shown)


tests/test-scenarios-in-runtime.sh
```diff
@@ -21,6 +21,11 @@ realpath() {
 
 TEST_HOME=`dirname $(realpath "$0")`
 
+SKIP_CONTAINER_POLICY=0
```
nam-jaehyun, line 24: If we merge this script, it tests host policies by default.
So, let's set the flags for container policies by default.
Then, similar to -EnableHostPolicy, why don't we have some options (arguments) for host policy testing?

oneiro-naut: It doesn't actually test the host policy and scenarios if we don't have the enableHostPolicy flag passed to kubearmor. It just skips them as of now. Do you mean we should set the SKIP_HOST_POLICY based on some argument passed to script rather than depending on -enableHostPolicy flag?

oneiro-naut: I am going to add one argument called -testHostPolicy to both test-scenarios-in-runtime.sh and test-scenarios-local.sh.

PR #237, opened 2021-08-04
Contribution Documentation improvements

geyslan:
This improve the contribution documentation.

- Development Guide
- Testing Guide

### DIFF
--- a/contribution/testing_guide.md
+++ b/contribution/testing_guide.md
@@ -34,7 +34,8 @@
         If it's still not applied, do so.
 
         ```text
-        $ kubectl apply -f ~/KubeArmor/pkg/KubeArmorPolicy/config/crd/bases/security.kubearmor.com_kubearmorpolicies.yaml
+        $ kubectl apply -f ~/KubeArmor/deployments/CRD/KubeArmorPolicy.yaml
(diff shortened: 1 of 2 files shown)


contribution/testing_guide.md
```diff
@@ -25,6 +25,20 @@
...
+        Beforehand, check if the KubeArmorPolicy or KubeArmorHostPolicy CRD is already applied.
+
+        ```text
+        $ kubectl explain KubeArmorPolicy
+        ```
+
+        If it's still not applied, do so.
+
+        ```text
+        $ kubectl apply -f ~/KubeArmor/pkg/KubeArmorPolicy/config/crd/bases/security.kubearmor.com_kubearmorpolicies.yaml
```
nam-jaehyun: Can you also add KubeArmorHostPolicy?
One more thing. Can you use /KubeArmor/deployments/CRD/... rather than /KubeArmor/pkg/...
those are the same things. Just for easy-to-use.

nam-jaehyun: done

geyslan: @nam-jaehyun, I was going back to it right now but thanks for doing these changes.

PR #242, opened 2021-08-09
update minimum docker version requirement to v18.09 and fix minor bugs

### LINKED ISSUE #223: Policies not enforced correctly when HostPolicy is enabled for docker version 18.06 and earlier
Some policies are not working correctly after hostpolicy is enabled in kubearmor. I think this might be some AppArmor issue.
This issue happens regardless of the kernel version. I tested this on both kubearmor-dev and kubearmor-dev-next.
Docker versions affected: versions older than 18.09 (tested on 18.03 and 18.06)
Steps to reproduce the issue:
CASE I: With HostPolicies enabled
1. Run kubearmor with -enableHostPolicy.
```
vagrant@kubearmor-dev-next:~/KubeArmor/KubeArmor$ make && make run-with-host
```
2. Deploy multiubuntu pods.
3. Apply sample policy [ksp-ubuntu-3-proc-dir-allow.yaml](https://github.com/kubearmor/KubeArmor/blob/master/examples/multiubuntu/security-policies/ksp-ubuntu-3-proc-dir-allow.yaml). (This policy allows binaries in directory /bin to be able to run)
4. Run some binary on the ubuntu-3-deployment pod from the /bin directory.
```
vagrant@kubearmor-dev-next:~$ kubectl exec -ti -n multiubuntu ubuntu-3-deployment-59bf5f5564-b2xxz -- /bin/bash
command terminated with exit code 129
vagrant@kubearmor-dev-next:~$ kubectl exec -n multiubuntu ubuntu-3-deployment-59bf5f5564-b2xxz -- /bin/bash
command terminated with exit code 139
vagrant@kubearmor-dev-next:~$ kubectl exec -n multiubuntu ubuntu-3-deployment-59bf5f5564-b2xxz -- /bin/echo "test"
command terminated with exit code 139
```
Observation I: The policy is not working since we are not able to run binaries in /bin directory. This is unexpected result.
5. Delete multiubuntu deployment.

CASE II: WIthout HostPolicies enabled
1. Run kubearmor without -enableHostPolicy.
```
vagrant@kubearmor-dev-next:~/KubeArmor/KubeArmor$ make && make run
```
2. Deploy multiubuntu pods.
3. Apply sample policy [ksp-ubuntu-3-proc-dir-allow.yaml](https://github.com/kubearmor/KubeArmor/blob/master/examples/multiubuntu/security-policies/ksp-ubuntu-3-proc-dir-allow.yaml). (This policy allows binaries in directory /bin to be able to run)
4. Run some binary on the ubuntu-3-deployment pod from the /bin directory.
```
vagrant@kubearmor-dev-next:~$ kubectl exec -ti -n multiubuntu ubuntu-3-deployment-59bf5f5564-cbcnw -- /bin/bash
bash: /etc/bash.bashrc: Permission denied
bash: /root/.bashrc: Permission denied
bash-4.4# echo "test"
test
bash-4.4# 
```
Observation II: The policy works since we are able to run binaries in /bin directory. This is expected result.

Ideally we should get the same result for both the cases .

Edit: The issue is version specific and does not happen for docker versions >= 18.09.

oneiro-naut:
Fixes: #223

### DIFF
--- a/.github/workflows/ci-kubeadm.yml
+++ b/.github/workflows/ci-kubeadm.yml
@@ -96,7 +96,7 @@ jobs:
         run: kubectl proxy &
 
       - name: Run Auto Testing Framework
-        run: ./tests/test-scenarios-local.sh -testAll -enableHostPolicy
+        run: ./tests/test-scenarios-local.sh
         timeout-minutes: 15
 
       - name: Archive log artifacts
--- a/.github/workflows/ci-kubeadm.yml
+++ b/.github/workflows/ci-kubeadm.yml
@@ -97,6 +97,7 @@ jobs:
 
       - name: Run Auto Testing Framework
         run: ./tests/test-scenarios-local.sh -testAll -enableHostPolicy
+        timeout-minutes: 15
 
       - name: Archive log artifacts
         if: ${{ failure() }}
--- a/.github/workflows/ci-kubeadm.yml
+++ b/.github/workflows/ci-kubeadm.yml
@@ -96,7 +96,7 @@ jobs:
         run: kubectl proxy &
 
       - name: Run Auto Testing Framework
-        run: ./tests/test-scenarios-local.sh
+        run: ./tests/test-scenarios-local.sh -testAll -enableHostPolicy
 
       - name: Archive log artifacts
         if: ${{ failure() }}
--- a/KubeArmor/main_test.go
+++ b/KubeArmor/main_test.go
@@ -6,10 +6,12 @@ package main
 import (
 	"flag"
 	"os"
+	"strconv"
 	"testing"
 )
 
-var clusterPtr, gRPCPtr, logPathPtr, enableHostPolicyPtr, enableEnforcerPerPodPtr *string
+var clusterPtr, gRPCPtr, logPathPtr *string
+var enableHostPolicyPtr, enableEnforcerPerPodPtr *bool
 
 func init() {
 	// options (string)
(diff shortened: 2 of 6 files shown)


.github/workflows/ci-kubeadm.yml
```diff
@@ -96,7 +96,7 @@ jobs:
         run: kubectl proxy &
 
       - name: Run Auto Testing Framework
-        run: ./tests/test-scenarios-local.sh
+        run: ./tests/test-scenarios-local.sh -testAll -enableHostPolicy
```
daemon1024:
```suggestion
        run: ./tests/test-scenarios-local.sh -testAll -enableHostPolicy
        timeout-minutes: 15
```
We should timeout the test run after 15 minutes to avoid it running infinitely. ( Default timeout is 360 )

KubeArmor/main_test.go
```diff
@@ -28,6 +30,6 @@ func TestMain(t *testing.T) {
 	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 
 	// Set os args to set flags in main
-	os.Args = []string{"cmd", "-cluster", *clusterPtr, "-gRPC", *gRPCPtr, "-logPath", *logPathPtr, "-enableHostPolicy", *enableHostPolicyPtr, "-enableEnforcerPerPod", *enableEnforcerPerPodPtr}
+	os.Args = []string{"cmd", "-cluster", *clusterPtr, "-gRPC", *gRPCPtr, "-logPath", *logPathPtr, "-enableHostPolicy", strconv.FormatBool(*enableHostPolicyPtr), "-enableEnforcerPerPod", strconv.FormatBool(*enableEnforcerPerPodPtr)}
```
nam-jaehyun, line 33: is "strconv.FormatBool" required? those are already the bool types, aren't they?

daemon1024: We are converting bool to string. 
We need it to be string since [`os.Args`](https://pkg.go.dev/os#Args) is a string array.

daemon1024: But we need the flag to be boolean, since if we set it to string it's behavior changes and it demands an argument.

nam-jaehyun: then, it should be something like the below since the arguments will be parsed by the main function, shouldn't it?
os.Args = []string{"cmd", "-cluster", *clusterPtr, "-gRPC", *gRPCPtr, "-logPath", *logPathPtr, "-enableHostPolicy", "-enableEnforcerPerPod"}

daemon1024: `-enableHostPolicy` == `-enableHostPolicy true`

We will have to add if checks to set it to false, or not add it to the string array otherwise.

PR #243, opened 2021-08-09
add exec rules for fromSource

daemon1024:


tests/scenarios/multiubuntu_test_12/cmd2
```diff
@@ -0,0 +1,7 @@
+source: ubuntu-3-deployment
+cmd: cmd: bash -c ls
```
nam-jaehyun: cmd: cmd:

nam-jaehyun: if we directly run 'sleep', it should work. right?
can you add that as cmd3?

oneiro-naut:
@daemon1024 The exec qualifier would provide an unnecessary exec permission in the case when fromSource is non-executable(like it can be some directory). I think `ix` should only be given for the case when the source is an executable itself. And it should stay `r` in other cases. To make sure whether a source is executable or a directory I think we should first check its current fileMode bits(using `os.Stat()` maybe) and accordingly emit the rule. Just to make sure that rules emitted are strict.

Issue #269, opened 2021-08-17
eventAuditor: main Makefile needs prior logic for cgo compilation


geyslan:
`CGO_CFLAGS` and `CGO_LDFLAGS` need to be set for the kubearmor/libbpf building process.

Before that, C libbpf needs to be available (cloned and compiled).

WIP

PR #270, opened 2021-08-17
event-auditor: KubeArmor: update Makefiles for CGO compilation

geyslan:
This also updates go.mod and go.sum.

Fixes: #269


geyslan:
I think the main `Makefile` deserves a refactoring, related to the rules and recipes. Let's also discuss to where we could move the eventAuditor/libbpf, as it must be used later throughout the project.

Issue #272, opened 2021-08-18
event-auditor: conflicting VM_NAME in vagrant file


geyslan:
`VM_NAME` must refer to a single vagrant image.

`NETNEXT=1` and `NETNEXT=2` are conflicting.

PR #273, opened 2021-08-18
event-auditor: contribution: fix VM_NAME

geyslan:
Fixes: #272

### DIFF
--- a/contribution/vagrant/Vagrantfile
+++ b/contribution/vagrant/Vagrantfile
@@ -2,10 +2,10 @@ Vagrant.require_version ">= 2.0.0"
 
 if ENV['NETNEXT'] == "1" then
   VM_IMG = "ubuntu/impish64"
-  VM_NAME = "kubearmor-dev-next"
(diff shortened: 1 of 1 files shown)


contribution/vagrant/Vagrantfile
```diff
@@ -2,10 +2,10 @@ Vagrant.require_version ">= 2.0.0"
 
 if ENV['NETNEXT'] == "1" then
   VM_IMG = "ubuntu/impish64"
-  VM_NAME = "kubearmor-dev-next"
+  VM_NAME = "kubearmor-dev-next-1"
 elsif ENV['NETNEXT'] == "2" then
   VM_IMG = "ubuntu/focal64"
-  VM_NAME = "kubearmor-dev-next"
+  VM_NAME = "kubearmor-dev-next-2"
```
nyrahul, line 8: Is it possible to use both the VMs in parallel on the same host with this change?

nyrahul:
LGTM. I guess we can use both the VMs in parallel on the same host with this change.

PR #279, opened 2021-08-23
Add len check in readByteSliceFromBuff

oneiro-naut:


nam-jaehyun:
@oneiro-naut if the length check is failed, what would be the expected output?
For example, dropping the current event? or building a log structure without certain strings?

oneiro-naut:
> @oneiro-naut if the length check is failed, what would be the expected output?
> For example, dropping the current event? or building a log structure without certain strings?

@nam-jaehyun Currently we completely drop the buffer whenever we get some error here.
https://github.com/kubearmor/KubeArmor/blob/309713f0e8a34d50b93b354bfe5bac9922a6d75b/KubeArmor/monitor/systemMonitor.go#L447
We just skip it. I also thought of finding a way of preserving some partially overwritten buffer but it seems that it's not possible.

PR #338, opened 2021-09-08
add sample audit policies

Ankurk99:
Signed-off-by: Ankur Kothiwal <ankur.kothiwal@accuknox.com>


nam-jaehyun (changes requested):
selector is missing.
please check these audit policies with the latest CRD.

nam-jaehyun (approved):

Issue #315, opened 2021-09-02
KVM Agent - Setting up GRPC Connection to the KVMService


nandhued:
_Start-Date:_ 09/02/2021
_Effort Estimate:_ **1.25**
- [x] Design - 0.25
- [x] Coding - 0.5
- [ ] Unit Test - 0.5

PR #339, opened 2021-09-08
#315 -- Support for nonk83 external workload

seswarrajan:
Enabling KvmAgent  as part of Kubearmor process. 
KVMAgent is a service which runs as a part of kubearmor for ext-non-k8s-workload .

To enable the same follow the below command 

./kubearmor --enableKvmAgent true #This will enable kvm agent and by default the value is false and kvm agent will not run


KubeArmor/kvmAgent/KvmAgent.go
```diff
@@ -0,0 +1,51 @@
...
+		Identity: os.Getenv("USER"),
+	}
+
+	client = pb.NewKVMClient(grpcClientConn)
+	response, err := client.RegisterAgentIdentity(context.Background(), &identity)
+	if err != nil {
+		log.Println("Failed to register identity", err)
+		return err
+	} else {
+		fmt.Println("Response is ", response.Status, response.ErrorMessage)
```
nyrahul: use `log.Printf`

KubeArmor/kvmAgent/KvmAgent.go
```diff
@@ -0,0 +1,51 @@
...
+	identity := pb.AgentIdentity{
+		Identity: os.Getenv("USER"),
+	}
+
+	client = pb.NewKVMClient(grpcClientConn)
+	response, err := client.RegisterAgentIdentity(context.Background(), &identity)
+	if err != nil {
+		log.Println("Failed to register identity", err)
+		return err
+	} else {
```
nyrahul: remove `else` since returning in if

KubeArmor/main.go
```diff
@@ -42,12 +42,13 @@ func main() {
 	// options (boolean)
 	enableHostPolicyPtr := flag.Bool("enableHostPolicy", false, "enabling host policies")
 	enableEnforcerPerPodPtr := flag.Bool("enableEnforcerPerPod", false, "enabling the enforcer per pod")
+	enableKvmAgent := flag.Bool("enableKvmAgent", false, "Flag to enable/disable KVM Agent on runtime")
```
nyrahul: let's use pothole case for command-line arguments?
For e.g.
`enable_kvma`
`enable_enforcer_per_pod`
`enable_host_policy`

KubeArmor/main.go
```diff
@@ -42,12 +42,13 @@ func main() {
 	// options (boolean)
 	enableHostPolicyPtr := flag.Bool("enableHostPolicy", false, "enabling host policies")
```
nyrahul, line 43: This should be default `true` for KVMA.

nyrahul (changes requested):

KubeArmor/kvmAgent/KvmAgent.go
seswarrajan: Sure Rahul. This is just for dev purpose to make sure the connection is successful. Will remove it.

KubeArmor/main.go
seswarrajan: Need to change the existing enable flags as well ? Pothole looks great and readable. Can finalise as per comments from others as well.

KubeArmor/main.go
seswarrajan: This can be provided as a param in script for kvma. I suggest not to change the existing value. Others also please suggest

KubeArmor/core/kubeArmor.go
```diff
@@ -279,7 +280,7 @@ func GetOSSigChannel() chan os.Signal {
 // ========== //
 
 // KubeArmor Function
-func KubeArmor(clusterName, gRPCPort, logPath, logFilter string, enableHostPolicy, enableEnforcerPerPod bool) {
+func KubeArmor(clusterName, gRPCPort, logPath, logFilter string, enableHostPolicy, enableEnforcerPerPod bool, enableKvmAgent bool) {
```
nam-jaehyun, line 283: can you update these on the latest code?

KubeArmor/main.go
nam-jaehyun: enableEnforcerPerPod was removed while enableKubeArmoPolicy and enableKubeArmorHostPolicy were added.
let's update these on the latest code.

KubeArmor/core/kubeArmor.go
seswarrajan: Sure @nam-jaehyun .
Working on this review comment.

PR #351, opened 2021-09-14
fix ownerOnly behaviour

### LINKED ISSUE #300: ownerOnly doesn't work in Ubuntu 18.04
## Bug Report

**General Information**

- Environment description

Checked this issue from 'kubearmor-dev' (vagrant, ubuntu 18.04)

- Kernel version

4.15.0-151-generic

- Orchestration system version in use

Kubernetes version: 1.21.4

**To Reproduce**

Assume that KubeArmor is already running.

1. Apply wordpress-mysql

```
$ kubectl apply -f KubeArmor/examples/wordpress-mysql
```

2. Apply test.yaml

```
$ vi test.yaml

kind: KubeArmorPolicy
metadata:
  name: ksp-wordpress-block-config
  namespace: wordpress-mysql
spec:
  severity: 10
  selector:
    matchLabels:
      app: wordpress
  file:
    matchPaths:
    - path: /var/www/html/wp-config.php
      ownerOnly: true

      # http://[NodeIP]:30080
      # cat /var/www/html/wp-config.php
 
  action:
    Block

$ kubectl apply -f test.yaml
```

3. Get into the WordPress pod and run the following commands

```
$ kubectl -n wordpress-mysql exec -it [WordPress pod] -- bash

root@wordpress-dbcd8b97b-pwfmc:/var/www/html# ls -l wp-config.php
-rw-r--r-- 1 www-data www-data 3151 Aug 27 10:21 wp-config.php
```

Here, the owner of wp-config.php is 'www-data', not 'root'.

- Expectation

```
root@wordpress-dbcd8b97b-pwfmc:/var/www/html# cat /var/www/html/wp-config.php
cat: wp-config.php: Permission denied

<Alert>

http://[NodeIP]:30080 -> can access the Wordpress blog
```

- Reality

```
root@wordpress-dbcd8b97b-pwfmc:/var/www/html# cat /var/www/html/wp-config.php
...content...

<No Alert>
```

**Expected behavior**

'ownerOnly' means that the owner (Linux permission system) of the file can only execute the file; otherwise, all should be blocked. However, it seems that the 'ownerOnly' option doesn't work on Ubuntu 18.04.

This option worked previously. I guess that this option may not work according to AppArmor versions.

daemon1024:
ownerOnly didn't work as intended due to additional "file" rule. We add a deny rule for "other" users for the target match so as to make it work as  while having the "file" rule at same time

Fix #300

### DIFF
--- a/KubeArmor/enforcer/appArmorProfile.go
+++ b/KubeArmor/enforcer/appArmorProfile.go
@@ -20,7 +20,7 @@ import (
 func allowedProcessMatchPaths(path tp.ProcessPathType, processWhiteList *[]string, fromSources map[string][]string) {
 	if len(path.FromSource) == 0 {
 		if path.OwnerOnly {
-			line := fmt.Sprintf("  owner %s ix,\n", path.Path)
+			line := fmt.Sprintf("  owner %s ix,\n  deny other %s x,\n", path.Path, path.Path)
 			if !kl.ContainsElement(*processWhiteList, line) {
 				*processWhiteList = append(*processWhiteList, line)
 			}
@@ -56,7 +56,7 @@ func allowedProcessMatchPaths(path tp.ProcessPathType, processWhiteList *[]strin
 			}
 
 			if path.OwnerOnly {
-				line := fmt.Sprintf("  owner %s ix,\n", path.Path)
+				line := fmt.Sprintf("  owner %s ix,\n  deny other %s x,\n", path.Path, path.Path)
 				if !kl.ContainsElement(fromSources[source], line) {
 					fromSources[source] = append(fromSources[source], line)
 				}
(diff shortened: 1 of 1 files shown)


nam-jaehyun:
as far as I know, 'others' works from a specific version and above. where did you confirm this PR?
Do you think that this update will solve the 'ownerOnly' issue on the minimum environment (ubuntu 18.04 with AppArmor 2.12)?

daemon1024:
Yes it works inn the following environment.

```
> dpkg -l apparmor | tee
...
ii  apparmor       2.12-4ubuntu5.1 amd64        user-space parser utility for AppArmor

> uname -r
4.15.0-156-generic
```

PR #374, opened 2021-09-19
KVM Service - Protobuf Definitions for the message exchanges across KEWS and KEWA

seswarrajan:


seswarrajan:
Added wrong files in pull request. Creating new pull request again.

PR #381, opened 2021-09-20
event-auditor: code generation

chkhalt:
Event Auditor Code Generation:

Priority:
- [x] Argument parsing (numeric values)
- [x] Avoid to parse common rules multiple times - `EventCodeBlockCache`
- [x] Avoid to rebuild the same eBPF code - `EventProgramCache`
- [x] Populate `ka_ea_event_jmp_table` (progfd for tailcall) 
#419 
- [x] Populate `ka_ea_event_filter_map` (index of the target program on `ka_ea_event_jmp_table`)
#419 
- [x] Implement the skeleton functions on `codegen.bpf.h` #456 

Extra:
- [x] Generate `runtime.h` with the necessary macros to allow referencing loaded maps
- [ ] Garbage collect old (unused) eBPF auto generated programs
- [ ] Allow macros to be defined in terms of another macros

### DIFF
--- a/KubeArmor/BPF/codegen.bpf.h
+++ b/KubeArmor/BPF/codegen.bpf.h
@@ -1,0 +1,280 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+/* Copyright 2021 Authors of KubeArmor */
+
+#ifndef __CODEGEN_BPF_H
+#define __CODEGEN_BPF_H
+
+#include "runtime.h"
+#include "maps.bpf.h"
+#include "common.bpf.h"
+
+union v4addr {
+	__u32 d1;
+	__u8 octet[4];
+} __packed;
+
+union v6addr {
+	struct {
+		__u32 d1;
+		__u32 d2;
+		__u32 d3;
+		__u32 d4;
+	};
+	__u8 octet[16];
+} __packed;
+
+struct syscalls_enter_execve_args {
+	unsigned long long unused;
+	long syscall_nr;
+	long filename;
+	long argv;
+	long envp;
+};
+
+struct syscalls_enter_execveat_args {
+	unsigned long long unused;
+	long syscall_nr;
+	long fd;
+	long filename;
+	long argv;
+	long envp;
+	long flags;
+};
+
+struct syscalls_enter_open_args {
+	unsigned long long unused;
+	long syscall_nr;
+	long filename;
+	long flags;
+	long mode;
+};
+
+struct syscalls_enter_openat_args {
+	unsigned long long unused;
+	long syscall_nr;
+	long fd;
+	long filename;
+	long flags;
+	long mode;
+};
+
+struct syscalls_enter_socket_args {
+	unsigned long long unused;
+	long syscall_nr;
+	long family;
+	long type;
+	long protocol;
+};
+
+struct syscalls_enter_connect_args {
+	unsigned long long unused;
+	long syscall_nr;
+	long fd;
+	long uservaddr;
+	long addrlen;
+};
+
+struct syscalls_enter_accept_args {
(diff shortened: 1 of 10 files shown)


KubeArmor/BPF/codegen.bpf.h
```diff
@@ -0,0 +1,285 @@
...
+	struct process_filter_value *value;
+
+	task = (struct task_struct *) bpf_get_current_task();
+	key.pid_ns = BPF_CORE_READ(task, nsproxy, pid_ns_for_children, ns).inum;
+	key.mnt_ns = BPF_CORE_READ(task, nsproxy, mnt_ns, ns).inum;
+	key.host_pid = (u32) bpf_get_current_pid_tgid();
+
+	value = bpf_map_lookup_elem(&ka_ea_process_filter_map, &key);
+	return (value && value->inspect);
+}
```
geyslan: What do you think to put part of this "get task info" logic inside `shared.h` (this file can be renamed to common.h). This would make possible its reuse for many bpf programs to come.

KubeArmor/BPF/maps.bpf.h
```diff
@@ -45,4 +45,11 @@ struct event_filter_value {
 	__u32 jmp_idx;
 };
 
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__type(key, struct process_filter_key);
+	__type(value, struct process_filter_value);
+	__uint(max_entries, 1 << 10);
+} ka_ea_process_filter_map SEC(".maps");
+
```
geyslan: We've discussed about this and I agreed with you to put the bpf map structure in a common place. But I was wrong since any bpf object loaded which include this header will raise a new copy in memory of all the maps that lies inside it.

My suggestion is to maintain inside `maps.bpf.h` the common structures used for keys and values, since it could be imported by cgo directives to be used in Go source files, and let the bpf map structure initialization - SEC(".maps") - inside `map_name_bpf.h` or in the same `prog.bpf.c` for the sake of clarity.

Another possible approach (not tested by me) would be to define the bpf map structure inside the common `maps.bpf.h` and let the initialization - SEC(".maps") - happen after the header inclusion.

What do you think?

geyslan (approved):
@johngv2 I've dropped a few words for discussion.

geyslan (changes requested):
Requesting changes.

KubeArmor/BPF/codegen.bpf.h
chkhalt: mmm, good point! I'm gonna move the code.

chkhalt: @geyslan, I moved the common part declaring some functions in `common.bpf.h`.
The file `shared.h` is imported in the beginning of `sharedMaps.go`, so I wasn't able to define too much things inside it (include `vmlinux.h`, for example) wihout compilation error. Thus, for now, the code in `shared.h` still the same.

What you think?

KubeArmor/BPF/maps.bpf.h
chkhalt: @geyslan, @it-with-us. I'm including some alternative logic to reuse loaded maps in the commit cebdf2661b23d8258fda5cf7814135a0a7715491, if you prefer we can make this part of another PR. What you think?

KubeArmor/BPF/codegen.bpf.h
geyslan: LGTM.

geyslan:
I've put a test here for the dynamic (runtime) map reuse from bpf side: https://github.com/geyslan/libbpfgo/tree/main/selftest/ringbuffers-duplicate

PR #388, opened 2021-09-21
Add more sample file security policies

oneiro-naut:
This PR adds sample security policies to multiubuntu. The policies target file path.


examples/multiubuntu/build/Dockerfile
```diff
@@ -32,5 +32,7 @@ RUN useradd --create-home --shell /bin/bash user1
 RUN echo 'user1:passwd1' | chpasswd
 COPY hello /home/user1/hello
 RUN chown user1:user1 /home/user1/hello
+COPY secret.txt /home/user1/secret_data1.txt
```
nam-jaehyun: so, we need to rebuild the multiubuntu image, right?

oneiro-naut: Yes some yaml files rely on this. We have a binary so it would be good to add a simple text file owned by user1.

PR #391, opened 2021-09-21
Get unmarshalled embedded crd

daemon1024:
Added function to provide generated CRD which is embedded at compile time


elfadel:
@daemon1024 does this PR fix any specific issue? Or is it an improvement?

daemon1024:
This is an improvement.

Context for PR. 
We need to apply resources through our https://github.com/kubearmor/kubearmor-client/pull/5

We have resources other than CRD defined here, https://github.com/kubearmor/kubearmor-client/blob/1d41ba98a0bb335d0f6e768fc077ca89193082cc/install/resource.go

This PR will provide a function to facilitate CRD definition. Since CRD yamls are generated using `controller-gen` It wouldn't be useful to store a seperate copy in client repo hence the function to provide that.

daemon1024:
I have implemented things based on this PR using replace directive.

https://github.com/daemon1024/kubearmor-client/commit/fde34f2f012eda3710749c0ecd9ab8589f321ea2 should showcase the usecase of what this PR adds.

elfadel (changes requested):
Overall, LGTM. One comment, however:
`pkg/KubeArmorHostPolicy/crd/KubeArmorHostPolicy.yaml` and `pkg/KubeArmorPolicy/crd/KubeArmorPolicy.yaml` are the same under `deployments/CRD/`. Please, remove this redundancy; unless there is any specific reason to proceed as you do now.

daemon1024:
`deployment/CRD` is outside the scope of the package so I couldn't embed them. `go:embed` can only refer to files in current or sub directories.
Hence the added redundancy.

elfadel (approved):

elfadel:
@nyrahul can you merge? I don't have write access.

PR #313, opened 2021-09-01
event-auditor: ebpf: load and unload process events

geyslan:
- [x] Load ebpf programs #340
- [x] Maintain information of the ebpf object files for reuse (bpf programs)  in `KABPFManager` #340
- [x] Implement `ka_ea_sched_process_exec` and `ka_ea_sched_process_exit` ebpf programs logic.

**More information in** https://github.com/kubearmor/KubeArmor/pull/313/commits/dd55bbc7fb3e11b38fcd7f0fc2611f52142bd90a.

Fix: #281

PR #397, opened 2021-09-26
event-auditor: introduce match() for patterns

geyslan:
This PR (continuation of #313) brings a lot of changes attempting to shrink a basic match()
function into the ka_ea_process.bpf.c program.

match() treats only the '*' and '?' wildcards. Its algorithm is inspired by
two sources [1-2], being modified to satisfy the ebpf verifier when it comes
to bounded loops.

Despite the algorithm design undertaken, the verifier still complained of
1 million instructions being exceeded.

Two decisions were made:

- reduce iterations (reducing MAX_FILENAME_LEN from to 256 to 42 and
  MAX_PATTERN_LEN (former PATTERN_MAX_LEN) from 16 to 7.
- split the ebpf program logic into 3 making use of tail calls,
  leaving one entirely for the match() logic.

After these, the verifier didn't complain.

The first decision was taken as a temporary measure giving rise to
questions about the design:

- Would '#pragma unroll n' be a solution? (this committer was unable
  to find a magic n)
- Should we take another route for pattern matching?

This commit also adds:

- ka_ea_process_jmp_map to ka_ea_process.bpf.c
- global variables accessible from the tail programs through .bss map
- get_task_filename - program logic split
- get_task_ids() - program logic split
- task_auditable_0() and task_auditable_1() - program logic split
- ka_ea_sched_process_exec_0() and ka_ea_sched_process_exec_1() - program
  logic split
- check_hash_elem() - handler for bpf_for_each_map_elem() iterator
- callback_ctx struct - input/output for the check_hash_elem()

- KABPFTailProg struct to ebpfCommon.go - and put it as a KABPFProg slice field
- ProcessJMPMapElement struct and methods to sharedMaps.go
- PopulateProcessJMPMap() to processSpecHandler.go

This gets rid of:

- task_context struct
- get_task_context()

[1] https://github.com/kirkjkrauss/MatchingWildcards/blob/master/Listing1.cpp
[2] http://dodobyte.com/wildcard.html

### DIFF
--- a/KubeArmor/BPF/ka_ea_process.bpf.c
+++ b/KubeArmor/BPF/ka_ea_process.bpf.c
@@ -8,6 +8,13 @@
 #include "maps.bpf.h"
 
 struct {
+	__uint(type, BPF_MAP_TYPE_PROG_ARRAY);
+	__type(key, u32);
+	__type(value, u32);
+	__uint(max_entries, 2);
+} ka_ea_process_jmp_map SEC(".maps");
+
+struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__type(key, struct pattern_key);
 	__type(value, struct pattern_value);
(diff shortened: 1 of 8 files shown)


geyslan:
Addendum:

In some cases, after deletion of some code not related to the bounded loop and its tail program, the verifier gone crazy showing an error about not allowed memory accessed, forcing me to find other magic number for the constants involved in the bounded loop iterations. This is not helpful indeed.

Issue #86, opened 2021-02-21
Policy enforcement failure due to rule conflicts


nam-jaehyun:
Let's say that we have the following policies.

Policy A
```
process:
  matchPaths:
  - path: /bin/sleep
     fromSource:
     - path: /bin/bash
  action:
    Allow
```

Policy B
```
process:
  matchPaths:
  - path: /bin/bash
action:
  Allow
```

Those policies are converted something like this.

```
...
/bin/bash cx,
profile /bin/bash {
  /bin/sleep ix,
}
/bin/bash ix,
...
```

Causing the policy enforcement failure because of two /bin/bash lines.

daemon1024:
Hi, I am working on this. Could this issue be assigned to me?

nam-jaehyun:
@daemon1024 I assigned this issue to you.

daemon1024:
Hey, I think have got an understanding of the problem.

https://github.com/accuknox/KubeArmor/blob/38fc028685a3864f4485c05ff65ec78962f5eca6/KubeArmor/enforcer/appArmorProfile.go#L15

Currently we check `paths` and `FromSource paths` independently, we need to check such that sources shouldn't have any conflicts with processes white/blacklists.

In the example specified in the description, the final construct should be something like
```
/bin/bash cix,
profile /bin/bash {
  /bin/sleep ix,
}
```

Let me know if I am headed into the right direction.

nam-jaehyun:
@daemon1024 Yes, that was what I thought, but I'm wondering if there is no problem when we have such AppArmor rules.
If you validate those AppArmor rules before starting the implementation, it would be helpful.
Please check if there is no syntax and semantic error.

nam-jaehyun:
@daemon1024 any update on this issue?

daemon1024:
Hey @nam-jaehyun, I have been working on #170 so haven't made any progress here.

nam-jaehyun:
@daemon1024 no problem :) just for checking.

daemon1024:
Not blocking this if someone else wants to take it up meanwhile. I believe I have already been unassigned on the xlsx.

daemon1024:
>I'm wondering if there is no problem when we have such AppArmor rules.

@nam-jaehyun It does error out,

AppArmor does have issues with rule conflicts,

`profile kubearmor-multiubuntu-ubuntu-1-container: has merged rule /bin/bash with conflicting x modifiers
[...]

PR #405, opened 2021-09-29
Fix rules conflict

elfadel:
This PR should fix issue #86.

- PR overview:
In the current code, `processWhiteList` contains _paths_ that do not have nested _fromSource paths_ whereas `fromSource` contains _paths_ with nested _fromSource paths_.
The apparmor profiles rules are generated from above variables: `processWhiteList` and `fromSource`. 
For _path_ in `processWhiteList`, the generated policy is: `path ix` whereas for path in `fromSource`, the generated policy is : 
``` 
path cx
profile path {
    ...
}
```
Therefore if there is at least one duplicate path in `processWhiteList` and `fromSource`, the generated apparmor profile contains conflicts rule because `ix` and _cx_ are incompatibles.

Then, the basic idea is the following:
Once the `processWhiteList` and `fromSource` are build from yaml policies, we apply `resolvedProcessWhiteListConflicts()` function. This function will build from `processWhiteList` and `fromSource` a `fusionList` that contains paths that are duplicated and removed from `processWhiteList` those paths => `path ix` will not be generated anymore. From fusionList, we will generated `path cix` instead of `path cx`.

Finally:
```
path ix     
path cx
profile path {
        ...
}
```
will be replaced by:
``` 
path cix
profile path {
        ...
}
```


elfadel:
* request review *
@nyrahul @nam-jaehyun

PR #413, opened 2021-10-07
Update README.md

namdeirf:
Under the ## Community section, it lists a 'Weekly Meetup' but in the document, it is listed as 'every other Thursday...'. I recommend changing to 'Biweekly', or the less specific 'Regular' or 'Official'

### Biweekly Meetup

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ If you want to make a contribution, please refer to the following documents too.
 
 ## Community
 
-### Weekly Meetup
+### Biweekly Meetup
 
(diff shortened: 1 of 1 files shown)


nyrahul:
The ICS file also needs modification.
Attached is the modified ICS. Can you update [the existing ICS](https://github.com/kubearmor/KubeArmor/blob/master/getting-started/resources/KubearmorMeetup.ics) with the attached one in this PR?
[KubeArmor.zip](https://github.com/kubearmor/KubeArmor/files/7315886/KubeArmor.zip)(Using zip since ICS file is not allowed to be attached in GH).

Issue #416, opened 2021-10-10
Implement process-spec-update in the user space


geyslan:
- Extract the process specs from audit policies
- Maintain the process specs (list)
- Generate unique values for each process value
- Update the process specs into eBPF maps (process-spec-map, pattern map)
- If some process specs are removed from audit policies, delete the corresponding keys from eBPF maps

PR #417, opened 2021-10-10
event-auditor: process-spec and pattern maps update from user space

geyslan:
This introduces updatePatternMap(), updateProcessSpecMap() and getProcessElements() helper functions called from UpdateProcessMaps() to EventAuditor.

The whole logic take care of:

- Extracting the process specs from audit policies
- Maintaining the process specs (list)
- Generating unique values for each process value
  The algorithm used (32-bit FNV-1 hash) can be changed in the future to a simpler implementation.
- Updating the process specs into eBPF maps (process-spec-map, pattern map)
- If some process specs are removed from audit policies, deleting the corresponding keys from eBPF maps

Fixes: #416

This also fix naming and comments and updates go.mod.


geyslan:
@nam-jaehyun During the implementation tests I noticed that the `dm.EndPoints` passed to `dm.EventAuditor.UpdateProcessMaps(&dm.Containers, &dm.ContainersLock, &dm.EndPoints, &dm.EndPointsLock)` don't always contain all audit policies applied, fact that deserves investigation.

PR #428, opened 2021-10-13
Add policy name parameter to log matching grep command in test script

oneiro-naut:
Current log matching grep command uses $POD $OP $COND $ACTION. These parameters aren't enough to identify some log.
There are cases when two or more logs can be having same parameter values.
This commit adds $POLICY parameter which is for policy name to make log matching more strict and accurate.

### DIFF
--- a/tests/test-scenarios-local.sh
+++ b/tests/test-scenarios-local.sh
@@ -195,7 +195,7 @@ function should_not_find_any_log() {
 
     sleep 3
 
-    audit_log=$(grep -E "$1.*MatchedPolicy.*$2.*resource.*$3.*$4" $ARMOR_LOG | tail -n 1 | grep -v Passed)
+    audit_log=$(grep -E "$1.*policyName.*$2.*MatchedPolicy.*$3.*resource.*$4.*$5" $ARMOR_LOG | tail -n 1 | grep -v Passed)
     if [ $? == 0 ]; then
         if [ "$audit_log" == "$LAST_LOG" ]; then
             audit_log="<No Log>"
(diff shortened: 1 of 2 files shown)


tests/test-scenarios-local.sh
```diff
@@ -347,7 +347,7 @@ function run_test_scenario() {
 
     YAML_FILE=$(ls *.yaml)
     policy_type=$(echo $YAML_FILE | awk '{split($0,a,"-"); print a[1]}')
-    
+    POLICY=$(echo $YAML_FILE | awk '{split($0,a,".yaml"); print a[1]}')
```
nam-jaehyun: we may have different names (file name != policy name).

nam-jaehyun:
@oneiro-naut you also need to apply your changes for matching host policies.
Also, you need to update them in test-scenarios-in-runtime.sh.

tests/test-scenarios-local.sh
oneiro-naut: then can we use something like `kubectl get ksp -n $namespace` to get the policy name?

oneiro-naut:
> @oneiro-naut you also need to apply your changes for matching host policies. Also, you need to update them in test-scenarios-in-runtime.sh.

@nam-jaehyun Sure thing.

tests/test-scenarios-local.sh
dku-boanlab: we can read "name: xxx" from the $YAML_FILE.

oneiro-naut: Yeah we can do that. I will make these changes.

Issue #430, opened 2021-10-15
Pattern IDs match using hash


geyslan:
Generate hash pattern IDs based on the given `char[]` in the eBPF code.

PR #431, opened 2021-10-15
event-auditor: pattern match using hash

geyslan:
- [x] implement hash function for string (jenkins) - ebpf
- [x] change `ka_ea_process.bpf.c` to use hashes as map keys
- [x] ~~implement~~ import hash function for string (jenkins) - user space
- [x] change user space to use hashes as map keys

Fixes: #430


geyslan:
@nam-jaehyun I decided not to implement the jenkins inside the kubearmor for now as we already have the leb.io/hashland library available.

PR #434, opened 2021-10-18
[VM] Ungraceful shutdown of kubearmor and reboot of the machine, VM is not booting up

### LINKED ISSUE #409: [VM] Ungraceful shutdown of kubearmor and reboot of the machine, VM is not booting up
**Environment**
- Baremetal/VirtualMachine
- Ubuntu 20.04 with kernel version 5.11.0-37-generic

**To Reproduce**
1. Build kubearmor in a host machine/VM.
2. Run kubearmor with option **sudo ./kubearmor -enableKubeArmorHostPolicy true**
3. Now the kubearmor runs in VM and creates a new profile file (kubearmor.host) under /etc/apparmor.d/
4. Now close the kubearmor abruptly (not a graceful exit). 
5. Since step 4 is carried out, kubearmor is not gracefully shutdown and also the file (/etc/apparmor.d/kubearmor.host) is not removed.
6. Reboot the VM.

**Expected behavior**
In this case, the VM is not booting up properly. 

**Actual behavior**
The actual behavior is that the VM should boot up with no issue.

seswarrajan:
After applying policies to VMs/Baremetals/Pods, when KubeArmor shuts down unfgracefully, the VM/pod fails to bootup on reboot.

### DIFF
--- a/KubeArmor/enforcer/appArmorEnforcer.go
+++ b/KubeArmor/enforcer/appArmorEnforcer.go
@@ -508,6 +508,13 @@ func (ae *AppArmorEnforcer) UpdateAppArmorHostProfile(secPolicies []tp.HostSecur
 		} else {
 			ae.Logger.Errf("Failed to update %d host security rules to the KubeArmor host profile in %s (%s)", policyCount, ae.HostName, err.Error())
 		}
+
+		/* Remove contents of AppArmor profile once the policy is applied
+		 * This will prevent reboot issues related to ungraceful shutdown of kubearmor
+		 */
+		if err := os.Truncate("/etc/apparmor.d/kubearmor.host", 0); err != nil {
+			ae.Logger.Err(err.Error())
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


KubeArmor/enforcer/appArmorEnforcer.go
```diff
@@ -508,6 +508,13 @@ func (ae *AppArmorEnforcer) UpdateAppArmorHostProfile(secPolicies []tp.HostSecur
 		} else {
 			ae.Logger.Errf("Failed to update %d host security rules to the KubeArmor host profile in %s (%s)", policyCount, ae.HostName, err.Error())
 		}
+
+		/* Remove contents of AppArmor profile once the policy is applied
+		 * This will prevent reboot issues related to ungraceful shutdown of kubearmor
+		 */
+		if err := os.Truncate("/etc/apparmor.d/kubearmor.host", 0); err != nil {
```
nyrahul: can you `const` this file name and use it everywhere else?

nyrahul (changes requested):
Thanks .. Just for my understanding, the rationale for this change is that we can delete the kubearmor created apparmor profile for host because it is not needed once the apparmor profile is submitted using `apparmor_parser`. When k8s cluster restarts or the k8s node restarts then the k8s will anyways ask kubearmor to apply the k8s host policy again, in which case the file will be created/applied/deleted again.

CC: @nam-jaehyun

PR #454, opened 2021-10-22
Support a vagrant environment with both docker and containerd

### LINKED ISSUE #353: Fix ownerOnly policy match condition for audit policies
## Bug Report
**General Information**

- Environment: vagrant-dev-env(kubearmor-dev-next)
- Kernel version: `Linux kubearmor-dev-next 5.13.0-14-generic #14-Ubuntu SMP Mon Aug 2 12:43:35 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`
```
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.1", GitCommit:"632ed300f2c34f6d6d15ca4cef3d3c7073412212", GitTreeState:"clean", BuildDate:"2021-08-19T15:45:37Z", GoVersion:"go1.16.7", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.4", GitCommit:"3cce4a82b44f032d0cd1a1790e6d2f5a55d20aae", GitTreeState:"clean", BuildDate:"2021-08-11T18:10:22Z", GoVersion:"go1.16.7", Compiler:"gc", Platform:"linux/amd64"}
```
- Link to relevant artifacts (policies, deployments scripts, ...)
- Target containers/pods

**To Reproduce**

1. Apply a KubeArmorPolicy with action as `Audit` having `ownerOnly` for some file path(or some process path). For example-
```
kind: KubeArmorPolicy
metadata:
  name: ksp-ubuntu-3-file-path-owner-audit
  namespace: multiubuntu
spec:
  severity: 7
  selector:
    matchLabels:
      container: ubuntu-3
  file:
    matchPaths:
    - path: /home/user1/secret_data1.txt  
      ownerOnly: true
  action:
    Audit
```
2. Switch to that user using `su - <user_name>`. Access that file.
```
vagrant@kubearmor-dev-next:~$ kubectl exec -ti -n multiubuntu ubuntu-3-deployment-59bf5f5564-tsft2 -- bash
root@ubuntu-3-deployment-59bf5f5564-tsft2:/# su - user1
user1@ubuntu-3-deployment-59bf5f5564-tsft2:~$ echo "this is secret" >> secret_data1.txt 
user1@ubuntu-3-deployment-59bf5f5564-tsft2:~$ ls -l
total 16
-rwxrwxr-x 1 user1 user1 8304 Jul 13 10:15 hello
-rw-rw-r-- 1 user1 user1   15 Sep 15 07:29 secret_data1.txt
```
3. Check UID of the owner using `cat /etc/passwd | grep <user_name>` . Here it's 1000.
```
user1@ubuntu-3-deployment-59bf5f5564-tsft2:~$ cat /etc/passwd|grep user1
user1:x:1000:1000::/home/user1:/bin/bash
```
4. Check generated logs. There will be a MatchedPolicy log generated for the echo command used by owner in step 2. Notice that the UID is 1000 flag in the log.
```
{"timestamp":1631695097,"updatedTime":"2021-09-15T08:38:17.719589Z","hostName":"kubearmor-dev-next","namespaceName":"multiubuntu","podName":"ubuntu-3-deployment-59bf5f5564-tsft2","containerID":"e928f60608f10f11fd79903458b9658515640d0721900df9c0c3b6e3b01478ee","containerName":"ubuntu-3-container","hostPid":99747,"ppid":271,"pid":279,"uid":1000,"policyName":"ksp-ubuntu-3-file-path-owner-audit","severity":"7","type":"MatchedPolicy","source":"/bin/cat /home/user1/secret_data1.txt","operation":"File","resource":"/home/user1/secret_data1.txt","data":"syscall=SYS_OPENAT fd=-100 flags=O_RDONLY","action":"Audit","result":"Passed"}
```
**Expected behavior**
We shouldn't get MatchedPolicy log for owner file accesses as it is permitted in the policy and is not a violation. This needs to be ensured for both Process and File ownerOnly operations.

### LINKED ISSUE #450: Vagrant: Kubernetes (kubelet) failed to start with containerd as CRI runtime
When KubeArmor is launched (from vagrant script) with `containerd` as container runtime, k8s cannot work properly. A subsequent error message is: `KubeletNotReady              [container runtime status check may not have completed yet, container runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: cni plugin not initialized]`.

elfadel:
The issue described in #450 appears on Kubernetes (components) version `1.21.3-00`. Using version `kubelet=1.21.0-00` resolves the issue.

Fixes #450.
Definitely fixes #353.


deployments/generic/kubearmor.yaml
```diff
@@ -98,6 +98,12 @@ spec:
         - name: containerd-sock-path # containerd (read-only)
           mountPath: /var/run/containerd/containerd.sock
           readOnly: true
+        - name: docker-sock-path # docker (read-only)
```
nam-jaehyun: when we use containerd, is it possible to use the docker socket file?

deployments/GKE/kubearmor.yaml
```diff
@@ -98,6 +98,12 @@ spec:
         - name: containerd-sock-path # containerd (read-only)
           mountPath: /var/run/containerd/containerd.sock
           readOnly: true
+        - name: docker-sock-path # docker (read-only)
```
nam-jaehyun: when we use containerd, is it possible to use the docker socket file?

nam-jaehyun (commented):
@elfadel can you double-check if we can use the docker socket path and the docker storage path when we use contained instead of docker?

deployments/generic/kubearmor.yaml
elfadel: @nam-jaehyun indeed, the YAML files modifications are additional fixes for issue #353 (when we use docker).

nam-jaehyun: Both docker and containerd have containerd.sock.
That's why the generic deployment YAML file only has containerd.sock.
However, if we add docker.sock, this YAML file cannot be applied to the environment with only containerd since there would be no docker.sock.

In sum, we need a solution that can cover both docker and containerd.

elfadel: I double-checked and the YAML files applied successfully. Should I remove the docker.sock anyway??
@nam-jaehyun

nam-jaehyun:
@elfadel As we discussed, I'm waiting for a generic way to support both platforms.

nam-jaehyun (approved):
I updated the code and I'll raise a PR for additional YAML updates

PR #521, opened 2021-12-07
Enabled server-side secure grpc for connections established with VM

seswarrajan:


nyrahul (approved):
LGTM

KubeArmor/kvmAgent/kvmAgent.go
```diff
@@ -53,9 +59,22 @@ func NewKVMAgent(eventCb tp.KubeArmorHostPolicyEventCallback) *KVMAgent {
...
+		InsecureSkipVerify: false,
+		RootCAs:            cp,
+		ServerName:         "0.0.0.0",
+		MinVersion:         tls.VersionTLS12,
+	}
+
 	// Connect to gRPC server
 	ctx, _ := context.WithTimeout(context.Background(), time.Second*5)
-	gRPCConnection, err := grpc.DialContext(ctx, kvm.gRPCServer, grpc.WithInsecure(), grpc.WithBlock())
+	gRPCConnection, err := grpc.DialContext(ctx, kvm.gRPCServer, grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)), grpc.WithBlock())
```
nam-jaehyun, line 77: Can we add the case if certs do not exist as exception handling?
In this implementation, it seems that KubeArmor will fail if no certs.
I feel when people test KubeArmor locally, they may not have any certs.

nam-jaehyun (changes requested):
we may need to handle possible corner cases.

KubeArmor/kvmAgent/kvmAgent.go
seswarrajan: Yes @nam-jaehyun. It is possible. Is it like when certs are not available, we have to establish connection in insecure mode ?
We can use InSecureSkipVerify in TLSConfig if certs are not available or we can use grpc.WithInSecure() option in Dial.

nam-jaehyun (commented):
was this PR merged into the main branch?

seswarrajan:
Currently closing the same as this has issue loading certificate. Will handle later.

Issue #515, opened 2021-12-01
Fix eBPF compilation on CentOS 8 and RHEL 8


daemon1024:
## Bug Report

KubeArmor can't compile eBPF code on CentOS 8 and RHEL 8 and hence failing to start.

This due to our conditional checking in 
https://github.com/kubearmor/KubeArmor/blob/afc99cb1e28a2b1779a884479e232fd239f6d2ca/KubeArmor/BPF/system_monitor.c#L171

But redhat has back-ported a few commits back to 4.18 so that voids are checking

Related https://github.com/draios/sysdig/pull/1650
Also ref #504 

**General Information**

- Environment description : CentOS 8 and RHEL 8
- Kernel version  `Linux kubearmor-dev-sel 4.18.0-348.2.1.el8_5.x86_64 #1 SMP Tue Nov 16 14:42:35 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`

**To Reproduce**

1. Setup KubeArmor on CentOS 8 or RHEL 8

**Expected behavior**

KubeArmor should just run as is

**Log**
<details>
<summary>KubeArmor log</summary>
<p>

```
2021-12-01 07:02:48.664548	INFO	Initializing an eBPF program
/virtual/main.c:161:31: error: no member named 'pids' in 'struct task_struct'; did you mean 'pid'?
    return task->real_parent->pids[PIDTYPE_PID].pid->numbers[level].nr;
                              ^~~~
                              pid
include/linux/sched.h:797:11: note: 'pid' declared here
        pid_t                           pid;
                                        ^
/virtual/main.c:161:35: error: subscripted value is not an array, pointer, or vector
    return task->real_parent->pids[PIDTYPE_PID].pid->numbers[level].nr;
           ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
/virtual/main.c:172:32: error: no member named 'pids' in 'struct task_struct'; did you mean 'pid'?
    return task->group_leader->pids[PIDTYPE_PID].pid->numbers[level].nr;
                               ^~~~
                               pid
include/linux/sched.h:797:11: note: 'pid' declared here
        pid_t                           pid;
                                        ^
/virtual/main.c:172:36: error: subscripted value is not an array, pointer, or vector
    return task->group_leader->pids[PIDTYPE_PID].pid->numbers[level].nr;
           ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
[...]

PR #529, opened 2021-12-09
Package KubeArmor as a systemd unit

### LINKED ISSUE #500: kubearmor as systemd process
## Feature Request

Currently, kubearmor supports k8s env by default. Also kubearmor can operate on a standalone VM/Bare-metal as a docker container using [KVMService](https://github.com/kubearmor/KVMService/).

Kubearmor could be ran as a host-process as a systemd service. The policy application and visibility could be accessed using karmor CLI.

Even though k8s is picking up steam there are lots of envs that depend on VM/Bare-metal. Currently, kubearmor installs as a docker container on VM. Installing kubearmor directly as a systemd process would remove docker dep and also make it easier for user to experiment directly on the host.

- [x] Install kubearmor as systemd process. Have a installation script to handle it. Start kubearmor with appropriate arguments.
- [x] Release kubearmor binaries. Have release cycle.
- [x] Update karmor to support applying policy using it.

daemon1024:
Fixes #500 #515

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,44 @@
+name: goreleaser
+
+on:
+  push:
+    tags:
+      - "*"
+
+permissions:
+  contents: write
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Build bcc
+        run: |
+          set -x
+          sudo apt-get update
+          sudo apt-get -y install build-essential cmake bison flex git python3 python3-pip clang-9 libllvm9 llvm-9-dev libclang-9-dev zlib1g-dev libelf-dev libedit-dev libfl-dev
+          pushd /tmp
+          git clone https://github.com/iovisor/bcc.git
+          mkdir -p bcc/build; cd bcc/build
+          sudo ln -s /usr/lib/llvm-9 /usr/local/llvm
+          cmake .. -DPYTHON_CMD=python3 -DCMAKE_INSTALL_PREFIX=/usr
+          make -j$(nproc)
+          sudo make install
+          popd
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          distribution: goreleaser
+          version: latest
+          args: release --rm-dist
+          workdir: KubeArmor
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,7 +32,10 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.17
+          go-version: 1.
+      - name: Installing Karmor
+        run: curl -sfL https://raw.githubusercontent.com/kubearmor/kubearmor-client/main/install.sh | sudo sh -s -- -b .
+        working-directory: KubeArmor
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v2
         with:
--- a/KubeArmor/BPF/system_monitor.c
+++ b/KubeArmor/BPF/system_monitor.c
@@ -15,6 +15,12 @@
 #define asm_inline asm
 #endif
 
+#ifdef RHEL_RELEASE_CODE
+#if (RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(8, 0))
+#define RHEL_RELEASE_GT_8_0
+#endif
+#endif
+
 #include <linux/version.h>
 
 #include <linux/nsproxy.h>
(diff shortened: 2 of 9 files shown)


nam-jaehyun:
When we use deb or rpm to install KubeArmor, how can we see KubeArmor (error) messages that were shown in the standard output? If KubeArmor fails, do we have any restart option?

KubeArmor/packaging/kubearmor.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=KubeArmor
+
+[Service]
+User=root
+KillMode=process
+EnvironmentFile=/etc/sysconfig/kubearmor.conf
```
nam-jaehyun: can we use /usr/local/etc instead of /etc?

KubeArmor/packaging/kubearmor.service
```diff
@@ -0,0 +1,12 @@
...
+ExecStart=/usr/local/bin/kubearmor -logPath=/tmp/kubearmor.log -enableKubeArmorHostPolicy=$ENABLE_HOST_POLICY
```
nam-jaehyun: If we're going to use kubearmor.conf, can you also add the other options (e.g., -eanbleKubeArmorPolicy, -enableKubeArmorVm) too?

KubeArmor/nfpm.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - src: ./kubearmor
+    dst: /usr/local/bin/kubearmor
+  - src: ./packaging/kubearmor.conf
+    dst: /etc/sysconfig/kubearmor.conf
+    type: config
+  - src: ./packaging/kubearmor.service
+    dst: /usr/lib/systemd/system/kubearmor.service
+    type: config
+  - src: /usr/local/bin/kubearmor
+    dst: /usr/sbin/kubearmor
```
nam-jaehyun: do we need to add kubearmor into /usr/sbin too?

KubeArmor/nfpm.yaml
```diff
@@ -0,0 +1,32 @@
...
+    type: symlink
+  - src: ./BPF/*
+    dst: /usr/local/bin/BPF/
```
nam-jaehyun: I think this is not a good way.

why don't we create new directories (e.g., /opt/kubearmor{,/etc,/log,/BPF})
and make a symbolic link (e.g., /opt/kubearmor/kubearmor -> /usr/local/bin/kubearmor)?

KubeArmor/BPF/system_monitor.c
```diff
@@ -157,7 +163,7 @@ static __always_inline u32 get_task_ns_ppid(struct task_struct *task)
 {
     unsigned int level = task->real_parent->nsproxy->pid_ns_for_children->level;
 
-#if LINUX_VERSION_CODE < KERNEL_VERSION(4, 19, 0)
+#if LINUX_VERSION_CODE < KERNEL_VERSION(4, 19, 0) && !defined(RHEL_RELEASE_GT_8_0)
```
nam-jaehyun, line 166: why don't we directly use "RHEL_RELEASE_CODE < RHEL_RELEASE_VERSION(8, 0)"?

.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
...
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Build bcc
```
nam-jaehyun, line 19: when we use deb or rpm, what will happen if bcc is not installed in a host?

daemon1024:
> How can we see KubeArmor (error) messages that were shown in the standard output?

You can debug it as you would for any systemd process.

`systemctl status kubearmor`
or
`journalctl -u kubearmor.service`

> If KubeArmor fails, do we have any restart option?

`sudo systemctl restart kubearmor`

@nam-jaehyun

KubeArmor/BPF/system_monitor.c
daemon1024: Not sure how could we club RHEL < 8 and this if condition.
Also it will error out on non RHEL systems if we directly use the condition, we need #ifdef (if that's what you wanted to infer )

KubeArmor/nfpm.yaml
daemon1024: Would it be a good idea to use /usr/share/KubeArmor and we can set the working directory of systemd service to /usr/share/kubearmor?

KubeArmor/nfpm.yaml
daemon1024: > /usr/sbin -> This directory contains program binaries for  system  administration 

So I thought we should put it there too 🤔 I don't think it's necessary

.github/workflows/release.yml
daemon1024: KubeArmor will fail to start as of now..

KubeArmor/packaging/kubearmor.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=KubeArmor
+
+[Service]
+User=root
+KillMode=process
+EnvironmentFile=/opt/kubearmor/kubearmor.conf
+WorkingDirectory=/opt/kubearmor/
+ExecStart=/usr/local/bin/kubearmor -logPath=${LOG_PATH} -enableKubeArmorHostPolicy=${ENABLE_HOST_POLICY} -enableKubeArmorVm=${ENABLE_KVM} -gRPC=${GRPC}
```
daemon1024: We should consider using [Viper](https://github.com/spf13/viper) for configuration management. It can accept config from flags, files and env files, any of them works.

nyrahul (approved):
LGTM. 

Foll issues we will handle separately:
* BCC dependency
* RHEL compilation flags

PR #536, opened 2021-12-16
Add gRPC endpoint for native policy handling

daemon1024:
Ref #500


nyrahul (approved):
LGTM

KubeArmor/policy/policy.go
```diff
@@ -0,0 +1,33 @@
...
+	"encoding/json"
+
+	kg "github.com/kubearmor/KubeArmor/KubeArmor/log"
+	tp "github.com/kubearmor/KubeArmor/KubeArmor/types"
+	pb "github.com/kubearmor/KubeArmor/protobuf"
+)
+
+// PolicyServer Structure
+type PolicyServiceServer struct {
+	pb.PolicyServiceServer
```
seswarrajan, line 17: @daemon1024 In this struct for server, please check any references. As per my understanding we need to use **pb.UnimplementedPolicyServiceServer**

This might throw an error later when establishing connection

protobuf/policy.proto
```diff
@@ -0,0 +1,17 @@
+syntax = "proto3";
+ 
+package policy;
```
seswarrajan, line 3: Suggest to provide status, policy and hostPolicy with starting Caps. Looks better. 
Status, Policy and HostPolicy

daemon1024: image
This changes the json field names to Title Case which doesn't seem the standard convention. What do you think?

KubeArmor/policy/policy.go
daemon1024: seems to be working fine, in it's current state so marking it as resolved

PR #543, opened 2021-12-20
Run tests only if changes in relevant files

daemon1024:
Our test framework in https://github.com/kubearmor/KubeArmor/blob/main/.github/workflows/ci-kubeadm.yml is quite expensive, taking >10 min to run. 
It's not needed when we make documentation related changes or some other changes.
This PR restricts the CI tests to only run if there are changes in `KubeArmor/**` or `tests/**`

### DIFF
--- a/.github/workflows/ci-kubeadm.yml
+++ b/.github/workflows/ci-kubeadm.yml
@@ -3,8 +3,16 @@ name: ci-test
 on:
   push:
     branches: [main]
+    paths: 
+    - 'KubeArmor/**'
+    - 'tests/**'
+    - 'protobuf/**'
   pull_request:
     branches: [main]
+    paths: 
+    - 'KubeArmor/**'
+    - 'tests/**'
+    - 'protobuf/**'
 
 jobs:
   build:


.github/workflows/ci-kubeadm.yml
```diff
@@ -3,8 +3,14 @@ name: ci-test
 on:
   push:
     branches: [main]
+    paths:
```
nyrahul, line 6: how about adding `protobuf/**`  path as well

.github/workflows/ci-kubeadm.yml
```diff
@@ -3,8 +3,14 @@ name: ci-test
 on:
   push:
     branches: [main]
+    paths: 
+    - 'KubeArmor/**'
+    - 'tests/**'
   pull_request:
     branches: [main]
+    paths:
```
nyrahul, line 12: I guess we might have to consider deployments folder too? Since it has yamls which if changed needs to be validated by tests.

nyrahul (changes requested):
Thanks for this. Very helpful. PFA my comments.

.github/workflows/ci-kubeadm.yml
daemon1024: These tests are based on KubeArmor as a process not deployment, so I don't think they are relevant in this case.

nyrahul: @Ankurk99  is adding tests based on images ... will this not impact?

daemon1024: They will be in a separate action(file)

.github/workflows/ci-kubeadm.yml
```diff
@@ -3,8 +3,14 @@ name: ci-test
 on:
   push:
     branches: [main]
+    paths: 
+    - 'KubeArmor/**'
```
nyrahul, line 7: @daemon1024 , still don't see protobuf/** added!

nyrahul (approved):
LGTM

nam-jaehyun:
Good PR!

PR #541, opened 2021-12-18
telemetry/alerts issue fix; dockerfile fix

nyrahul:
* ensuring protobuf compilation before kubearmor src compilation
* upgraded Docker image to use of golang 1.17 from 1.15
* upgraded Docker image to use of alpine 3.15 from 3.12
* fixes to file/dir copy during image build
* Removed use of docker tags for GITHUB_SHA ... using docker labels instead
* added verbose log when Watch*() APIs break

Signed-off-by: Rahul Jadhav <r@accuknox.com>

PR #547, opened 2021-12-23
Refactor stream handling in gRPC service

daemon1024:
Consider server context before trying to send out messages

Ref #541


nyrahul (approved):
🚀 LGTM

PR #548, opened 2021-12-24
Host Visibility, Config update

### LINKED ISSUE #546: enabling host visibility for VM/Bare-metal
## Feature Request

**Enabling Host Visibility for VM/Bare-Metal nodes**
#404 added support for enabling host visibility. But this support works only for k8s nodes that can be annotated with `kubectl annotate node kubearmor-dev kubearmor-visibility=process,file,network,capability`.

In case of non-k8s VM/Bare-Metal nodes, currently there is no way of enabling the host visibility.

Policy discovery and other observability features are dependent on Host visibility feeds. These features currently do not work for VM/Bare-Metal nodes because of this issue.

Enable host visibility based on startup time option. For e.g., one can start kubearmor with `--enableHostVisibility=process,file,network,capability`.

Other option is to enable/disable visibility dynamically using grpc endpoint. However, at this time there is no strong use-case for this dynamicity.

nyrahul:
- [x] enable host visibility for kubearmor in systemd mode
- [x] use viper for configuration management. Remove dependency on `kubearmor.conf`
- [x] use kubearmor.yaml as configuration file for systemd mode

Fixes: #546 

### DIFF
--- a/KubeArmor/config/config.go
+++ b/KubeArmor/config/config.go
@@ -12,6 +12,7 @@ import (
 	"github.com/spf13/viper"
 )
 
+// KubearmorConfig Configuration structure for Kubearmor
 type KubearmorConfig struct {
 	Grpc           string // GRPC Port to use
 	Cluster        string // Cluster name to use for feeds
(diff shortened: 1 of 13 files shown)


KubeArmor/config/config.go
```diff
@@ -0,0 +1,123 @@
...
+
+func LoadConfig() error {
+	// Set defaults
+	viper.SetDefault(CFG_CLUSTER, "default")
+	viper.SetDefault(CFG_GRPC, "32767")
+	viper.SetDefault(CFG_LOGPATH, "/tmp/kubearmor.log")
+	viper.SetDefault(CFG_KUBEARMORPOLICY, true)
+	viper.SetDefault(CFG_KUBEARMORHOSTPOLICY, false)
+	viper.SetDefault(CFG_KUBEARMORVM, false)
+	viper.SetDefault(CFG_HOSTVISIBILITY, "process,file,network,capabilities")
```
daemon1024: Ideally it should default to whatever defaults are set in command line flags.
We shouldn't be needing to deal with so many variables, just a single instance of `KubearmorConfig` should be enough.

https://github.com/carolynvs/stingoftheviper can take insights from here.

nyrahul: Good point. We can get defaults directly from command line flags.

KubeArmor/config/config.go
```diff
@@ -0,0 +1,100 @@
...
+
+var GlobalCfg KubearmorConfig
+
+const CFG_CLUSTER string = "cluster"
+const CFG_GRPC string = "gRPC"
+const CFG_LOGPATH string = "logPath"
+const CFG_HOSTVISIBILITY string = "hostVisibility"
+const CFG_KUBEARMORPOLICY string = "enableKubeArmorPolicy"
+const CFG_KUBEARMORHOSTPOLICY string = "enableKubeArmorHostPolicy"
+const CFG_KUBEARMORVM string = "enableKubeArmorVm"
```
daemon1024:
```suggestion
const CFG_KUBEARMORPOLICY string = "policy"
const CFG_KUBEARMORHOSTPOLICY string = "host-policy"
const CFG_KUBEARMORVM string = "vm"
```
Since the flags are for kubearmor only, I don't think they need to have it. Also POSIX suggests to have only lowercase flag names.
We can use `KUBEARMOR` prefix for environment variables like you have for the config file using
`v.SetEnvPrefix("KUBEARMOR")`

Wdyt?

nyrahul: I didn't want to change existing command line parameter key labels (at least in this PR). If we change those then there changes to be done in tests/documentation etc. I started off making changes that way and then reverted back because of these deps.

nam-jaehyun: @nyrahul can you handle the go-lint warnings (e.g., adding single comments or go-lint ignoring tag on top of each line)?

PR #524, opened 2021-12-08
add github actions to test image before pushing the latest 

Ankurk99:
fixes #527

PR #553, opened 2021-12-28
fix release issue due to k3s tests

nyrahul:
A new workflow was introduced (#524) such that kubearmor images could be tested in a local k3s env before pushing the images to the docker hub. There were following problems with that workflow:
1. k3s by default uses containerd and we were building images in local docker. Thus the image was never used. Changed k3s install script to by default use docker and changed the `kubearmor-test-k3s.yaml` to volume mount docker paths.
2. `test-scenarios-github.sh` always required that kubearmor be instantiated as a local process. This involved scanning local logs to check the output conditions of the test result. Changed `test-scenarios-github.sh` to check if there exists a kubearmor running in k8s env before local instantiation.
3. Updated `.github/workflows/latest-release.yml` to use the `./contribution/k3s/install_k3s.sh` ... so that there is a single k3s script for all purpose.
4. Removed k3s dependency on calico CNI, using flannel as default in k3s.
5. Removed unnecessary sleeps in the test scripts, optimizing test time.

### DIFF
--- a/.github/workflows/latest-release.yml
+++ b/.github/workflows/latest-release.yml
@@ -19,19 +19,8 @@ jobs:
 
       - name: Setup Enviroment
         run: |        
-          # create a single-node K3s cluster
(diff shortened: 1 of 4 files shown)


Ankurk99:
Also can you please add an actions timeout (about 20 min)?

PR #558, opened 2021-12-31
Added error handling in kvm-agent for identity removal in kvm-service

seswarrajan:

### DIFF
--- a/KubeArmor/kvmAgent/kvmAgent.go
+++ b/KubeArmor/kvmAgent/kvmAgent.go
@@ -21,6 +21,9 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
+// const variables
+const errIdentityRemoved = "err-identity-removed"
+
 // KVMAgent Structure
 type KVMAgent struct {
 	Identity         string
@@ -96,7 +99,7 @@ func (kvm *KVMAgent) ConnectToKVMService() {
 		if err != nil {
 			kg.Warn("Failed to connect stream")
 
-			if strings.Contains(string(err.Error()), "err-identity-removed") {
+			if strings.Contains(string(err.Error()), errIdentityRemoved) {
 				kg.Warn("Identity removed from server")
 				// close the connection
 				if err = kvm.gRPCConnection.Close(); err != nil {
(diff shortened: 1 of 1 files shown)


KubeArmor/kvmAgent/kvmAgent.go
```diff
@@ -95,6 +96,15 @@ func (kvm *KVMAgent) ConnectToKVMService() {
 		if err != nil {
 			kg.Warn("Failed to connect stream")
 
+			if strings.Contains(string(err.Error()), "err-identity-removed") {
```
nyrahul: I couldn't find the code location that returns "err-identity-removed" as err. Can you provide a ref?

Nonetheless, it would be better to define a constant for such strings and use it from the imported module since any change in the string could have impact across modules.

nyrahul: Also I didnot understand why the GRPC connection is only closed in the event of Identity removal! Why is it ok to keep the connection open in other cases and not return.

nyrahul (commented):
My review comments inline.

KubeArmor/kvmAgent/kvmAgent.go
seswarrajan: 1. The code where the error is sent is as part of kvm-service. A PR is raised for the same.
2. Yes, we have to define a constant for the same.
3. The reason for return is that, when identity is removed, we may not be able to connect with the same. We need to establish a new connection with a new identity. Hence we will be running a new installation script.

KubeArmor/kvmAgent/kvmAgent.go
```diff
@@ -123,6 +127,16 @@ func (kvm *KVMAgent) ConnectToKVMService() {
 			if err == io.EOF {
 				continue
 			} else if err != nil {
+
+				if strings.Contains(string(err.Error()), errIdentityRemoved) {
```
nyrahul, line 131: It is better to apply this check after the grpc connection close below. This way we can avoid repeatative grpc connection close.

seswarrajan: Hi @nyrahul It is not possible to use this error handling below/after the grpc connection close. GRPC connection close will be successful and the err is nil. Hence the condition is not valid. That is the reason for handling the error handling for identity-removal before.

PR #567, opened 2022-01-20
Use cfg.GlobalCfg

nam-jaehyun:
- use cfg.GlobalCfg rather than copying given arguments to each module
- update test scripts and workflows (including Makefile)
- update setup scripts and Vagrantfile to support the development environments based on both Ubuntu and CentOS
- update stdout messages
- clean up whitespaces


KubeArmor/core/kubeArmor.go
```diff
@@ -339,39 +352,33 @@ func KubeArmor() {
...
+		if dm.Node.NodeIP == "" {
+			for timeout := 0; timeout <= 60; timeout++ {
+				if dm.Node.NodeIP != "" {
+					break
+				}
 
-		dm.Node.KernelVersion = kl.GetCommandOutputWithoutErr("uname", []string{"-r"})
-		dm.Node.KernelVersion = strings.TrimSuffix(dm.Node.KernelVersion, "\n")
+				if dm.Node.NodeIP == "" && timeout == 60 {
+					kg.Print("The node information is not available")
```
nyrahul: It is better to print this info every second ... This info usually helps me in knowing that `kubectl proxy &` is not executed.

nyrahul (changes requested):

KubeArmor/core/kubeArmor.go
kloudmax: @nyrahul Added the update as you commented.

nam-jaehyun: @nyrahul Added the update as you commented.

Issue #573, opened 2022-01-26
Policies not applied when deployment is deleted/reapplied


nyrahul:
## Bug Report

The Kubearmor policy application does not work when a deployment is deleted and the same deployment is applied again.

**General Information**

- Environment description (GKE, VM-Kubeadm, vagrant-dev-env, minikube, microk8s, ...)
GKE (but should be able to produce in any k8s env)
- Kernel version (run `uname -a`)
GKE-COS (but should be able to reproduce with any supported OS)
- Orchestration system version in use (e.g. `kubectl version`, ...)
k8s 1.22
- Link to relevant artifacts (policies, deployments scripts, ...)
Sample Deployment [demo-dep.yaml](https://raw.githubusercontent.com/nyrahul/src/master/docs/demo/demo-dep.yaml)
Sample Policy [deny-pols.yaml](https://raw.githubusercontent.com/nyrahul/src/master/docs/demo/deny-pols.yaml)

**To Reproduce**

1. `kubectl apply -f demo.dep.yaml`
2. `kubectl apply -f deny-pols.yaml`
3. Run `sleep` command in the ubuntu pod and you should get `Permission Denied`
4. Now, delete the deployment and reapply it instantly. `kubectl delete -f demo.dep.yaml && kubectl apply -f demo.dep.yaml`
5. Note that the policy is still present.
6. Run `sleep` command in the ubuntu pod and you won't get `Permission Denied`

**Expected behavior**
Even in Step 6, the user should have got `Permission Denied` since the policy is still applied.

**Analysis**
The previous pods are in *Terminating* state for some time and the new pods get into *Running* state before the previous pods are fully deleted. When KubeArmor gets a *Deleted* event after the previous pods are fully terminated, the AppArmor profile is unregistered. However, the apparmor profile was in use for the new *Running* pods. Thus (since the apparmor profile is deleted) the new *Running* pod does not have any associated apparmor profile anymore.

PR #576, opened 2022-01-28
fix AppArmor profile management

nam-jaehyun:
- fix AppArmor profile management (#573)


nyrahul (approved):
LGTM

PR #580, opened 2022-01-28
Refactor exported objects into separate subpackage

daemon1024:
Current objects weren't importable due to deployment generate tooling being a program and not an importable package, so this commit refactors exported objects into a separate subpackage to serve the intended usecase

This PR aims to fix build errors in https://github.com/kubearmor/kubearmor-client/pull/34
and revert https://github.com/kubearmor/kubearmor-client/commit/0e8b70cc95bd16cef7705bbfb025ababc8dbf6ed

### DIFF
--- a/deployments/go.mod
+++ b/deployments/go.mod
@@ -10,8 +10,8 @@ replace (
 
 require (
 	github.com/clarketm/json v1.17.1
-	github.com/kubearmor/KubeArmor/pkg/KubeArmorHostPolicy v0.0.0-00010101000000-000000000000
-	github.com/kubearmor/KubeArmor/pkg/KubeArmorPolicy v0.0.0-00010101000000-000000000000
+	github.com/kubearmor/KubeArmor/pkg/KubeArmorHostPolicy v0.0.0-20220128051912-b9f5851b939b
+	github.com/kubearmor/KubeArmor/pkg/KubeArmorPolicy v0.0.0-20220128051912-b9f5851b939b
 	k8s.io/api v0.22.3
 	k8s.io/apimachinery v0.22.3
 	sigs.k8s.io/yaml v1.2.0
--- a/deployments/get/defaults.go
+++ b/deployments/get/defaults.go
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2021 Authors of KubeArmor
 
-package main
+package deployments
 
 import corev1 "k8s.io/api/core/v1"
 
--- a/deployments/get/objects.go
+++ b/deployments/get/objects.go
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2021 Authors of KubeArmor
 
-package main
+package deployments
 
 import (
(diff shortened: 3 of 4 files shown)


nyrahul (approved):

deployments/go.mod
```diff
@@ -10,27 +10,61 @@ replace (
 
 require (
 	github.com/clarketm/json v1.17.1
-	github.com/kubearmor/KubeArmor/pkg/KubeArmorHostPolicy v0.0.0-00010101000000-000000000000
-	github.com/kubearmor/KubeArmor/pkg/KubeArmorPolicy v0.0.0-00010101000000-000000000000
+	github.com/kubearmor/KubeArmor/pkg/KubeArmorHostPolicy v0.0.0-20220128051912-b9f5851b939b
```
nam-jaehyun, line 13: let's use replace() not to depend on the specific version like [go.mod](https://github.com/kubearmor/KubeArmor/blob/main/KubeArmor/go.mod)

nam-jaehyun (changes requested):
I added a small comment. please take a look at it.

deployments/go.mod
daemon1024: We still have the replace directive but I specified version number regardless of that because importing this package with the v000 errored out with 
> go: github.com/kubearmor/KubeArmor/deployments@v0.0.0-00010101000000-000000000000 requires
        github.com/kubearmor/KubeArmor/pkg/KubeArmorHostPolicy@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000

in kArmor

nam-jaehyun: Is this error popped up because you didn't specify all required paths?
```
replace (
	github.com/kubearmor/KubeArmor => ../
	github.com/kubearmor/KubeArmor/deployments => ./
	github.com/kubearmor/KubeArmor/deployments/get => ./get
	github.com/kubearmor/KubeArmor/pkg => ../pkg
	github.com/kubearmor/KubeArmor/pkg/KubeArmorPolicy => ../pkg/KubeArmorPolicy
	github.com/kubearmor/KubeArmor/pkg/KubeArmorHostPolicy => ../pkg/KubeArmorHostPolicy
)
```

I don't think that defining the specific version of our code will raise unexpected side-effects (e.g., difficulty in debugging, additional modification when we have some changes, unintended errors when some other guys make code changes because they don't expect that they also need to update this file) later.

nam-jaehyun (commented):
Let's rethink if we have to depend on the specific version...

deployments/go.mod
daemon1024: When we use replace directive it disregards the version specified in require directive for local module development.

I cannot add these replace driectives to kArmor since it is an independent tool. Perhaps if someone else creates a new utility for KubeArmor they cannot leverage the replace directive as well.

go 1.18 beta 2 introduced a new mode called workspace mode to handle these well but I am not sure what's expected to be done right now https://go.googlesource.com/proposal/+/master/design/45713-workspace.md

nam-jaehyun (approved):
LGTM

PR #589, opened 2022-02-03
fix kubearmor on vm/bare metal

daemon1024:
We started terminating KubeArmor if there's no k8s or KVMAgent found in https://github.com/kubearmor/KubeArmor/commit/d2cf1f2c47c0a3a4d1413f94177455789200ba2c But we have a bare metal/vm mode which supports host policy enforcement leveraging kArmor tool used by our systemd service. This commits reverts back the termination


nyrahul (approved):
LGTM

PR #591, opened 2022-02-04
Enable hostpolicy on VM mode for -enableKubeArmorHostPolicy flag

seswarrajan:


KubeArmor/common/common.go
```diff
@@ -322,7 +322,7 @@ func GetExternalIPAddr() string {
 // IsK8sLocal Function
 func IsK8sLocal() bool {
 	k8sConfig := os.Getenv("KUBECONFIG")
-	if _, err := os.Stat(filepath.Clean(k8sConfig)); err == nil {
+	if _, err := os.Stat(filepath.Clean(k8sConfig)); k8sConfig != "" && err == nil {
 		return true
 	}
```
nyrahul: No point in doing a `os.Stat` and then checking if `k8sConfig != ""`!
```suggestion
        if k8sConfig != "" {
	    if _, err := os.Stat(filepath.Clean(k8sConfig)); err == nil {
	   	return true
	    }
	}
```

KubeArmor/core/kubeArmor.go
```diff
@@ -321,7 +321,8 @@ func KubeArmor() {
 	// create a daemon
 	dm := NewKubeArmorDaemon()
 
-	if cfg.GlobalCfg.KVMAgent {
+	// Enable KubeArmorHostPolicy for both VM and KVMAgent
+	if !kl.IsK8sEnv() && (cfg.GlobalCfg.KVMAgent || cfg.GlobalCfg.HostPolicy) {
```
nyrahul: What happens in cases where user has `~/.kube/config` file, but still wants to use kubearmor in pure VM/systemd mode?
Just the presence of kubeconfig file will cause kubearmor to think that it is using k8s mode of operation!

nyrahul (changes requested):

KubeArmor/core/kubeArmor.go
seswarrajan: Yes. Possible. In that case it will deny/ignore to run as systemd mode.

nam-jaehyun: How can we know if a user wants to run KubeArmor as a systemd mode in such as case?
@seswarrajan What happens if a user uses -enableKubeArmorVm without kvm-service?

seswarrajan: @nam-jaehyun User can enable KubeArmor systemd mode only by -enableKubeArmorHostPolicy flag. It is not possible to run kubeArmor in VM with -enableKubeArmorVm flag as this flag tries to connect to kvm-service and if the connection fails, KubeArmor will terminate.

seswarrajan: @nyrahul In that case we need to ignore the IsK8sEnv() validation check in the code. But internally in the code while applying for new policies IsK8sEnv() check is still run.

nam-jaehyun (approved):
I'm okay with this PR.

KubeArmor/core/kubeArmor.go
nyrahul: This has to be fixed holistically ... currently the code just allows to bypass the first hurdle. If the IsK8sEnv() check is present elsewhere, then VM/systemd mode cannot be used.

Can you check if kubearmor works in systemd mode (with policy enforcement) when there is ~/.kube/config in the path?

nyrahul (changes requested):

KubeArmor/common/common.go
```diff
@@ -23,6 +23,8 @@ import (
 // == Common == //
 // ============ //
 
+var IsNonK8sEnv bool
```
nyrahul: Please fix the lint warnings.

nyrahul (changes requested):

KubeArmor/Makefile
```diff
@@ -39,7 +39,7 @@ run: build
 run-host-only: build
 	cd $(CRDDIR); kubectl apply -f KubeArmorHostPolicy.yaml
 	cd $(CURDIR); sudo rm -f /tmp/kubearmor.log
-	cd $(CURDIR); sudo -E ./kubearmor -logPath=/tmp/kubearmor.log -enableKubeArmorPolicy=false -enableKubeArmorHostPolicy=true
+	cd $(CURDIR); sudo -E ./kubearmor -logPath=/tmp/kubearmor.log -enableKubeArmorPolicy=false -enableKubeArmorHostPolicy=true -k8s=false
```
nyrahul: Cannot set k8s to false in the Makefile ... This will impact regular execution.

KubeArmor/config/config.go
```diff
@@ -144,9 +150,11 @@ func LoadConfig() error {
 		GlobalCfg.Policy = false
 		GlobalCfg.HostPolicy = true
 	}
+	GlobalCfg.K8sEnv = viper.GetBool(ConfigK8sEnv)
+	kl.IsK8SEnv = viper.GetBool(ConfigK8sEnv)
```
nyrahul: Please store the IsK8sEnv in GlobalCfg and then use GlobalCfg in kl module.

KubeArmor/config/config.go
```diff
@@ -28,6 +28,7 @@ type KubearmorConfig struct {
 	Policy     bool // Enable/Disable policy enforcement
 	HostPolicy bool // Enable/Disable host policy enforcement
 	KVMAgent   bool // Enable/Disable KVM Agent
+	K8sEnv     bool // Is k8s env ?
```
nyrahul, line 30: This variable is not used currently. Please set this variable.

KubeArmor/common/common.go
```diff
@@ -23,6 +23,9 @@ import (
 // == Common == //
 // ============ //
 
+// IsK8SEnv used for specifying k8s environment
+var IsK8SEnv bool
+
```
nyrahul:
```suggestion
```
Remove this var and use GlobalCfg.K8sEnv bool in the check below.

KubeArmor/common/common.go
```diff
@@ -329,9 +332,15 @@ func GetExternalIPAddr() string {
 
 // IsK8sLocal Function
 func IsK8sLocal() bool {
+	if !IsK8SEnv {
```
nyrahul:
```suggestion
	if !cfg.K8sEnv {
```

nyrahul (changes requested):

KubeArmor/common/common.go
seswarrajan: It is not possible to include config in common as it causes circular dependency. Hence this global variable.

KubeArmor/common/common.go
seswarrajan: It is not possible to include config in common as it causes circular dependency. Hence this global variable.

KubeArmor/config/config.go
seswarrajan: It is not possible to include config in common as it causes circular dependency. Hence this global variable.

KubeArmor/Makefile
seswarrajan: But this is got Kubearmor run-host-only. When run host-only is invoked via make and also .kube/config is present, then host mode will not work.

nyrahul (approved):

PR #603, opened 2022-02-08
add ContainerImage field

nam-jaehyun:
- change the order of fields in protobuf
- add missing fields (i.e., ContainerImage) in gRPC


nyrahul (approved):
LGTM

PR #608, opened 2022-02-09
create stable release image

Ankurk99:
Creates a docker image for stable release and push it to the Docker Hub.
The image name is fetched from the file "STABLE-RELEASE"


.github/workflows/stable-release.yml
```diff
@@ -0,0 +1,35 @@
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    branches:
+      - 'main'
+      - 'v*'
```
nyrahul:
```suggestion
```
No need to handle this in version branches

.github/workflows/stable-release.yml
```diff
@@ -0,0 +1,35 @@
...
+
+    paths: 
+      - 'STABLE-RELEASE'
+
+name: KubeArmor stable release 
+
+jobs:
+  build:
+    name: Create KubeArmor Release - 18.04
```
nyrahul, line 14:
```suggestion
    name: Create KubeArmor Release - 18.04
    if: github.repository == 'kubearmor/kubearmor'
```
Otherwise this will keep failing on individual forks.

.github/workflows/stable-release.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Checkout KubeArmor code
+        uses: actions/checkout@v2
+
+      - name: Pull Docker Hub image
+        run: |
+            value=`cat STABLE-RELEASE`
+            docker pull docker.io/kubearmor/kubearmor:$value
+            docker image tag kubearmor/kubearmor:$value kubearmor/kubearmor:stable
+
+      - name: Login to Docker Hub
```
nyrahul, line 28: Shouldn't this step be handled before "Pull Docker Hub image" step? Since the earlier step requires docker authentication.

.github/workflows/stable-release.yml
```diff
@@ -0,0 +1,35 @@
...
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_AUTHTOK }}
+
+      - name: Push KubeArmor images to Docker
```
nyrahul, line 34: I think we do not need to explicitly push ... `docker image tag kubearmor/kubearmor:$value kubearmor/kubearmor:stable` will take care of creating the tag on the docker repo.

nyrahul (changes requested):

.github/workflows/stable-release.yml
Ankurk99: Do you mean the docker pull? Docker pull doesn't require login

nyrahul (changes requested):
Also we need STABLE-RELEASE file to be pushed as part of this PR.

Keep the contents of `STABLE-RELEASE` as:
```
v0.2
```

.github/workflows/stable-release.yml
nyrahul: I mean `docker image tag` would require authentication.

nyrahul (approved):

PR #611, opened 2022-02-09
Update documentation to install latest version of bcc

### LINKED ISSUE #609: Update documentation to install latest version of bcc
## Feature Request

**Short Description**
Currently documentation mentions to only install bcc and doesn't specify that the version should be 0.21.0 or greater (here https://github.com/kubearmor/KubeArmor/blob/main/getting-started/kubearmor_vm.md), this causes KubeArmor failing in initializing system monitor upon deployment if the user installs lower version of bcc (for eg: using package manager)

Older bcc and newer kernel version have issue where bcc internal bpf.h doesn't define BPF_SK_LOOKUP but newer kernel header file uses it
Issue: https://github.com/kubearmor/KubeArmor/issues/200

Mention in documentation to install latest version of bcc (0.21.0 or greater)

slayer321:
Signed-off-by: slayer321 <sachin.maurya7666@gmail.com>
fix #609

### DIFF
--- a/getting-started/kubearmor_vm.md
+++ b/getting-started/kubearmor_vm.md
@@ -7,6 +7,8 @@ This recipe explains how to use KubeArmor directly on VM/Bare-Metal host and was
 1. Download the [latest release of KubeArmor](https://github.com/kubearmor/KubeArmor/releases)
 2. Install KubeArmor `sudo apt install ./kubearmor_${VER}_linux-amd64.deb` ... where VER is the kubearmor release version. This will automatically install the required dependencies.
 
+> Note: We automatically install `bpfcc-tools` with our package, your distribution might have an older version of BCC so consider installing BCC from [source](https://github.com/iovisor/bcc/blob/master/INSTALL.md#source) in case of errors.
+
 <details><summary>For distributions other than Ubuntu/Debian</summary>
 <p>
 


getting-started/kubearmor_vm.md
```diff
@@ -4,8 +4,10 @@ This recipe explains how to use KubeArmor directly on VM/Bare-Metal host and was
 
 ## Download and Install KubeArmor
 
-1. Download the [latest release of KubeArmor](https://github.com/kubearmor/KubeArmor/releases)
-2. Install KubeArmor `sudo apt install ./kubearmor_${VER}_linux-amd64.deb` ... where VER is the kubearmor release version. This will automatically install the required dependencies.
+1. Install latest bcc from [source](https://github.com/iovisor/bcc/blob/master/INSTALL.md#source), based on the linux distribution you are using.`version 0.21.0 or greater`
```
daemon1024: We are yet to make a new release, but after our next release bpfcc-tools are automatically installed in our deb package! We would still need folks on ubuntu 18.04 to install BCC from source but I wouldn't suggest adding it as a step. Rather it would be better add a note/blockquoute after the second point stating 

> We automatically install `bpfcc-tools` with our package, your distribution might have an older version of BCC so consider installing BCC from source in case of errors...

Can improve the wording. But something in line with this is perhaps better imo.

daemon1024 (changes requested):

daemon1024:
Thanks a lot for the PR! Left a comment in review :)

getting-started/kubearmor_vm.md
```diff
@@ -4,10 +4,11 @@ This recipe explains how to use KubeArmor directly on VM/Bare-Metal host and was
...
 ## Download and Install KubeArmor
 
-1. Install latest bcc from [source](https://github.com/iovisor/bcc/blob/master/INSTALL.md#source), based on the linux distribution you are using.`version 0.21.0 or greater`
+1. Download the [latest release of KubeArmor](https://github.com/kubearmor/KubeArmor/releases)
+2. Install KubeArmor `sudo apt install ./kubearmor_${VER}_linux-amd64.deb` ... where VER is the kubearmor release version. This will automatically install the required dependencies.
 
-2. Download the [latest release of KubeArmor](https://github.com/kubearmor/KubeArmor/releases)
-3. Install KubeArmor `sudo apt install ./kubearmor_${VER}_linux-amd64.deb` ... where VER is the kubearmor release version. This will automatically install the required dependencies.
+##### Note:
+We automatically install `bpfcc-tools` with our package, your distribution might have an older version of BCC so consider installing BCC from [source](https://github.com/iovisor/bcc/blob/master/INSTALL.md#source) in case of errors.
```
daemon1024:
```suggestion
> Note: We automatically install `bpfcc-tools` with our package, your distribution might have an older version of BCC so consider installing BCC from [source](https://github.com/iovisor/bcc/blob/master/INSTALL.md#source) in case of errors.
```
Let's use a blockquote

daemon1024 (changes requested):
One minor nit and please squash the commits into a single one and We are good to go :D

daemon1024 (approved):
LGTM 🚀

nyrahul (approved):

Issue #616, opened 2022-02-11
Add Github action to check if each commit is compiling


h3llix:
## Feature Request

A PR could be made up of multiple commits. We expect every commit to be self-sufficient i.e., solves a problem and is compilable/testable/revertable. We want the ability to check/ensure that every commit within a PR is compilable (and later testable). This way if needed we can revert individual commits. Further we want the ability to handle `git-bisect` to figure out which commit might have broken the tests.

Thus the requirement is that if a PR contains multiple commits, the make has to be done at every individual commit level.

@Ankurk99 had handled the same for cilium repo in https://github.com/cilium/cilium/pull/15659 (for ref). (We already have PR-level make and tests for Kubearmor)

Tasks
- [ ] Compile/handle [ci-go workflow](https://github.com/kubearmor/KubeArmor/blob/main/.github/workflows/ci-go.yml) for every commit
- [ ] Handle basic tests for every commit (We can discuss what basic tests would mean here)

Here is the [reference GH workflow from cilium](https://github.com/cilium/cilium/blob/master/.github/workflows/lint-build-commits.yaml).

zeborg:
Hey @h3llix, I'd like to proceed with working on this issue. Thanks!

zeborg:
> - [ ] Handle basic tests for every commit (We can discuss what basic tests would mean here)

Hey @nyrahul, I'd like to know about the tests that we're supposed to handle within the CI.

nam-jaehyun:
@h3llix we already check if the new code is compilable during ci-test. can you check if the ci-test contains what you think?

h3llix:
@nam-jaehyun https://github.com/kubearmor/KubeArmor/blob/main/tests/test-scenarios-github.sh#L155 We run the test here. But I don't think we run and test it for individual commits.

nam-jaehyun:
Just curiosity. Does the GitHub Action work per commit? or... Does it work per PR?
I mean the point of view of KubeArmor repo (not a personal repo for development)

nyrahul:
> Just curiosity. Does the GitHub Action work per commit? or... Does it work per PR? I mean the point of view of KubeArmor repo (not a personal repo for development)

The GH action is fired at PR level push event. But inside the event, we can iterate through the commits and build/test for each commit.

s1ntaxe770r:
i think i have a solution working over [here](https://github.com/s1ntaxe770r/ektion/) could someone make a PR and try it out?

[...]

PR #617, opened 2022-02-13
Added CI job to compile KubeArmor

zeborg:
### Related issue
#616 
### Proposed changes
* Added CI job to compile KubeArmor at every push or PR against `main` branch
* Fixed an error while installing KubeArmor dependencies within workflow container by defining `LLVM_ROOT` env var before bcc cmake

### DIFF
--- a/.github/workflows/ci-go.yml
+++ b/.github/workflows/ci-go.yml
@@ -46,6 +46,24 @@ jobs:
         run: make gosec
         working-directory: KubeArmor
 
+  go-build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@v2
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: v1.16
+
+      - name: Install Dependencies
+        run: ./contribution/self-managed-k8s/setup.sh
+
+      - name: Compile KubeArmor
+        run: |
+          cd KubeArmor
+          make
+
   license:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


zeborg:
cc - @h3llix @nyrahul

nyrahul:
@zeborg, I don't see any code in the yaml that builds the code at "every commit" level ! CC: @Ankurk99

zeborg:
@nyrahul I guess I might not have understood the issue properly!

Can you please elaborate what a "commit" refers to here? I thought it meant every push / PR made against `main`.

nyrahul:
> @nyrahul I guess I might not have understood the issue properly!
> 
> Can you please elaborate what a "commit" refers to here? I thought it meant every push / PR made against `main`.

A PR could be made up of multiple commits. We expect every commit to be self-sufficient i.e., solves a problem and is compilable/testable/revertable. We want the ability to check/ensure that every commit within a PR is compilable (and later testable). This way if needed we can revert individual commits. Further we want the ability to handle git-bisect to figure out which commit might have broken the tests.

Thus the requirement is that if a PR contains multiple commits, the `make` has to be done at every individual commit level.

@Ankurk99 had handled the same for cilium repo in [this PR](https://github.com/cilium/cilium/pull/15659) (for ref). (We already have PR-level make and tests for Kubearmor)

zeborg:
Oh ok, I understand it now, thanks!
I kinda had this similar thought earlier as well, but I believe to tackle with this issue the user could just squash the commits in their PR, which is what we're doing in the `kubernetes-sigs/cluster-api-provider-aws` repo as well by making it a necessary procedure for every PR.

nyrahul:
Closing the PR for inactivity. The draft code does not handle the requirement.

PR #618, opened 2022-02-14
Removed enableKubeArmorHostPolicy from kubearmor-test-k3s.yaml

seswarrajan:


nyrahul (changes requested):
using `-enableKubeArmorHostPolicy` should not result in kubearmor getting started in VM mode. This PR is incorrect IMO.

PR #624, opened 2022-02-16
feeder: solved deadlock issue with telemetry

### LINKED ISSUE #622: kubearmor/kubearmor-relay stopping abruptly
## Bug Report

It is found that the logs/telemetry from kubearmor-relay are abruptly stopped once in while.
The root-cause of the problem is not known yet.

Following are the observations (attached below is a screenshot):
1. Observed failure to send a log in kubearmor GRPC server
2. Observed in one case that the NodeIP (of kubearmor node/pod) is empty on kubearmor-relay
3. `WatchLogs()` no more working but `WatchAlerts()` keeps working fine.

**General Information**

GKE-COS

NA

k8s 1.22

**To Reproduce**
We do not have exact steps to reproduce the issue. However, the issue springs up when `karmor log` is started-stopped while connecting to kubearmor-relay.

**Expected behavior**
It is expected that log/telemetry keeps working regardless of failures on karmor, kubearmor-relay or kubearmor itself.

**Screenshots**
image

nyrahul:
If log sending fails, then subsequently the messages on the broadcast channel were blocked. Before sending a message on the broadcast channel, a LogLock is taken. Now if WatchLogs() has returned err due to failure in sending log ... then it tries to delete the client entry and the LogLock is again acquired there, causing deadlock.

The same condition is applicable for WatchAlerts() and WatchMessages(). The solution is to simply remove the lock operation before broadcasting on a channel.

Fixes: #622

### DIFF
--- a/KubeArmor/feeder/feeder.go
+++ b/KubeArmor/feeder/feeder.go
@@ -242,7 +242,7 @@ func (ls *LogService) WatchLogs(req *pb.RequestMessage, svr pb.LogService_WatchL
 				case codes.OK:
 					// noop
 				case codes.Unavailable, codes.Canceled, codes.DeadlineExceeded:
-					kg.Warnf("Failed to send a log=[%+v] err=[%s]", resp, status.Err().Error())
+					kg.Warnf("Failed to send a log=[%+v] err=[%s] CODE=%d", resp, status.Err().Error(), status.Code())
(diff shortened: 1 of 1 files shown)


nam-jaehyun (approved):
Good to merge it.

daemon1024 (approved):

Issue #615, opened 2022-02-10
POD_NAME variable refers to wrong container instead of KubeArmor container


viveksahu26:
### Short Description
Basically all the logs stored inside KubeArmor container whenever any activity is triggered whether it's against or validating to the  applied KuberArmorPolicy in the Kubernetes cluster. Since, all logs are stored inside `/tmp/kubearmor.logs`. Therefore user need to exec inside that KubeArmor pod to see those logs. 
But in the [documentation](https://github.com/kubearmor/KubeArmor/blob/main/examples/multiubuntu.md) logs are checked inside wrong container. Where POD_NAME variable referencing to wrong pod instead of KubeArmor pod.

**Expected behavior**

Logs should be checked inside KubeArmor pod instead of other pods. For the same POD_NAME need to replace. 

**Discussion**

For more, see to [slack](https://kubearmor.slack.com/archives/C01FG4G2PEY/p1644502965753289) discussion.

viveksahu26:
I would love to work on this issue.
/assign me.

nyrahul:
Lets not use `/tmp/kubearmor.log` directly for alerts/telemetry purpose ... We have a [karmor](https://github.com/kubearmor/kubearmor-client) tool (`karmor log --logFilter all`) that can be used to get the alerts/telemetry. Going forward we may not use `/tmp/kubearmor.log` to store the alerts/telemetry.

viveksahu26:
Ok. But under [Check audit logs ](https://github.com/kubearmor/KubeArmor/blob/main/examples/multiubuntu.md) section do we need to add `kubectl -n kube-system  exec -it kubearmor-7tjlk -- bash -c  "tail  /tmp/kubearmor.log`"  command or  `karmor log --logFilter all` for checking logs.

daemon1024:
@viveksahu26 I believe we should update to something like https://github.com/kubearmor/KubeArmor/blob/main/getting-started/kubearmor_vm.md#get-alerts-for-policies-and-telemetry

nam-jaehyun:
the purpose of /tmp/kubearmor.log is for debugging and quickly checking if KubeArmor works as we expected.
As Rahul and Barun mentioned, we need to use a proper way (e.g., karmor) for alerts and telemetry.

viveksahu26:
Ok, got it. Basically, we have to use standard way to see the logs and i.e. `karmor log --json`. And this also seems much convenient.

PR #626, opened 2022-02-16
Added correct and standard command to see logs

viveksahu26:
Signed-off-by: viveksahu26 <vivekkumarsahu650@gmail.com>
#615 
Corrected command to see logs. It's better and more standard way to see the logs.

### DIFF
--- a/contribution/testing_guide.md
+++ b/contribution/testing_guide.md
@@ -56,45 +56,25 @@ $ kubectl -n [namespace name] exec -it [pod name] -- bash -c [command]
 
 ## 1.6. Check generated alerts
 
-- Log file
+- Log using [kArmor](https://github.com/kubearmor/kubearmor-client) cli tool
 
     ```text
-    $ tail (-f) /tmp/kubearmor.log
+    $ karmor log [flags]
     ```
-
-    If you changed the location of a log file, replace the default file path to your file.
+    
+    flags:
 
     ```text
-    $ tail (-f) [your log file path]
+        --gRPC string        gRPC server information
+    -h, --help               help for log
+        --json               Flag to print alerts and logs in the JSON format
+        --logFilter string   Filter for what kinds of alerts and logs to receive, {policy|system|all} (default "policy")
+        --logPath string     Output location for alerts and logs, {path|stdout|none} (default "stdout")
+        --msgPath string     Output location for messages, {path|stdout|none} (default "none")
     ```
-
-- Log client
-
-    Compile a log client
-
-    ```text
-    $ git clone https://github.com/kubearmor/kubearmor-log-client
-    $ cd kubearmor-log-client
-    ~/kubearmor-log-client$ make
-    ```
-
-    Run the log client
-
-    ```text
-    ~/kubearmor-log-client$ ./kubearmor-log-client (options...)
-    ```
-
-    Log client options:
-
-    ```text
-    -gRPC=[ipaddr:port]             gRPC server information (default: localhost:32767)
-    -msgPath={path|stdout|none}     Output location for KubeArmor's messages (default: none)
-    -logPath={path|stdout|none}     Output location for KubeArmor's alerts and logs (default: stdout)
-    -logFilter={policy|system|all}  Filter for what kinds of alerts and logs to receive (default: policy)
-    -json                           Flag to print messages, alerts, and logs in a JSON format
-    ```
-
-    Note that you will see the messages, alerts, and logs generated right after the log client runs, which means that the log client should be ran before any policy violations happen.
+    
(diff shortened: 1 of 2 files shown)


daemon1024 (approved):
LGTM

nyrahul:
Thanks @viveksahu26 for the contribs.

Can we also remove the `/tmp/kubearmor.log` dependency from [testing_guide.md](https://github.com/kubearmor/KubeArmor/blob/main/contribution/testing_guide.md#16-check-generated-alerts)?

viveksahu26:
Hey @nyrahul , I don't think it would be better idea to completely remove checking logs using `/tmp/kubearmor.log`. Because what if user wants to check in that way. What we can do is, In parallel to that we can add another way i.e. `karmor  log --json` under OR section. Something like:
```
Log file

$ tail (-f) /tmp/kubearmor.log
If you changed the location of a log file, replace the default file path to your file.

$ tail (-f) [your log file path]
```

OR

```
karmor log --json
```
Let's leave on user which method he/she prefers.

nyrahul:
> Let's leave on user which method he/she prefers.

The `/tmp/kubearmor.log` should never be exposed to the user since it is an internal log file for kubearmor and may not be present in the future.

viveksahu26:
Hey @nyrahul , sorry for late response. Btw I was out of station. Does the PR changes seems good or need some more correction?

contribution/testing_guide.md
```diff
@@ -56,16 +56,10 @@ $ kubectl -n [namespace name] exec -it [pod name] -- bash -c [command]
...
 
     ```text
-    $ tail (-f) /tmp/kubearmor.log
-    ```
-
-    If you changed the location of a log file, replace the default file path to your file.
-
-    ```text
-    $ tail (-f) [your log file path]
+    $ tail |  karmor log --json
```
daemon1024: We don't need to tail, `karmor log` will only show future generated alerts not any previous ones.
```suggestion
    $ karmor log [options....]
```
Followed up by available options.

Also I think we can remove the log client part, since karmor handles it.

daemon1024 (changes requested):

contribution/testing_guide.md
viveksahu26: Hmm, yeah that's correct.

contribution/testing_guide.md
```diff
@@ -56,46 +56,15 @@ $ kubectl -n [namespace name] exec -it [pod name] -- bash -c [command]
...
-    $ tail (-f) /tmp/kubearmor.log
-    ```
-
-    If you changed the location of a log file, replace the default file path to your file.
+    $ karmor log (options...)
 
-    ```text
-    $ tail (-f) [your log file path]
+    For example: 
+    $ karmor log --json
```
daemon1024:
```suggestion
    $ karmor log [options...]
```

Followed up by available options, like the earlier log client options were explained. Not everyone wants to see JSON :)

daemon1024 (changes requested):

contribution/testing_guide.md
```diff
@@ -56,28 +56,25 @@ $ kubectl -n [namespace name] exec -it [pod name] -- bash -c [command]
...
+        --json               Flag to print alerts and logs in the JSON format
+        --logFilter string   Filter for what kinds of alerts and logs to receive, {policy|system|all} (default "policy")
+        --logPath string     Output location for alerts and logs, {path|stdout|none} (default "stdout")
+        --msgPath string     Output location for messages, {path|stdout|none} (default "none")
     ```
-
+    
+    Note: Messages, alerts, and logs will be generated right after the karmor runs logs, to avoid any kind of interference, run the above command in other terminal to see logs live and in an interactive way.
+    
     Run the log client
```
daemon1024: Let's remove information about `log-client` since as I mentioned kArmor solves the purpose, we don't need to mention about log client.

contribution/testing_guide.md
```diff
@@ -56,28 +56,25 @@ $ kubectl -n [namespace name] exec -it [pod name] -- bash -c [command]
 
 ## 1.6. Check generated alerts
 
-- Log file
+- Log using [karmor](https://github.com/kubearmor/kubearmor-client) cli tool
```
daemon1024:
```suggestion
- Log using [kArmor](https://github.com/kubearmor/kubearmor-client) cli tool
```

daemon1024 (changes requested):

contribution/testing_guide.md
daemon1024: minor nit :)

contribution/testing_guide.md
viveksahu26: ok,

daemon1024 (approved):
LGTM 🚀

PR #628, opened 2022-02-17
update stable tag to point to the latest version release branch

Ankurk99:


nyrahul (changes requested):
I think we discussed that these changes won't work. We have two scenarios where stable tag should be updated:
1. when `STABLE-RELEASE` file is changed in `main` branch.
2. when a push event happens in the v* branch which is also the branch name in `STABLE-RELEASE` file in `main` branch.

.github/workflows/stable-release.yml
```diff
@@ -18,6 +16,8 @@ jobs:
     steps:
       - name: Checkout KubeArmor code
         uses: actions/checkout@v2
+        with:
+          ref: main
```
nyrahul: The primary problem with this github action is that the stable tag will be updated very frequently ... i.e., on any push to the main branch or any v* branches!

In docker hub this will end up in showing wrong (for us) update date ...

Can you put a check to only do `docker image tag` if the STABLE-RELEASE version in main branch points to the current version branch?

nyrahul (changes requested):

.github/workflows/stable-version-release.yml
```diff
@@ -0,0 +1,43 @@
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    branches:
+      - 'v*'
+
+name: KubeArmor stable release
```
nyrahul: Can you give this a different name?
```suggestion
name: KubeArmor stable release update
```

.github/workflows/stable-version-release.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_AUTHTOK }}
+
+      - name: Push KubeArmor images to Docker
+        if: steps.match.outputs.tag == 'true'
+        run: |
+            if [ ${{ github.ref }} == $value ]; then
```
nyrahul: I think this condition wont be necessary since the tag is set to true and checked for this step anyways.
```suggestion
```

.github/workflows/stable-version-release.yml
```diff
@@ -0,0 +1,43 @@
...
+            value=`cat STABLE-RELEASE`
+            if [ ${{ github.ref }} == $value ]; then
+              docker pull docker.io/kubearmor/kubearmor:$value
+              docker image tag kubearmor/kubearmor:$value kubearmor/kubearmor:stable
+              echo ::set-output name=tag::true
+            else
+              echo ::set-output name=tag::false
+            fi
+
+      - name: Login to Docker Hub
```
nyrahul, line 33:
```suggestion
      - name: Login to Docker Hub
        if: steps.match.outputs.tag == 'true'
```

nyrahul (changes requested):
PFA my review. Thanks

nyrahul (approved):
LGTM. Thanks

PR #602, opened 2022-02-08
Equip KubeArmor with Default Armors

daemon1024:
Ref #595

Issue #531, opened 2021-12-13
Use `client-go` to make requests to kube-apiserver instead of direct API calls


daemon1024:
## Feature Request

Use `client-go` to make requests to kube-apiserver instead of direct API calls

**Is your feature request related to a problem? Please describe the use case.**

Currently we need to have a running proxy for kube-apiserver for KubeArmor in process mode to interact with k8s. 

**Describe the solution you'd like**

Current : https://github.com/kubearmor/KubeArmor/blob/60c105f04f1f3bf6fdc5d90264236e104c153d34/KubeArmor/core/k8sHandler.go#L238

Proposing: 
```
nodeWatcher, err := clientset.CoreV1().Nodes().Watch(metav1.ListOptions{})
```
Ref https://github.com/kubernetes/client-go#how-to-use-it

afzal442:
hey @daemon1024, I got that proposed way to refactor that as you mentioned abv but unsure about incluster perspective.
Let me know how you would send response after getting `nodeWatcher` as example. Thanks.

// @nyrahul

daemon1024:
Hey @afzal442, Thanks for taking interest in this issue.

> but unsure about incluster perspective.

As per my understanding, only difference is that how you authenticate with the server, rest the API is the same. The authentication part is already handled in https://github.com/kubearmor/KubeArmor/blob/main/KubeArmor/core/k8sHandler.go#L97-L102. So it's not an issue.

> Let me know how you would send response after getting nodeWatcher as example.

You can take a look at https://github.com/iximiuz/client-go-examples/tree/main/watch-typed-simple for inspiration. 
Basically we make repeated requests inside a for loop right now. Instead we can loop the channel provided by nodeWatcher.

If you have more queries feel free to ping me in this thread or on our [slack channel](https://kubearmor.herokuapp.com/) ✌🏽

sbin64:
Thanks so much, @daemon1024! That makes sense now. 

> You can take a look at https://github.com/iximiuz/client-go-examples/tree/main/watch-typed-simple for inspiration.

That's really inspired me, that's why I landed on here. 😄 
That is written wrt outside of cluster perspective.

 > Instead we can loop the channel provided by `nodeWatcher`.

```
go func() {
		for event := range nodeWatcher.ResultChan() {
			fmt.Printf(
				"Watch Event: %s %s\n",
				event.Type, event.Object.GetObjectKind().GroupVersionKind().Kind,
			)
		}
	}()
```
[...]

PR #630, opened 2022-02-26
Handle per namespace default posture implementation 

### LINKED ISSUE #595: Implement `defaultArmor` option
## Feature Request

Implement `defaultPosture` option in CLI and CRD

Bug:
KubeArmor doesn't have a default posture in global context in case we only have from Source based policies.
Feature:
Add an option to audit events instead of blocking in case of whitelisting policies.

Add options in KubeArmor to have a configurable default posture.

Initial Goals

- [ ] ~~`-defaultArmor` CLI option~~
	- [ ] Allow ( Current Behaviour )
	- [ ] Block
	- [ ] Audit
- [ ] `defaultFileArmor` - default position for both files/processes 
- [ ] `defaultNetworkArmor`
- [ ] Discuss on how to handle in CRD

cc @nyrahul @nam-jaehyun

daemon1024:
Follows up #602 

Closes #595

```mermaid
flowchart LR
	G[/Watch KSP/] <--> C
	A[/Watch NameSpace Annotation/] --> X(Add/Modified)--> B[[Map Namespace -> Default Posture]] <==> C[Endpoint Default Posture ] --> D[Enforcement Decision]
	A --> Y(Delete) -.Delete Map Element.-> B
	E[/Log Event/] <--> B
	E --> F[Decide Posture Log/Alert]
	B <-.On Delete.-> H[[Global Posture]]
	C <-.If not in map.-> H
```

### DIFF
--- a/KubeArmor/core/kubeArmor.go
+++ b/KubeArmor/core/kubeArmor.go
@@ -66,6 +66,10 @@ type KubeArmorDaemon struct {
 	HostSecurityPolicies     []tp.HostSecurityPolicy
 	HostSecurityPoliciesLock *sync.RWMutex
 
+	//DefaultPosture (namespace -> postures)
+	DefaultPostures     map[string]tp.DefaultPosture
+	DefaultPosturesLock *sync.Mutex
+
 	// container id -> (host) pid
 	ActivePidMap     map[string]tp.PidMap
 	ActiveHostPidMap map[string]tp.PidMap
@@ -113,6 +117,9 @@ func NewKubeArmorDaemon() *KubeArmorDaemon {
 
 	dm.HostSecurityPolicies = []tp.HostSecurityPolicy{}
 	dm.HostSecurityPoliciesLock = new(sync.RWMutex)
+
+	dm.DefaultPostures = map[string]tp.DefaultPosture{}
+	dm.DefaultPosturesLock = new(sync.Mutex)
 
 	dm.ActivePidMap = map[string]tp.PidMap{}
 	dm.ActiveHostPidMap = map[string]tp.PidMap{}
@@ -506,6 +513,10 @@ func KubeArmor() {
 		// watch security policies
 		go dm.WatchSecurityPolicies()
 		dm.Logger.Print("Started to monitor security policies")
+
+		// watch default posture
+		go dm.WatchDefaultPosture()
+		dm.Logger.Print("Started to monitor per-namespace default posture")
 	}
 
 	if dm.K8sEnabled && cfg.GlobalCfg.HostPolicy {
--- a/KubeArmor/core/kubeUpdate.go
+++ b/KubeArmor/core/kubeUpdate.go
@@ -4,6 +4,7 @@
 package core
 
 import (
+	"context"
 	"encoding/json"
 	"io"
 	"io/ioutil"
@@ -16,6 +17,9 @@ import (
 	cfg "github.com/kubearmor/KubeArmor/KubeArmor/config"
 	kg "github.com/kubearmor/KubeArmor/KubeArmor/log"
 	tp "github.com/kubearmor/KubeArmor/KubeArmor/types"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
 )
 
 // ================= //
@@ -218,6 +222,20 @@ func (dm *KubeArmorDaemon) UpdateEndPointWithPod(action string, pod tp.K8sPod) {
 		}
 		dm.ContainersLock.Unlock()
 
+		dm.DefaultPosturesLock.Lock()
+		if val, ok := dm.DefaultPostures[newPoint.NamespaceName]; ok {
+			newPoint.DefaultPosture = val
+		} else {
+			globalDefaultPosture := tp.DefaultPosture{
+				FileAction:         cfg.GlobalCfg.DefaultFilePosture,
+				NetworkAction:      cfg.GlobalCfg.DefaultNetworkPosture,
+				CapabilitiesAction: cfg.GlobalCfg.DefaultCapabilitiesPosture,
+			}
+			dm.DefaultPostures[newPoint.NamespaceName] = globalDefaultPosture
+			newPoint.DefaultPosture = globalDefaultPosture
+		}
+		dm.DefaultPosturesLock.Unlock()
+
 		// update selinux profile names to the endpoint
 		for k, v := range pod.Annotations {
 			if strings.HasPrefix(k, "kubearmor-selinux") {
@@ -329,6 +347,20 @@ func (dm *KubeArmorDaemon) UpdateEndPointWithPod(action string, pod tp.K8sPod) {
 			dm.Containers[containerID] = container
 		}
 		dm.ContainersLock.Unlock()
+
+		dm.DefaultPosturesLock.Lock()
+		if val, ok := dm.DefaultPostures[newEndPoint.NamespaceName]; ok {
+			newEndPoint.DefaultPosture = val
+		} else {
+			globalDefaultPosture := tp.DefaultPosture{
+				FileAction:         cfg.GlobalCfg.DefaultFilePosture,
+				NetworkAction:      cfg.GlobalCfg.DefaultNetworkPosture,
+				CapabilitiesAction: cfg.GlobalCfg.DefaultCapabilitiesPosture,
+			}
+			dm.DefaultPostures[newEndPoint.NamespaceName] = globalDefaultPosture
+			newEndPoint.DefaultPosture = globalDefaultPosture
+		}
+		dm.DefaultPosturesLock.Unlock()
 
 		// update selinux profile names to the endpoint
 		for k, v := range pod.Annotations {
@@ -1693,6 +1725,87 @@ func (dm *KubeArmorDaemon) restoreKubeArmorHostPolicies() {
 			dm.UpdateHostSecurityPolicies()
 		} else {
 			kg.Warn("No policies found for restoration")
+		}
+	}
+}
+
+// WatchDefaultPosture Function
+func (dm *KubeArmorDaemon) WatchDefaultPosture() {
+	nsWatcher, err := K8s.K8sClient.CoreV1().Namespaces().Watch(context.Background(), metav1.ListOptions{})
+	defer nsWatcher.Stop()
+	if err == nil {
+		for resp := range nsWatcher.ResultChan() {
+			if resp.Type == watch.Modified || resp.Type == watch.Added {
+				if ns, ok := resp.Object.(*corev1.Namespace); ok {
+					defaultPosture := tp.DefaultPosture{
+						FileAction:         validateDefaultPosture("kubearmor-file-posture", ns, cfg.GlobalCfg.DefaultFilePosture),
+						NetworkAction:      validateDefaultPosture("kubearmor-network-posture", ns, cfg.GlobalCfg.DefaultNetworkPosture),
+						CapabilitiesAction: validateDefaultPosture("kubearmor-capabilities-posture", ns, cfg.GlobalCfg.DefaultCapabilitiesPosture),
+					}
+					dm.UpdateDefaultPosture(string(resp.Type), ns.Name, defaultPosture)
+
+				}
+			} else if resp.Type == watch.Deleted {
+				if ns, ok := resp.Object.(*corev1.Namespace); ok {
+					dm.UpdateDefaultPosture(string(resp.Type), ns.Name, tp.DefaultPosture{})
+				}
+			}
+		}
+	}
+}
+
+func validateDefaultPosture(key string, ns *corev1.Namespace, defaultPosture string) string {
+	if posture, ok := ns.Annotations[key]; ok {
+		if posture == "audit" || posture == "Audit" {
+			return "audit"
+		} else if posture == "block" || posture == "Block" {
+			return "block"
+		}
+		// Invalid Annotation Value, Updating the value to global default
+		ns.Annotations[key] = defaultPosture
+		updatedNS, err := K8s.K8sClient.CoreV1().Namespaces().Update(context.Background(), ns, metav1.UpdateOptions{})
+		if err != nil {
+			kg.Warnf("Error updating invalid default posture annotation for %v", updatedNS)
+		}
+	}
+	return defaultPosture
+}
+
+// UpdateDefaultPosture Function
+func (dm *KubeArmorDaemon) UpdateDefaultPosture(action string, namespace string, defaultPosture tp.DefaultPosture) {
+	dm.EndPointsLock.Lock()
+	defer dm.EndPointsLock.Unlock()
+
+	dm.DefaultPosturesLock.Lock()
+	defer dm.DefaultPosturesLock.Unlock()
+
+	if action == "DELETED" {
+		delete(dm.DefaultPostures, namespace)
+	}
+
+	dm.DefaultPostures[namespace] = defaultPosture
+
+	for idx, endPoint := range dm.EndPoints {
+		// update a security policy
+		if namespace == endPoint.NamespaceName {
+			if dm.EndPoints[idx].DefaultPosture == defaultPosture {
+				continue
+			}
+
+			dm.Logger.UpdateDefaultPosture(action, namespace, defaultPosture)
+
+			dm.EndPoints[idx].DefaultPosture = defaultPosture
+			dm.Logger.Printf("Updating default posture for %s with %v/%v", endPoint.EndPointName, dm.EndPoints[idx].DefaultPosture, dm.DefaultPostures[namespace])
+
+			if cfg.GlobalCfg.Policy {
+				// update security policies
+				if dm.RuntimeEnforcer != nil {
+					if dm.EndPoints[idx].PolicyEnabled == tp.KubeArmorPolicyEnabled {
+						// enforce security policies
+						dm.RuntimeEnforcer.UpdateSecurityPolicies(dm.EndPoints[idx])
+					}
+				}
+			}
 		}
 	}
 }
--- a/KubeArmor/feeder/policyMatcher.go
+++ b/KubeArmor/feeder/policyMatcher.go
@@ -206,6 +206,20 @@ func (fd *Feeder) newMatchPolicy(policyEnabled int, policyName, src string, mp i
 	return match
 }
 
+// UpdateDefaultPosture Function
+func (fd *Feeder) UpdateDefaultPosture(action string, namespace string, defaultPosture tp.DefaultPosture) {
+
+	fd.DefaultPosturesLock.Lock()
+	defer fd.DefaultPosturesLock.Unlock()
+
+	if action == "DELETED" {
+		delete(fd.DefaultPostures, namespace)
+	}
+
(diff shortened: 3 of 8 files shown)


nyrahul:
@daemon1024 , few points:
1. Can you specify a sample policy for applying security posture at namespace level?
2. we need a way to show defaultPosture and namespacePosture using karmor cli tool.

daemon1024:
Added a document about default posture with details about configuration and example at https://github.com/daemon1024/KubeArmor/blob/default-posture-ns/getting-started/default_posture.md

Ref https://github.com/kubearmor/KubeArmor/pull/602#pullrequestreview-901588739

getting-started/default_posture.md
```diff
@@ -0,0 +1,149 @@
+# Default Posture
+
+KubeArmor supports configurable default mode of operations. Default Posture comes into picture when there's atleast one `Allow` policy i.e. KubeArmor is handling policies in whitelisting manner ( More about this in [Consideration in Policy Action](https://github.com/kubearmor/KubeArmor/blob/event-auditor/getting-started/consideration_in_policy_action.md) ).
```
nyrahul:
```suggestion
KubeArmor supports configurable default security posture. The security posture could be allow/audit/deny. Default Posture is used when there's atleast one `Allow` policy for the given deployment i.e. KubeArmor is handling policies in whitelisting manner (more about this in [Considerations in Policy Action](https://github.com/kubearmor/KubeArmor/blob/event-auditor/getting-started/consideration_in_policy_action.md) ).
```

getting-started/default_posture.md
```diff
@@ -0,0 +1,149 @@
...
+
+There are two default mode of operations available `block` and `audit`. `block` mode blocks all the operations that are not allowed in the policy. `audit` generates telemetry events for operations that would have been blocked otherwise.
```
nyrahul, line 5: Shouldn't there be three options? allow/block/audit?

Allow mode would simply allow all the operations and only the explicitly denied operations would be blocked.

getting-started/default_posture.md
```diff
@@ -0,0 +1,149 @@
...
+  -defaultNetworkPosture string
+    	configuring default enforcement action in global network context [audit,block] (default "block")
+  -defaultCapabilitiesPosture string
+    	configuring default enforcement action in global capability context [audit,block] (default "block")
+```
+
+### Namespace Default Posture
+
+We use namespace annotations to configure default posture per namespace. Supported annotations keys are `kubearmor-file-posture`,`kubearmor-network-posture` and `kubearmor-capabilities-posture` with values `block` or `audit`. If a namespace is annotated with a supported key and an invalid value ( like `kubearmor-file-posture=invalid`), KubeArmor will update the value with the global default posture ( i.e. to `kubearmor-file-posture=block`).
+
```
nyrahul, line 34: It would be better to give an example using a sample yaml with annotation.

nyrahul (changes requested):
Great work! This feature can help users onboard kubearmor security in a methodical way.

getting-started/default_posture.md
daemon1024: > Allow mode would simply allow all the operations and only the explicitly denied operations would be blocked.

Why would one apply an 'Allow Policy' if this would be the case 😅 Won't this be the same as not applying any policy. Noting Default Posture only comes into picture when we have an Allow Policy.

KubeArmor/core/kubeArmor.go
```diff
@@ -105,6 +109,9 @@ func NewKubeArmorDaemon() *KubeArmorDaemon {
 	dm.Containers = map[string]tp.Container{}
 	dm.ContainersLock = new(sync.RWMutex)
 
+	dm.DefaultPostures = map[string]tp.DefaultPosture{}
```
nam-jaehyun: move 112-113 below dm.HostSecurityPoliciesLock = new(sync.RWMutex)

KubeArmor/core/kubeArmor.go
```diff
@@ -505,6 +512,10 @@ func KubeArmor() {
 
 		// watch security policies
 		go dm.WatchSecurityPolicies()
+
+		// watch default posture
+		go dm.WatchDefaultPosture()
```
nam-jaehyun, line 518: move 516-517 below "dm.Logger.Print("Started to monitor security policies")"
add "dm.Logger.Print("Started to monitor per-namespace default posture")"

KubeArmor/core/kubeUpdate.go
```diff
@@ -782,6 +814,45 @@ func (dm *KubeArmorDaemon) UpdateSecurityPolicy(action string, secPolicy tp.Secu
 	}
 }
 
+// UpdateDefaultPosture Function
+func (dm *KubeArmorDaemon) UpdateDefaultPosture(action string, namespace string, secPolicy tp.DefaultPosture) {
```
nam-jaehyun: secPolicy -> defaultPosture
secPolicy might be confusing with actual policy.

KubeArmor/core/kubeUpdate.go
```diff
@@ -782,6 +814,45 @@ func (dm *KubeArmorDaemon) UpdateSecurityPolicy(action string, secPolicy tp.Secu
...
+	if action == "DELETED" {
+		delete(dm.DefaultPostures, namespace)
+	}
+
+	dm.DefaultPostures[namespace] = secPolicy
+
+	for idx, endPoint := range dm.EndPoints {
+		// update a security policy
+		if namespace == endPoint.NamespaceName {
+			if dm.EndPoints[idx].DefaultPosture == secPolicy && dm.DefaultPostures[namespace] == secPolicy {
```
nam-jaehyun: why does it need to compare both endpoints[idx].defaultposture and dm.defaultpostures[namespace]?

KubeArmor/core/kubeUpdate.go
```diff
@@ -782,6 +814,45 @@ func (dm *KubeArmorDaemon) UpdateSecurityPolicy(action string, secPolicy tp.Secu
 	}
 }
 
+// UpdateDefaultPosture Function
```
nam-jaehyun: move this function to 1170 (above validateDefaultPosture)

KubeArmor/core/kubeUpdate.go
```diff
@@ -1696,3 +1767,45 @@ func (dm *KubeArmorDaemon) restoreKubeArmorHostPolicies() {
...
+		if err != nil {
+			kg.Warnf("Error updating invalid default posture annotation for %v", updatedNS)
+		}
+	}
+	return defaultPosture
+}
+
+// WatchDefaultPosture Function
+func (dm *KubeArmorDaemon) WatchDefaultPosture() {
+	nsWatcher, err := K8s.K8sClient.CoreV1().Namespaces().Watch(context.Background(), metav1.ListOptions{})
```
nam-jaehyun, line 1734: the way to watch k8s namespaces is different compared to the above watch functions
keep the consistency

KubeArmor/core/kubeUpdate.go
```diff
@@ -1696,3 +1767,45 @@ func (dm *KubeArmorDaemon) restoreKubeArmorHostPolicies() {
...
 	}
 }
+
+func validateDefaultPosture(key string, ns *corev1.Namespace, defaultPosture string) string {
+	if posture, ok := ns.Annotations[key]; ok {
+		if posture == "audit" || posture == "Audit" {
+			return "audit"
+		} else if posture == "block" || posture == "Block" {
+			return "block"
+		}
```
nam-jaehyun: what if given posture is not audit nor block?

KubeArmor/core/kubeUpdate.go
```diff
@@ -1696,3 +1767,45 @@ func (dm *KubeArmorDaemon) restoreKubeArmorHostPolicies() {
...
+		// Invalid Annotation Value, Updating the value to global default
+		ns.Annotations[key] = defaultPosture
+		updatedNS, err := K8s.K8sClient.CoreV1().Namespaces().Update(context.Background(), ns, metav1.UpdateOptions{})
```
nam-jaehyun: not sure if this is a right approach

KubeArmor/feeder/policyMatcher.go
```diff
@@ -206,6 +206,20 @@ func (fd *Feeder) newMatchPolicy(policyEnabled int, policyName, src string, mp i
 	return match
 }
 
+// UpdateDefaultPosture Function
+func (fd *Feeder) UpdateDefaultPosture(action string, namespace string, secPolicy tp.DefaultPosture) {
```
nam-jaehyun: secPolicy -> defaultPosture

KubeArmor/feeder/policyMatcher.go
```diff
@@ -1040,23 +1054,36 @@ func (fd *Feeder) UpdateMatchedPolicy(log tp.Log) tp.Log {
 				}
 			}
 
+			fd.DefaultPosturesLock.Lock()
```
nam-jaehyun, line 1057: why is this logic here?
why do we need to update fd.DefaultPostures during log generation?
does it cause additional overhead?

getting-started/default_posture.md
```diff
@@ -0,0 +1,149 @@
...
+KubeArmor supports configurable default mode of operations. Default Posture comes into picture when there's atleast one `Allow` policy i.e. KubeArmor is handling policies in whitelisting manner ( More about this in [Consideration in Policy Action](https://github.com/kubearmor/KubeArmor/blob/event-auditor/getting-started/consideration_in_policy_action.md) ).
+
+There are two default mode of operations available `block` and `audit`. `block` mode blocks all the operations that are not allowed in the policy. `audit` generates telemetry events for operations that would have been blocked otherwise.
+
+KubeArmor has 4 types of resources: Process, File, Network and Capabilities. Default Posture is configurable for each of the resources seperately except Process. Process based operations are treated under File resource only.
+## Configuring Default Posture
+
+### Global Default Posture
+
+Global default posture is configured using configuration options provided in KubeArmor binary.
```
nam-jaehyun: where can we configure global default posture?
arguments? config file?
would be better to give an example.

getting-started/default_posture.md
```diff
@@ -0,0 +1,149 @@
...
+### Namespace Default Posture
+
+We use namespace annotations to configure default posture per namespace. Supported annotations keys are `kubearmor-file-posture`,`kubearmor-network-posture` and `kubearmor-capabilities-posture` with values `block` or `audit`. If a namespace is annotated with a supported key and an invalid value ( like `kubearmor-file-posture=invalid`), KubeArmor will update the value with the global default posture ( i.e. to `kubearmor-file-posture=block`).
+
+## Example
+
+Let's start KubeArmor with configuring default network posture to audit.
+
+```sh
+sudo -E ./kubearmor -defaultNetworkPosture=audit
```
nam-jaehyun: people would be not familiar with the command-line-based execution.
what about using kuberamor.yaml -> command?

KubeArmor/core/kubeUpdate.go
```diff
@@ -782,6 +814,45 @@ func (dm *KubeArmorDaemon) UpdateSecurityPolicy(action string, secPolicy tp.Secu
...
+	dm.DefaultPostures[namespace] = secPolicy
+
+	for idx, endPoint := range dm.EndPoints {
+		// update a security policy
+		if namespace == endPoint.NamespaceName {
+			if dm.EndPoints[idx].DefaultPosture == secPolicy && dm.DefaultPostures[namespace] == secPolicy {
+				continue
+			}
+
+			dm.Logger.UpdateDefaultPosture(action, namespace, secPolicy)
```
nam-jaehyun: is it possible to use dm.DefaultPostures instead of maintaining fd.DefaultPostures?

KubeArmor/core/kubeUpdate.go
```diff
@@ -330,6 +348,20 @@ func (dm *KubeArmorDaemon) UpdateEndPointWithPod(action string, pod tp.K8sPod) {
...
+		dm.DefaultPosturesLock.Lock()
+		if val, ok := dm.DefaultPostures[newEndPoint.NamespaceName]; ok {
+			newEndPoint.DefaultPosture = val
+		} else {
+			globalDefaultPosture := tp.DefaultPosture{
+				FileAction:         cfg.GlobalCfg.DefaultFilePosture,
+				NetworkAction:      cfg.GlobalCfg.DefaultNetworkPosture,
+				CapabilitiesAction: cfg.GlobalCfg.DefaultCapabilitiesPosture,
+			}
+			dm.DefaultPostures[newEndPoint.NamespaceName] = globalDefaultPosture
```
nam-jaehyun, line 360: why do we need to update dm.DefaultPostures here?

KubeArmor/core/kubeUpdate.go
```diff
@@ -218,6 +222,20 @@ func (dm *KubeArmorDaemon) UpdateEndPointWithPod(action string, pod tp.K8sPod) {
...
+		dm.DefaultPosturesLock.Lock()
+		if val, ok := dm.DefaultPostures[newPoint.NamespaceName]; ok {
+			newPoint.DefaultPosture = val
+		} else {
+			globalDefaultPosture := tp.DefaultPosture{
+				FileAction:         cfg.GlobalCfg.DefaultFilePosture,
+				NetworkAction:      cfg.GlobalCfg.DefaultNetworkPosture,
+				CapabilitiesAction: cfg.GlobalCfg.DefaultCapabilitiesPosture,
+			}
+			dm.DefaultPostures[newPoint.NamespaceName] = globalDefaultPosture
```
nam-jaehyun, line 234: why do we need to update dm.DefaultPostures here?

KubeArmor/core/kubeUpdate.go
```diff
@@ -1696,3 +1767,45 @@ func (dm *KubeArmorDaemon) restoreKubeArmorHostPolicies() {
...
+	if err == nil {
+		for resp := range nsWatcher.ResultChan() {
+			if resp.Type == watch.Modified || resp.Type == watch.Added {
+				if ns, ok := resp.Object.(*corev1.Namespace); ok {
+					defaultPosture := tp.DefaultPosture{
+						FileAction:         validateDefaultPosture("kubearmor-file-posture", ns, cfg.GlobalCfg.DefaultFilePosture),
+						NetworkAction:      validateDefaultPosture("kubearmor-network-posture", ns, cfg.GlobalCfg.DefaultNetworkPosture),
+						CapabilitiesAction: validateDefaultPosture("kubearmor-capabilities-posture", ns, cfg.GlobalCfg.DefaultCapabilitiesPosture),
+					}
+					dm.UpdateDefaultPosture(string(resp.Type), ns.Name, defaultPosture)
```
nam-jaehyun, line 1745: why don't we update dm.DefaultPostures in this function? why updating that struct during endpoint updates?

nam-jaehyun (commented):
please check my comments.

KubeArmor/core/kubeUpdate.go
daemon1024: WatchDefaultPosture only updates the values in `dm.DefaultPosture` when there are annotation related updates. When there is an endpoint related update but that namespace has no default posture set due to no annotations, we set the namespace annotation to the global default posture for this endpoint and all future endpoints in this namespace.

KubeArmor/core/kubeUpdate.go
daemon1024: We are planning to migrate above watch functions to use client go as well, WIP in #531

KubeArmor/feeder/policyMatcher.go
daemon1024: WatchDefaultPosture only updates the values in `fd.DefaultPosture` when there are annotation related updates. When there is an log related update but that namespace has no default posture set due to no annotations, we set the namespace annotation to the global default posture for this log and all future log events in this namespace.

Should not have any significant overhead because we need to read for values anyway, if the value is already set we won't be assigning values here after that. The assignment happens once per namespace.

getting-started/default_posture.md
daemon1024: I followed up with an example using flags, will update it with configuration file.

KubeArmor/core/kubeUpdate.go
daemon1024: addressed in the above review comment https://github.com/kubearmor/KubeArmor/pull/630#discussion_r835704051

getting-started/default_posture.md
nyrahul: Sure, I see what your saying. But consider following scenario:

A user applies a block and an allow policy. Now the default posture could either be audit/deny. But just because user ended up adding an Allow policy, there is no way for the user to simply Allow everything else with getting alerts unless the user deletes the Allow policy. That is what I have issues with.

KubeArmor/core/kubeUpdate.go
daemon1024: dm construct doesn't seem to be available to feeder.

getting-started/default_posture.md
daemon1024: Even in this case with the supposed allow mode set, What would be the expected behavior just of the Allow policy? From what I understand it would be same as removing it?

getting-started/default_posture.md
daemon1024: On updating annotations using YAML 
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: multiubuntu
  annotations:
    kubearmor-network-posture: audit
```
`kubectl apply -f ...`
If we only update metadata (annotation), it prints `unchanged` despite updating the annotation value anyway. So it feels more right to use `kubectl annotate` since it responds accordingly.

WDYT?

nam-jaehyun: 'kubectl annotate' would be a good choice.

KubeArmor/core/kubeUpdate.go
```diff
@@ -1696,3 +1767,45 @@ func (dm *KubeArmorDaemon) restoreKubeArmorHostPolicies() {
 		}
 	}
 }
+
+// WatchDefaultPosture Function
+func (dm *KubeArmorDaemon) WatchDefaultPosture() {
```
nam-jaehyun, line 1733: let's make an order like the below to make a group of the functions correlated with each other at the end of this file.

UpdateDefaultPosture()
validateDefaultPosture()
WatchDefaultPosture()

getting-started/default_posture.md
```diff
@@ -0,0 +1,149 @@
...
+namespace/multiubuntu annotated
+~❯❯❯  kubectl describe ns multiubuntu
+Name:         multiubuntu
+Labels:       kubernetes.io/metadata.name=multiubuntu
+Annotations:  kubearmor-network-posture: audit
+Status:       Active
+```
+We can see that, annotation value was automatically updated to audit since that was global mode of operation for network in the KubeArmor configuration.
+
+
```
nam-jaehyun: remove empty lines

nam-jaehyun:
For DefaultPosture structures, RWLocks are used.
However, I couldn't see RLock and RUnlock. Where can we use those read locks?

daemon1024:
- Updated doc to use configuration file
- Using regular mutex lock instead of RWMutex now

nyrahul (approved):

getting-started/default_posture.md
nyrahul: Yes if the user removes the allow policy then it works fine. But asking user to remove a policy for something else to work was the point. Anyways, we can live with this behaviour. Thanks for the discussion.

PR #631, opened 2022-02-28
Kubearmor policies are not getting enforced on reboot

### LINKED ISSUE #605: policy enforcement issue when kubearmor is restarted in systemd mode
## Bug Report

When kubearmor is operating in systemd mode, and if someone applies a set of policies and then restarts the kubearmor then one has to reapply the set of policies manually.

**General Information**

systemd (non-k8s) mode 

Any supported kernel ver

systemd

**To Reproduce**

1. start kubearmor in systemd mode
2. apply policy1
3. restart kubearmor
4. policy1 is no more enforced

Note that this works seamlessly in k8s mode i.e., even if the kubearmor pod is deleted and restarted the previously enforced policies remain enforced.

**Expected behavior**

Previously enforced policies should remain enforced on restart in systemd mode.

CC: @seswarrajan

seswarrajan:


KubeArmor/config/config.go
```diff
@@ -32,6 +32,9 @@ type KubearmorConfig struct {
 	CoverageTest bool // Enable/Disable Coverage Test
 }
 
+// PolicyDir policy dir path for host policies backup
+const PolicyDir string = "/opt/kubearmor/policies/"
```
nam-jaehyun, line 36: Q1: we know the systemd mode uses /opt/kubearmor, but is it okay to have a static path in the code?
I think we need to consider k8s since we use host policies in that env too.

KubeArmor/core/kubeArmor.go
```diff
@@ -312,6 +314,35 @@ func GetOSSigChannel() chan os.Signal {
 	return c
 }
 
+// ======================== //
+// == HostPolicy Restore == //
+// ======================== //
+
+func (dm *KubeArmorDaemon) restoreKubeArmorHostPolicies() {
```
nam-jaehyun: can you put this function below backupKubeArmorHostPolicy() in kubeUpdate.go?

KubeArmor/config/config.go
seswarrajan: But @nam-jaehyun . In that case also do we expect to store the policies ? we will be storing policies only on kvmagent mode and systemd mode right ?

nam-jaehyun: according to this code, we will also store policies in k8s if hostPolicy is enabled.
However, /opt/kubearmor won't exist in the k8s env.

seswarrajan: Understood @nam-jaehyun 

We can check if policy is not enabled and then backup the policies. That should be a solution.

KubeArmor/core/kubeUpdate.go
```diff
@@ -1617,3 +1624,54 @@ func (dm *KubeArmorDaemon) WatchHostSecurityPolicies() {
...
+// ================================= //
+// == HostPolicy Backup & Restore == //
+// ================================= //
+
+// backupKubeArmorHostPolicy Function
+func (dm *KubeArmorDaemon) backupKubeArmorHostPolicy(policy tp.HostSecurityPolicy) {
+	// List all policies files from "/opt/kubearmor/policies" path
+	if _, err := os.Stat(cfg.PolicyDir); err != nil {
+		dm.Logger.Errf(err.Error())
+		return
```
daemon1024: Let's create the directory, if it doesn't exist while backing up

KubeArmor/core/kubeUpdate.go
```diff
@@ -1617,3 +1624,54 @@ func (dm *KubeArmorDaemon) WatchHostSecurityPolicies() {
...
+		}
+	}
+}
+
+func (dm *KubeArmorDaemon) restoreKubeArmorHostPolicies() {
+	// List all policies files from "/opt/kubearmor/policies" path
+	if _, err := os.Stat(cfg.PolicyDir); err != nil {
+		if err = os.MkdirAll(cfg.PolicyDir, 0700); err != nil {
+			kg.Warnf("%v", err.Error())
+			return
```
daemon1024: Can simply log nothing to be restored...

KubeArmor/core/kubeUpdate.go
```diff
@@ -1617,3 +1624,54 @@ func (dm *KubeArmorDaemon) WatchHostSecurityPolicies() {
...
+	if policyFiles, err := ioutil.ReadDir(cfg.PolicyDir); err == nil {
+		for _, file := range policyFiles {
+			if data, err := ioutil.ReadFile(cfg.PolicyDir + file.Name()); err == nil {
+				var hostPolicy tp.HostSecurityPolicy
+				if err := json.Unmarshal(data, &hostPolicy); err == nil {
+					dm.HostSecurityPolicies = append(dm.HostSecurityPolicies, hostPolicy)
+				}
+			}
+		}
+
```
daemon1024, line 1672: if policyFiles count is 0 let's log, no existing host policies to be restored...

daemon1024 (changes requested):

KubeArmor/core/kubeUpdate.go
```diff
@@ -1617,3 +1624,56 @@ func (dm *KubeArmorDaemon) WatchHostSecurityPolicies() {
...
+// ================================= //
+// == HostPolicy Backup & Restore == //
+// ================================= //
+
+// backupKubeArmorHostPolicy Function
+func (dm *KubeArmorDaemon) backupKubeArmorHostPolicy(policy tp.HostSecurityPolicy) {
+	// Check for "/opt/kubearmor/policies" path. If dir not found, create the same
+	if _, err := os.Stat(cfg.PolicyDir); err != nil {
+		if err = os.MkdirAll(cfg.PolicyDir, 0700); err != nil {
+			kg.Warnf("No policies restored")
```
daemon1024: We are backing up, no need to comment about restoration here.

daemon1024 (commented):
Small nit. Rest LGTM. Can we squash the commits as well?

daemon1024:
LGTM, let's squash the commits and we are good to merge it then 👌

seswarrajan:
> 

Hi @daemon1024 
Once approved, we can squash and merge.

daemon1024 (approved):

PR #639, opened 2022-03-03
Fix config behaviour

daemon1024:
Explicitly calling set, overrides config/env values with default values. This change fixes that behaviour. Note unlike the standard precedence order, value of flags will be overwritten by config/env


seswarrajan (approved):
LGTM

nyrahul (approved):

Issue #360, opened 2021-09-16
Applying KubeArmor pod annotations


nyrahul:
## Feature Request

**Short Description**

Currently, KubeArmor watches for deployments and uses the rolling update feature to apply the required annotations. There are following annotations that are of interest.
```
Annotations:  container.apparmor.security.beta.kubernetes.io/ubuntu-1-container: localhost/kubearmor-multiubuntu-ubuntu-1-container
              kubearmor-policy: enabled
              kubearmor-visibility: process,file,network
```
* the first annotation enables AppArmor in the context of the pod
* the second annotation enables kubearmor based policy "enforcement" on the pod. Another option is to use "audited" here to only get the audit events for the pod.
* the third annotation enables visibility in the context of the pod for process, file, and network.

**Is your feature request related to a problem? Please describe the use case.**

Currently, KubeArmor does not have a way to apply annotations in the context of individual pods i.e, pods created without deployments.

Thus if the user applies a policy, the policy apply works but the actual enforcement does not. There is no way for the user to debug this.

**Describe the solution you'd like**

* we could rid kubearmor of all the annotations-related code and just use an admission controller to apply the annotation using mutatingwebhook.
* This would work for all types of pods (including created directly or created through deployments)

However, this approach needs proper analysis of the dependencies.

**Describe alternatives you've considered**

* Just do documentation update: We do not expect pods to be created without deployments in the real world. However, this assumption may be wrong.
* Use the Kubearmor client tool to show the status. In the status, we can show the pods without deployment and highlight the risk.

CC: @nam-jaehyun

nthnieljson:
@nyrahul Hello, I am interested in this project and want to take this as my LFX mentorship project (referring to https://mentorship.lfx.linuxfoundation.org/project/81f0d863-30ca-43b5-a368-fd422ec681a6). I have already applied to the LFX mentorship website and also submitted my prerequisite tasks. Thank you!

nyrahul:
> @nyrahul Hello, I am interested in this project and want to take this as my LFX mentorship project (referring to https://mentorship.lfx.linuxfoundation.org/project/81f0d863-30ca-43b5-a368-fd422ec681a6). I have already applied to the LFX mentorship website and also submitted my prerequisite tasks. Thank you!

Great to see your interest in solving this problem. We would be happy discuss the problem statement and possible design choices we had considered.

achrefbensaad:
[...]

PR #640, opened 2022-03-03
[DO NOT MERGE] Annotation admission controller proposal

achrefbensaad:
# Annotations admission controller

## Context

We aim to decople all annotation related code from kubearmor and delegate the annotation process to an admission controller.

See [issue](https://github.com/kubearmor/KubeArmor/issues/360) for more informations

## How pods can be created on a K8S cluster?

Pod can be create via:

- Deployment
- ReplicaSet
- StatefulSet
- CronJob
- Job
- Directly via pod manifest (imperative creation)
- Staticly([Static pods](https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/))

## Expected behavior

When the admission controller receives a mutation  webhook from the API server, the admission controller should check if the pod contains annotaions or not.

if it is already annotated, no further action is required.

if its not, the admission controller should perform parent detection logic to determine the origin resource that caused the pod to be created and annotate it with kubearmor annotations.

In order to acheive this, we need to refactor the `WatchK8sPods` function into an admission controller, in addition we need to create patch logic for indicidual pods (static, created imperativiely), StatefullSets, ... . 

The patch logic can be devided into two main categories.

- Pod patch logic: under `metadata.annotations`

- PodGenerators (Deployment, ReplicaSet, ...) patch logic: under `spec.template.metadata.annotations`, for this we can repourpose the already existant code of Deployment patch e.g. `PatchDeploymentWithAppArmorAnnotations`. The repourpose is made possible here as all "PodGenerators" follows the same "Generic Yaml structure".

This method ensures minimal patches being applied to resources.


nyrahul:
This sounds good to me. I guess you (@achrefbensaad) can prepare a set of tasklist based on this that can be checked as we progress. @nam-jaehyun, @daemon1024 can you too pls review this? Thanks

achrefbensaad:
Hey @nyrahul , I suggest the following tasks.

 - [x] `task-1`: Scaffold a controller that watches for Pod events
 - [x] `task-2`: Impliment & test Pod annotations logic
 - [x] `task-3`: Impliment & test Pod Generators annotations logic
 - [ ] `task-4`: Add dynamic tests for non coverd Pod Generators (StatefulSet, CronJob, ...)
 - [x] `task-5`: Remove all annotations related code from under `KubeArmor/core`
 - [ ] `task-6`: Update `karmor`to install the controller

After completing `task-3` all already existant dynamic tests should succeed with no further mofications, as this change should be user transparent.

PR #643, opened 2022-03-07
pkg: update path and dir validation regex

### LINKED ISSUE #641: policy rule validation failure for a valid policy
## Bug Report

Trying to apply a policy with valid `file.matchPaths` rule, but it fails.

```yaml
kind: KubeArmorPolicy
metadata:
  name: autopol-system-1760649700
  namespace: default
spec:
  severity: 1
  selector:
    matchLabels:
      app: cartservice
  file:
    matchPaths:
    - path: /usr/lib/libstdc++.so.6.0.28
      fromSource:
      - path: /app/cartservice
  action: Block
```

Policy validation fails during policy apply giving following error:
```
❯ k apply -f sample.yaml
The KubeArmorPolicy "autopol-system-1760649700" is invalid: spec.file.matchPaths.path: Invalid value: "/usr/lib/libstdc++.so.6.0.28": spec.file.matchPaths.path in body should match '^\/([A-z0-9-_.]+\/)*([A-z0-9-_.]+)$'
```
> **The problem seems to be with the filename `libstdc++.so.6.0.28`. If I remove `++` from the filename then the policy apply works.**

**General Information**

any
any
any

**To Reproduce**

1. Install sample k8s cluster with kubearmor daemonset (follow [deployment-guide](https://github.com/kubearmor/KubeArmor/blob/main/getting-started/deployment_guide.md)).
2. Apply above yaml policy `kubectl apply -f sample.yaml`

**Expected behavior**

kubectl apply should return successfully and the policy should be applied.

daemon1024:
- unix file paths accepts all characters, not just alphanumeric. Updated validation regex accordingly
- matchDir now accepts "/" as well

FIxes #641

### DIFF
--- a/deployments/CRD/KubeArmorHostPolicy.yaml
+++ b/deployments/CRD/KubeArmorHostPolicy.yaml
@@ -71,7 +71,7 @@ spec:
                           items:
                             properties:
                               path:
-                                pattern: ^\/([A-z0-9-_.]+\/)*([A-z0-9-_.]+)$
+                                pattern: ^\/+.*[^\/]$
                                 type: string
                             type: object
                           type: array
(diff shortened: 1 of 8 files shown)


nyrahul (approved):
LGTM. Thanks

PR #656, opened 2022-03-17
fix for disabling HostLogs in k8s env

nyrahul:
* also a quick fix for not keeping cache for apk reducing the image size


Ankurk99 (approved):

nam-jaehyun (approved):
Good to merge it.

PR #658, opened 2022-03-21
Remove backed-up host policy on policy removal

seswarrajan:

### DIFF
--- a/KubeArmor/core/kubeUpdate.go
+++ b/KubeArmor/core/kubeUpdate.go
@@ -1586,8 +1586,12 @@ func (dm *KubeArmorDaemon) ParseAndUpdateHostSecurityPolicy(event tp.K8sKubeArmo
 	dm.UpdateHostSecurityPolicies()
 
 	if !cfg.GlobalCfg.K8sEnv && (cfg.GlobalCfg.KVMAgent || cfg.GlobalCfg.HostPolicy) {
-		// backup HostSecurityPolicy to file
-		dm.backupKubeArmorHostPolicy(secPolicy)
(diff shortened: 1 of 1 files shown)


KubeArmor/core/kubeUpdate.go
```diff
@@ -1629,6 +1633,21 @@ func (dm *KubeArmorDaemon) WatchHostSecurityPolicies() {
 // == HostPolicy Backup & Restore == //
 // ================================= //
 
+// removeBackUpPolicy Function
+func (dm *KubeArmorDaemon) removeBackUpPolicy(name string) {
+
+	fname := cfg.PolicyDir + name + ".yaml"
+	// Check for "/opt/kubearmor/policies" path. If dir not found, create the same
+	if _, err := os.Stat(fname); err != nil {
+		kg.Warnf("Backup policy [%v] not exist", fname)
```
nyrahul: This shoud not be a Warnf since if there is no policy applied before then this Warning will always be printed. It is better to have this as info.

KubeArmor/core/kubeUpdate.go
```diff
@@ -1629,6 +1633,21 @@ func (dm *KubeArmorDaemon) WatchHostSecurityPolicies() {
...
+		return
+	}
+
+	if err := os.Remove(fname); err != nil {
+		kg.Err(err.Error())
```
nyrahul:
```suggestion
		kg.Errf("unable to delete file:%s err=%s", fname, err.Error())
```

nyrahul (changes requested):
Mostly looks good to me. Couple of trivial comments.

nyrahul (approved):
LGTM. Please squash the commits.

PR #665, opened 2022-04-05
apparmor-enforcer: skip non-regular profiles

nyrahul:
SUSE creates few symbolic links (`cache`, `cache.d` in the `/etc/apparmor.d/` folder.
Kubearmor gives out error since the current check if based on IsDir()
and the sym links are not covered in that check. Now we are checking if
the file is a regular file and only use it in that case.


nam-jaehyun (approved):
Let's merge it.

PR #668, opened 2022-04-11
Update Policy Matcher

nam-jaehyun:
- update policy matcher
- upgrade protobuf version
- update test scenarios and policies


nam-jaehyun:
Once this PR is merged, I'll update the PR for the SELinux enforcer on top of this.

KubeArmor/core/kubeArmor.go
```diff
@@ -338,50 +338,53 @@ func KubeArmor() {
...
-		// == //
+			if dm.Node.NodeIP == "" {
+				for timeout := 0; timeout <= 60; timeout++ {
+					if dm.Node.NodeIP != "" {
+						break
+					}
 
-		// wait for a while
-		time.Sleep(time.Second * 1)
+					if dm.Node.NodeIP == "" && timeout == 60 {
```
nyrahul: We can push this condition outside of the for loop and just check for single condition `dm.Node.NodeIP == ""`.

KubeArmor/core/kubeArmor.go
```diff
@@ -338,50 +338,53 @@ func KubeArmor() {
 		dm.Node.KernelVersion = kl.GetCommandOutputWithoutErr("uname", []string{"-r"})
 		dm.Node.KernelVersion = strings.TrimSuffix(dm.Node.KernelVersion, "\n")
 
-		dm.Node.PolicyEnabled = tp.KubeArmorPolicyEnabled
+		kg.Print("Updated the node information")
 
-		cfg.GlobalCfg.Policy = false
-		cfg.GlobalCfg.HostPolicy = true
+	} else if cfg.GlobalCfg.K8sEnv {
+		if K8s.InitK8sClient() {
```
nyrahul: We can reduce nesting if the other condition is handled ... 
```suggestion
		if !K8s.InitK8sClient() {
		       print error
		       return
		}
		// ... normal processing
		
```

nyrahul (changes requested):

nam-jaehyun:
Updated the code according to @nyrahul 's feedback.

PR #669, opened 2022-04-11
Update SELinux Enforcer

nam-jaehyun:
- update SELinux enforcer
- update test scenarios and policies for hosts


nam-jaehyun:
there are duplicated fixes addressed in #668 and I added them because the CI test would fail due to the policy matcher bugs. Once #668 is merged, I'll remove the duplicated fixes from this PR.

PR #655, opened 2022-03-16
mutation webhook

achrefbensaad:
Work related to this [issue](https://github.com/kubearmor/KubeArmor/issues/360)

PR #671, opened 2022-04-12
add controller deployment

achrefbensaad:
Work needed so we can deploy the annotations controller https://github.com/kubearmor/KubeArmor/pull/655 via karmor.

### DIFF
--- a/deployments/get/defaults.go
+++ b/deployments/get/defaults.go
@@ -3,7 +3,10 @@
 
 package deployments
 
-import corev1 "k8s.io/api/core/v1"
+import (
+	admissionregistration "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+)
 
 var kubearmor = "kubearmor"
 var port int32 = 32767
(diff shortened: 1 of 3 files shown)


nam-jaehyun:
in the case of the admission controller, let's not use 'deploygen' because it still needs to work on non-ubuntu-based clusters. (in non-ubuntu-based clusters, the admission controller does nothing; thus, there is no reason that we need to deploy it in such clusters)

Also, we have the corresponding deployments in [deployments/annotations](https://github.com/kubearmor/KubeArmor/tree/main/deployments/annotations).
By using those YAML files, let's update karmor to deploy the admission controller during deploying kubearmor
if all nodes are ubuntu (or debian).

nam-jaehyun:
@achrefbensaad 

double-check: duplicated functionality?
we create kubearmor-annotation-manager.yaml by using Makefile in pkg/KubeArmorAnnotations (#695).
do we still need to generate YAML files using deploygen? I think we don't need to update deploygen (this PR).
what do you think?

if possible, can you raise an issue about the admission controller deployment during the kubearmor deployment if all nodes are ubuntu (or debian) at kubearmor/kubearmor-client?

nam-jaehyun:
@achrefbensaad double-check: duplicated functionality?
we create kubearmor-annotation-manager.yaml by using Makefile in pkg/KubeArmorAnnotations (#695).
do we still need to generate YAML files using deploygen? I think we don't need to update deploygen. what do you think?

achrefbensaad:
@nam-jaehyun IMO, we still need to update deploygen so that  we can install the annotations manager via `karmor install`.


> if possible, can you raise an issue about the admission controller deployment during the kubearmor deployment if all nodes are ubuntu (or debian) at kubearmor/kubearmor-client?

can you please further expalin?

nam-jaehyun:
karmor does not currently check if all nodes in a cluster are based on ubuntu.
under the current implementation, if we add the annotation controller part in deploygen, karmor will deploy cert-manager and annotation controller even though the cluster nodes are based on centos or non-ubuntu-based OSes.

If I'm a user who uses non-ubuntu-based clusters, I would not understand why I need to deploy cert-manager (including multiple k8s resources such as services, pods, roles, and a service account) and the admission controller (including multiple k8s resources such as pods, services, configmap, and a service account) which have no meaning in the clusters while consuming my computing resources.

thus, I suggest that you update the code of karmor ('install' part).
in your update, karmor first checks if all nodes are based on ubuntu (we can check it by looking at k8s nodes).
then, if all nodes are based on ubuntu, karmor deploys cert-manager and admission controller by reading the YAML files in deployments/annotations and applying them into k8s.

nam-jaehyun:
by the way, is this PR fully implemented?
when I compiled deploygen, it doesn't create anything related to the admission controller.

nam-jaehyun (changes requested):
If you still think that we need to update deploygen, I'm okay with that.
However, deploygen with your update doesn't still work. please check it again.
Also, please rebase the code.

achrefbensaad:
@nam-jaehyun, yes I am sure.

I just run this PR localy and it installed all needed resources.

image

This PR in it self does not install anything, it holds the definition objects that are refrenced by karmor to install all the needed resources.

To reproduce locally:
- Checkout this PR localy 
- Checkout karmor PR locally https://github.com/kubearmor/kubearmor-client/pull/65
- Add a `go.mod` replacement in karmor cli tool code to point to this PR/deployments
image
- Compile
- Run `./karmor install`

nam-jaehyun (approved):
LGTM

PR #672, opened 2022-04-14
use random grpc port

### LINKED ISSUE #623: using random kubearmor GRPC server port
## Feature Request

**Short Description**

Currently, the GRPC server port [[is a configuration](https://github.com/kubearmor/KubeArmor/blob/bf39613c0c01e986d3e77d2f46a59c526cbf5907/KubeArmor/config/config.go#L79)](https://github.com/kubearmor/KubeArmor/blob/bf39613c0c01e986d3e77d2f46a59c526cbf5907/KubeArmor/config/config.go#L79) and the default value is 32767.
This works fine on kubearmor in k8s mode where kubearmor is operating as a pod and the listen ports are deterministic within the pod.

However, in case of [systemd/host-process mode of operation](https://github.com/kubearmor/KubeArmor/blob/main/getting-started/kubearmor_vm.md), it is possible that port 32767 might have been in use by some other process on the host and if it is in use, the kubearmor startup will fail.

Requirement of this task:
- [ ] Ability to specify 0 as the GRPC port configuration in which case the bind should happen on the random port.
- [ ] This port information should be kept in a pid file.
- [ ] `karmor` [cli tool](https://github.com/kubearmor/kubearmor-client) will check for this pid file in case of k8s=false env and pick the port information from there to connect.

It is possible that in systemd mode of invocation the port 32767 on the host is already in use. In such cases, kubearmor fails to start because of "address already in use" error.

1. Ability to specify 0 as the GRPC port configuration in which case the bind should happen on the random port.
2. This port information should be kept in a pid file.
3. `karmor` [cli tool](https://github.com/kubearmor/kubearmor-client) will check for this pid file in case of k8s=false env and pick the port information from there to connect.

For pid file, check what is the appropriate location to keep the pid file.

nthnieljson:
KubeArmor/config/config.go : define a constant for the pid file path
KubeArmor/feeder/feeder.go : use a random grpc port number and save it to a file

previously we can't use random grpc port

now we can use random grpc port number by specifying "0" as the port number, by using port number 0, we will listen on random port and then we save the random port number to a pid file located in /opt/kubearmor

Signed-off-by: nthnieljson <nthnieljson@gmail.com>

### DIFF
--- a/KubeArmor/config/config.go
+++ b/KubeArmor/config/config.go
@@ -35,6 +35,9 @@ type KubearmorConfig struct {
 // PolicyDir policy dir path for host policies backup
 const PolicyDir string = "/opt/kubearmor/policies/"
 
+// PIDFilePath for pid file path
+const PIDFilePath string = "/opt/kubearmor/kubearmor.pid"
+
 // GlobalCfg Global configuration for Kubearmor
 var GlobalCfg KubearmorConfig
 
--- a/KubeArmor/core/kubeArmor.go
+++ b/KubeArmor/core/kubeArmor.go
@@ -177,7 +177,11 @@ func (dm *KubeArmorDaemon) DestroyKubeArmorDaemon() {
 	// delete pid file
 	if _, err := os.Stat(cfg.PIDFilePath); err == nil {
 		kg.Print("Deleting PID file")
-		os.Remove(cfg.PIDFilePath)
+		
+		err := os.Remove(cfg.PIDFilePath)
+		if err != nil {
+			kg.Errf("Failed to delete PID file")
+		}
 	}
 }
 
--- a/KubeArmor/core/kubeArmor.go
+++ b/KubeArmor/core/kubeArmor.go
@@ -173,6 +173,12 @@ func (dm *KubeArmorDaemon) DestroyKubeArmorDaemon() {
 	// wait for other routines
 	kg.Print("Waiting for routine terminations")
 	dm.WgDaemon.Wait()
+
+	// delete pid file
+	if _, err := os.Stat(cfg.PIDFilePath); err == nil {
+		kg.Print("Deleting PID file")
+		os.Remove(cfg.PIDFilePath)
+	}
 }
 
 // ============ //
--- a/KubeArmor/feeder/feeder.go
+++ b/KubeArmor/feeder/feeder.go
@@ -329,7 +329,9 @@ func NewFeeder(node *tp.Node) *Feeder {
 			return nil
 		}
 
-		defer pidFile.Close()
+		defer func() {
+			pidFile.Close()
+		}()
 
 		port := fmt.Sprintf("%d", listener.Addr().(*net.TCPAddr).Port)
 		fd.Port = fmt.Sprintf(":%s", port)
--- a/KubeArmor/feeder/feeder.go
+++ b/KubeArmor/feeder/feeder.go
@@ -329,6 +329,8 @@ func NewFeeder(node *tp.Node) *Feeder {
 			return nil
 		}
 
+		defer pidFile.Close()
+
 		port := fmt.Sprintf("%d", listener.Addr().(*net.TCPAddr).Port)
 		fd.Port = fmt.Sprintf(":%s", port)
 
--- a/KubeArmor/feeder/feeder.go
+++ b/KubeArmor/feeder/feeder.go
@@ -322,6 +322,23 @@ func NewFeeder(node *tp.Node) *Feeder {
 	}
 	fd.Listener = listener
 
+	if cfg.GlobalCfg.GRPC == "0" {
+		pidFile, err := os.Create(cfg.PIDFilePath)
+		if err != nil {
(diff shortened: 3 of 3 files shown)


nthnieljson:
this is addressing the issue on #623 to use random grpc port number

KubeArmor/feeder/feeder.go
```diff
@@ -322,6 +322,23 @@ func NewFeeder(node *tp.Node) *Feeder {
 	}
 	fd.Listener = listener
 
+	if cfg.GlobalCfg.GRPC == "0" {
```
daemon1024, line 325: Won't this logic be needed before we create the listener?

nthnieljson: So this part is actually is used to store the port number in the pid file. The port number itself can be retrieved after we create the listener. So if we decide to listen to port 0 on the listener, net.Listen("tcp", "0"), it will actually automatically find the random available port. If this part is implemented before the listener creation, we can't get the port number

nthnieljson: After listening to port 0, net.Listen("tcp","0"), we then can get the random port number that the listener found by using `listener.Addr().(*net.TCPAddr).Port `

seswarrajan (approved):
LGTM. @nyrahul need to approve.

KubeArmor/feeder/feeder.go
```diff
@@ -322,6 +322,23 @@ func NewFeeder(node *tp.Node) *Feeder {
...
+		pidFile, err := os.Create(cfg.PIDFilePath)
+		if err != nil {
+			kg.Errf("Failed to create file %s", cfg.PIDFilePath)
+			return nil
+		}
+
```
daemon1024, line 331:
```suggestion
		pidFile, err := os.Create(cfg.PIDFilePath)
		if err != nil {
			kg.Errf("Failed to create file %s", cfg.PIDFilePath)
			return nil
		}
		
		defer pidFile.Close()
```
I believe we should close the file after writing to it.

nyrahul (approved):
Thanks @nthnieljson for handling this. LGTM.

So as i understand the logic is:
If GRPC config is set to 0, then we listen on port 0 which means we will get a dynamically allocated port, whose value will be written in the pid file.
There is [another PR](https://github.com/kubearmor/kubearmor-client/pull/62) in kubearmor-client which will check for this pid file and use the port number.

@nam-jaehyun , WDYT? I like this solution.

KubeArmor/feeder/feeder.go
```diff
@@ -322,6 +322,25 @@ func NewFeeder(node *tp.Node) *Feeder {
...
+		defer pidFile.Close()
```
daemon1024: It seems defer is not a preferred method to close the file 😅 We can perhaps use
```
defer func() {
  err := pidFile.Close()
  // ...
}()
```
Ref https://github.com/securego/gosec/issues/512#issuecomment-675286833

KubeArmor/config/config.go
```diff
@@ -35,6 +35,9 @@ type KubearmorConfig struct {
 // PolicyDir policy dir path for host policies backup
 const PolicyDir string = "/opt/kubearmor/policies/"
 
+// PIDFilePath for pid file path
+const PIDFilePath string = "/opt/kubearmor/kubearmor.pid"
```
nam-jaehyun, line 39: if we use this PID file, when do we remove this file?

KubeArmor/feeder/feeder.go
nam-jaehyun: if we use this kind of defer, I think it fails during go-sec checks. does it pass go-sec?

nthnieljson: I just applied @daemon1024 suggestion, will resolve this conversation to see if it pass go-sec

KubeArmor/config/config.go
nthnieljson: I just add a mechanism to delete the file on `DestroyKubeArmorDaemon()` function

KubeArmor/feeder/feeder.go
daemon1024: We need to error check inside the defer func.

KubeArmor/core/kubeArmor.go
```diff
@@ -173,6 +173,12 @@ func (dm *KubeArmorDaemon) DestroyKubeArmorDaemon() {
 	// wait for other routines
 	kg.Print("Waiting for routine terminations")
 	dm.WgDaemon.Wait()
+
+	// delete pid file
+	if _, err := os.Stat(cfg.PIDFilePath); err == nil {
+		kg.Print("Deleting PID file")
+		os.Remove(cfg.PIDFilePath)
```
daemon1024: need to handle errors

daemon1024 (changes requested):

daemon1024 (approved):

nam-jaehyun:
@nthnieljson can you make all the changes in a single commit in order to merge this PR?

nam-jaehyun (approved):
after squashing the commits, let's merge the PR.

PR #687, opened 2022-05-04
add mutation webhook controller

achrefbensaad:
Remove annotations logic from kubearmor
update ci.
Signed-off-by: Achref Ben Saadd <achref9612@gmail.com>

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -55,6 +55,17 @@ jobs:
       - name: Setup Enviroment
         run: |
           ./contribution/k3s/install_k3s.sh
+          
+      - name: Install annotation controller
+        run: |
+          pushd $(pwd)
+          cd pkg/KubeArmorAnnotation
+          REGIP=$(ip -o route get to 8.8.8.8 | sed -n 's/.*src \([0-9.]\+\).*/\1/p')
+          make docker-build  docker-push IMG=$REGIP:5000/webhook
+          make deploy IMG=$REGIP:5000/webhook
+          sleep 10
+          kubectl get po -n kube-system
+          popd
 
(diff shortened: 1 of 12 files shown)


achrefbensaad:
need to be rebased before merge

achrefbensaad:
@nyrahul @nam-jaehyun @daemon1024 Please review this

.github/workflows/ci-test.yml
```diff
@@ -56,6 +56,34 @@ jobs:
         run: |
           ./contribution/k3s/install_k3s.sh
 
+      - name: Setup self-hosted docker registry
+        run: |
```
daemon1024: Can we move these to scripts instead? We would be needing them local dev environment too right? Like how we are calling k3s installation script.

achrefbensaad: Yes, its a good idea. I have implimented the required changes. Please take an other look.

daemon1024 (approved):

nam-jaehyun:
1. I'd like to still keep the code in kubeUpdate.go since the admission controller only works in the cluster where all nodes are based on ubuntu. Also, we need to test KubeArmor in a local env (e.g., make run), but the admission controller might be not in such an env.
2. I don't get why we need to create a local registry, which increases the overall test complexity.

nam-jaehyun (changes requested):
Multiple files are specifically updated for the admission controller, but the updated files are also used for some other purposes. We may need to think of this PR again.

nam-jaehyun:
refer to https://github.com/kubearmor/KubeArmor/pull/690

nam-jaehyun:
outdated PR. we have a separate PR for the CI update (#695).

PR #691, opened 2022-05-12
create stable release based on v0.3

Ankurk99:


nyrahul:
Hey @Ankurk99 , does the STABLE-RELEASE value matters in the branch other than main branch? [ref](https://github.com/kubearmor/KubeArmor/blob/7410c1c91ead25bb28e15ed628ceb908d30ca897/.github/workflows/stable-version-release.yml#L19)

Ankurk99:
Yes you are right! then it seems the problem is different, 'cause somehow the github flow is not running due to pattern matching [here](https://github.com/kubearmor/KubeArmor/runs/5748457622?check_suite_focus=true). 
I am taking a look at it.

Ankurk99:
The github flow is not running on kubearmor (https://github.com/kubearmor/KubeArmor/runs/6409876499?check_suite_focus=true) however it's running fine on fork (https://github.com/Ankurk99/KubeArmor/runs/6409821073?check_suite_focus=true#step:3:11)

PR #696, opened 2022-05-16
Fix for identical source/resource value for process operation on system log

### LINKED ISSUE #692: Source and Resource is same for all process based logs
**General Information**

- Environment : GKE, Minikube, k3s and VM
- Kernel version 5.13.0-40-generic

**To Reproduce**
1. Run Kubearmor in Systemd mode/k8s mode
2. Launch karmor in another tab and listen to all logs (karmor log --logFilter all)
3. Observe logs for all **Operation: Process**. 

**Expected behavior**

For process operation, the source and resource needs to be same.

**Screenshots**

The source and resource for all process operations are same/identical.

seswarrajan:

### DIFF
--- a/KubeArmor/monitor/processTree.go
+++ b/KubeArmor/monitor/processTree.go
@@ -77,7 +77,7 @@ func (mon *SystemMonitor) BuildPidNode(containerID string, ctx SyscallContext, e
 	node.ParentExecPath = mon.GetExecPath(containerID, ctx.HostPPID)
 	node.ExecPath = execPath
 
-	node.Source = execPath
+	node.Source = node.ParentExecPath
 	node.Args = ""
 
 	for idx, arg := range args {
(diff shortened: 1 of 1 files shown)


nam-jaehyun (approved):
LGTM

KubeArmor/monitor/processTree.go
```diff
@@ -173,6 +173,23 @@ func (mon *SystemMonitor) GetCommand(containerID string, hostPid uint32) string
...
+}
+
+// GetCommandWithArgs Function
+func (mon *SystemMonitor) GetCommandWithArgs(containerID string, hostPid uint32) string {
+	ActiveHostPidMap := *(mon.ActiveHostPidMap)
+	ActivePidMapLock := *(mon.ActivePidMapLock)
+
+	ActivePidMapLock.Lock()
+	defer ActivePidMapLock.Unlock()
+
```
nyrahul, line 192: I didn't understand this change!! What does this achieve?
New function GetCommandWithArgs() is introduced but not used!

nyrahul (commented):
Please check my comment inline.

KubeArmor/monitor/processTree.go
seswarrajan: Yes @nyrahul 
From the existing function GetCommand(), we removed adding arguments to source.
If anyone requires to use args to source, then this function is introduced (GetCommandWithArgs).

It is currently unused. It is just kept for future use if needed.

nyrahul: I see. Thanks

nyrahul (approved):
LGTM

PR #697, opened 2022-05-16
feat: support CRI-O

### LINKED ISSUE #666: support for CRIO container runtime
## Feature Request

Kubearmor currently supports docker, and containerd container runtimes for k8s.
There are several k8s engine (for e.g. [OpenShift](https://docs.openshift.com/container-platform/3.11/crio/crio_runtime.html)) that use [CRIO](https://cri-o.io/).

Kubearmor connects to container-runtime to get the mount-namespace and other details (for e.g., container image details etc). These details are used subsequently in the telemetry/log enrichment. For e.g.,
1. to check if the process event is in context to the given container/namespace.
2. to check what is the container image id for the event that is generated

The aim here is to look into containerd implementation in kubearmor and map it to CRIO as well.

Tasks:
- [ ] design points
- [ ] validate policy enforcement with CRIO
- [ ] validate alerts/telemetry with CRIO
- [ ] testing with local k8s env with CRIO

DelusionalOptimist:
Signed-off-by: Rudraksh Pareek <rudraksh@accuknox.com>

Fixes #666 
* This PR makes KubeArmor compatible with CRI-O container runtime.
* The approach used is similar to that of containerd with some changes here and there.

Tasks:
- [x] Implement monitoring 
- [x] Tests
- [x] Setup scripts


DelusionalOptimist:
@nam-jaehyun @nyrahul @daemon1024 Please give this a review. (I'm unable to request reviews :sweat_smile: ).
I've temporarily modified the gh workflow so that k3s runs with crio and tests can happen. Github tests 6, 8 & 13 are failing and I'll take a look into them in the meantime.
Also, please terminate https://github.com/kubearmor/KubeArmor/actions/runs/2360471345

DelusionalOptimist:
Github test 06 and 08 passed eventually.
However, github test 13 didn't pass because crio by default doesn't permit the capability `CAP_NET_RAW` and it has to be first enabled by editing default capabilities in `/etc/crio/crio.conf`. What should we do here? :thinking: 

Also, in local tests, multiubuntu test 02 and 07 are failing. 07 fails due to the same reason as above but I'm not sure about 02 as audit action works fine when tested manually. Here are logs from karmor when the test conditions are created  -
```
== Alert / 2022-05-22 15:42:57.985910 ==
Cluster Name: default
Host Name: kubearmor-dev
Namespace Name: multiubuntu
Pod Name: ubuntu-4-deployment-566bf47cd7-25dx6
Container ID: 72dbd9ab9ec9e2b997ac809087e228479f6f8e8df035b52e1ff251e8f5bb414f
Container Name: ubuntu-4-container
Policy Name: ksp-group-2-proc-path-audit
Severity: 4
Type: MatchedPolicy
Source: /bin/bash
Operation: Process
Resource: /bin/sleep 1
Data: syscall=SYS_EXECVE
Action: Audit
Result: Passed
```

For ref: the default capabilties applied by CRI-O:
```
default_capabilities = [
        "CHOWN",
        "DAC_OVERRIDE",
        "FSETID",
        "FOWNER",
        "SETGID",
        "SETUID",
        "SETPCAP",
        "NET_BIND_SERVICE",
        "KILL",
]
```

DelusionalOptimist:
@nam-jaehyun @nyrahul as discussed, I've updated the github workflows to test for all the runtimes and skip the tests that have been specifically failing only for crio.

nyrahul:
> @nam-jaehyun @nyrahul as discussed, I've updated the github workflows to test for all the runtimes and skip the tests that have been specifically failing only for crio.

Hey, I see that tests are still failing! Also can you please squash the commits?

DelusionalOptimist:
> Hey, I see that tests are still failing! Also can you please squash the commits?

Oof yes. Taking a look into why the tests with containerd are failing :eyes:

contribution/k3s/install_k3s.sh
```diff
@@ -12,9 +12,12 @@ if [ -x "$(command -v docker)" ]; then # docker
         curl -sfL https://get.k3s.io | K3S_KUBECONFIG_MODE="644" INSTALL_K3S_EXEC="--disable=traefik --docker" sh -
         [[ $? != 0 ]] && echo "Failed to install k3s" && exit 1
     fi
-else # containerd
+elif [ "$RUNTIME" == "containerd" ]; then # containerd
     curl -sfL https://get.k3s.io | K3S_KUBECONFIG_MODE="644" INSTALL_K3S_EXEC="--disable=traefik" sh -
     [[ $? != 0 ]] && echo "Failed to install k3s" && exit 1
+else #cri-o
```
nam-jaehyun: let's use containerd by default (not cri-o)

.github/workflows/ci-test.yml
```diff
@@ -18,19 +18,19 @@ on:
 
 jobs:
   build:
-    name: Auto-testing Framework / ${{ matrix.os }}
+    name: Auto-testing Framework / ${{ matrix.os }} / ${{ matrix.runtime }}
```
nam-jaehyun: let's not add cri-o-based test in the CI (no modification in ci-test.yml)

contribution/k3s/install_k3s.sh
```diff
@@ -3,7 +3,7 @@
 # Copyright 2021 Authors of KubeArmor
 
 # create a single-node K3s cluster
-if [ -x "$(command -v docker)" ]; then # docker
+if [ "$RUNTIME" == "docker" ]; then # docker
```
nam-jaehyun, line 6: who defines $RUNTIME? is it possible to detect cri-o automatically?

contribution/self-managed-k8s/crio/install-crio.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
nam-jaehyun, line 1: wondering if it is possible to use cri-o in self-managed-k8s-selinux?

tests/test-scenarios-github.sh
```diff
@@ -22,6 +22,13 @@ realpath() {
 TEST_HOME=`dirname $(realpath "$0")`
 CRD_HOME=`dirname $(realpath "$0")`/../deployments/CRD
 ARMOR_HOME=`dirname $(realpath "$0")`/../KubeArmor
+IGN_FILE=$TEST_HOME/tests.ignore
```
nam-jaehyun: we don't need the modification of test-scenarios-github.sh because we're not using cri-o in CI.
by the way, did you test ./test-scenarios-local.sh or ./test-scenarios-in-runtime.sh with multiubuntu_test_XX? if some test scenarios have CAP_NET_RAW, I think it is okay to add such scenarios in tests.ignore.

contribution/self-managed-k8s/k8s/initialize_kubernetes.sh
```diff
@@ -7,9 +7,20 @@ if [ "$CNI" == "" ]; then
...
 fi
 
+# use docker as default CRI
+if [ "$CRI_SOCKET" == "" ]; then
+	if [ -f /var/run/docker.sock ]; then
+		CRI_SOCKET=unix:///var/run/docker.sock
+	elif [ -f /var/run/containerd/containerd.sock ]; then
+		CRI_SOCKET=unix:///var/run/containerd/containerd.sock
+	elif [ -f /var/run/crio/crio.sock ]; then
+		CRI_SOCKET=unix:///var/run/crio/crio.sock
```
nam-jaehyun, line 17: what if there is no matched CRI socket?

contribution/self-managed-k8s/crio/uninstall-crio.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
nam-jaehyun, line 1: wondering if it is possible to use cri-o in self-managed-k8s-selinux?

KubeArmor/core/kubeArmor.go
```diff
@@ -475,7 +475,7 @@ func KubeArmor() {
 					return
 				}
 			}
-		} else { // containerd
+		} else if strings.HasPrefix(dm.Node.ContainerRuntimeVersion, "containerd") { // containerd
```
nam-jaehyun: let's use containerd by default

nam-jaehyun (commented):
can you also update contribution/vagrant/Vagrantfile?

cri-o needs to work under the following commands.
KubeArmor/KubeArmor$ make vagrant-up RUNTIME=crio
KubeArmor/KubeArmor$ make vagrant-up RUNTIME=crio NETNEXT=1
if possible, KubeArmor/KubeArmor$ make vagrant-up RUNTIME=crio OS=centos

.github/workflows/ci-test.yml
DelusionalOptimist: Agreed @nam-jaehyun. I added them so that we could run tests on GH with cri-o for the changes in this PR... will undo.

contribution/k3s/install_k3s.sh
DelusionalOptimist: @nam-jaehyun right now RUNTIME would be defined by user while executing the script.
To detect automatically, we may check for existence of socket files or run commands to check for a runtime as we were doing for docker.
However I thought it might be not so flexible. Many systems like GitHub actions come with docker pre-installed and running so in that case this script would always use docker.

contribution/self-managed-k8s/crio/install-crio.sh
DelusionalOptimist: Haven't taken a look yet but will give it a try and update the script accordingly. Thanks!

tests/test-scenarios-github.sh
DelusionalOptimist: Sure.
Yes, I did try both these tests and the ones with CAP_NET_RAW weren't successful because of https://github.com/kubearmor/KubeArmor/pull/697#issuecomment-1133923670.

DelusionalOptimist: I thought of adding these to tests.ignore but that would have ignored it for all container runtimes even the ones which are working fine right now, right? :thinking:

contribution/self-managed-k8s/k8s/initialize_kubernetes.sh
DelusionalOptimist: In that case CRI_SOCKET remains empty and if there is only one container runtime, kubeadm uses it. If there are multiple, kubeadm fails with:
```
Found multiple CRI sockets, please use --cri-socket to select one: /run/containerd/containerd.sock, /var/run/crio/crio.sock
```

DelusionalOptimist:
@nam-jaehyun @nyrahul 

I and @daemon1024 were thinking about extending the use of cri-api and using it for docker and containerd as well. The current code for handling the two runtimes would be deprecated and the code used here would be used to communicate with the runtimes' socket.
It is possible as kubernetes mandates a cri friendly interface for every runtime. Containerd implements the `cri` plugin which it needs to communicate with kubelet and docker has dockershim (or cri-dockerd now). So since both of these are compatible with cri we can save the effort of maintaining code for their respective APIs.

Does this sound good?

nyrahul:
> @nam-jaehyun @nyrahul
> 
> I and @daemon1024 were thinking about extending the use of cri-api and using it for docker and containerd as well. The current code for handling the two runtimes would be deprecated and the code used here would be used to communicate with the runtimes' socket. It is possible as kubernetes mandates a cri friendly interface for every runtime. Containerd implements the `cri` plugin which it needs to communicate with kubelet and docker has dockershim (or cri-dockerd now). So since both of these are compatible with cri we can save the effort of maintaining code for their respective APIs.
> 
> Does this sound good?

Imo, it sounds very good. But my understanding was that we are using some containerd or docker specific apis which are not part of cri-api spec. If this understanding is wrong, what you saying is the best way to handle things i.e. just use cri-api for everything ... This will simplify lot of our program logic. But I am keen on what @nam-jaehyun has to say.

daemon1024 (approved):
LGTM

contribution/self-managed-k8s-selinux/crio/uninstall_crio.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2021 Authors of KubeArmor
+
+sudo systemctl stop crio.service
+
+sudo dnf remove cri-o
```
nyrahul: To install `yum` was used. Why `dnf` to remove? Why not `yum remove`? I understand dnf is almost same as yum, but is it necessary that dnf always exists where yum exists?

KubeArmor/core/crioHandler.go
```diff
@@ -0,0 +1,391 @@
...
+}
+
+// NewCrioHandler Function creates a new Crio handler
+func NewCrioHandler() *CrioHandler {
+	ch := &CrioHandler{}
+
+	sockFile := "unix://"
+
+	// TODO: identify other k8s distros which support CRIO by default
+	for idx, candidate := range []string{"/var/run/crio/crio.sock"} {
```
nyrahul: There is an env var CRI_SOCKET that is set in the scripts. Lets check the presence of that env var and add it to this string list (as the first element).

KubeArmor/core/crioHandler.go
```diff
@@ -0,0 +1,391 @@
...
+		if _, err := os.Stat(filepath.Clean(candidate)); err == nil {
+			sockFile = sockFile + candidate
+
+			if idx == 0 {
+				// default crio
+				ch.StoragePath = "/run/crio"
```
nyrahul: This var `StoragePath` seems to be unused. Can this be removed?

KubeArmor/core/crioHandler.go
```diff
@@ -0,0 +1,391 @@
...
+	// path to the rootfs
+	container.MergedDir = containerInfo.RuntimeSpec.Root.Path
+
+	pid := strconv.Itoa(containerInfo.Pid)
+
+	if data, err := os.Readlink("/proc/" + pid + "/ns/pid"); err == nil {
+		if _, err := fmt.Sscanf(data, "pid:[%d]\n", &container.PidNS); err != nil {
+			kg.Warnf("Unable to get PidNS (%s, %s, %s)", containerID, pid, err.Error())
+		}
+	}
```
nyrahul, line 134: Can we print an error in the else condition? If the else condition is hit, we know that there will be problems.

KubeArmor/core/crioHandler.go
```diff
@@ -0,0 +1,391 @@
...
+
+	if data, err := os.Readlink("/proc/" + pid + "/ns/mnt"); err == nil {
+		if _, err := fmt.Sscanf(data, "mnt:[%d]\n", &container.MntNS); err != nil {
+			kg.Warnf("Unable to get MntNS (%s, %s, %s)", containerID, pid, err.Error())
+		}
+	}
```
nyrahul, line 142: Print error in else condition.

KubeArmor/core/kubeArmor.go
```diff
@@ -485,6 +485,27 @@ func KubeArmor() {
 					return
 				}
 			}
+		} else if strings.HasPrefix(dm.Node.ContainerRuntimeVersion, "cri-o") { // cri-o
+			sockFile := false
+
+			for _, candidate := range []string{"/var/run/crio/crio.sock"} {
+				if _, err := os.Stat(candidate); err == nil {
```
nyrahul: The same `for` loop is used in `KubeArmor/core/crioHandler.go` as well .. It would be better to create a function for this for two reasons:
1. reduced code
2. it is better to check for env var CRI_SOCKET and use that as the preferred one (first element in the string array) if it exists.

nyrahul (changes requested):
Great PR!! Please check my comments inline.

contribution/self-managed-k8s-selinux/crio/uninstall_crio.sh
DelusionalOptimist: True. Will use `yum` for both :+1:

KubeArmor/core/crioHandler.go
DelusionalOptimist: @nyrahul the `CRI_SOCKET` variable is local to the script for initializing self managed kubernetes. It is used with the `--cri-socket` flag in `kubeadm init` command.

DelusionalOptimist: But yes, we should be able to set a runtime socket while starting kubearmor. Should we take a `--cri-socket` flag approach as taken by `kubeadm`, `crictl` and others? We can prolly have it as an option in [KubeArmorConfig](https://github.com/kubearmor/KubeArmor/blob/7e1810d2ad418655fb0d90760d57c29f8b550422/KubeArmor/config/config.go#L18)?

KubeArmor/core/kubeArmor.go
DelusionalOptimist: Will refactor :+1:

KubeArmor/core/crioHandler.go
nyrahul: That makes sense. Point is, there should be some way to change the sock path if the location of the path varies for different deployments.

KubeArmor/core/crioHandler.go
```diff
@@ -48,36 +44,11 @@ type CrioContainerInfo struct {
 // Crio Handler
 var Crio *CrioHandler
 
-// init Function
-func init() {
-	Crio = NewCrioHandler()
```
DelusionalOptimist: Had to move this out of the `init` function and into `MonitorCrioEvents` because init function is called before config is read which leads to `CRI_SOCKET` and `criSocket` flag not having any effect.

nyrahul:
@DelusionalOptimist , can you squash the commits?

DelusionalOptimist:
@nyrahul done :+1:

KubeArmor/core/crioHandler.go
daemon1024: Can we now try to auto detect CRI Socket in kArmor and pass it on when deploying?

DelusionalOptimist: @daemon1024 I wonder how karmor will get CRI socket for remote nodes :thinking:

daemon1024: In kubeadm atleast we get,
`kubeadm.alpha.kubernetes.io/cri-socket: /var/run/dockershim.sock`
In annotations of the node.

Is that the case in other k8s variants?

DelusionalOptimist: Yep, this goes for kubeadm. Though I haven't checked with others but k3s doesn't have this or any other way to get the socket.

nyrahul: I think we have an action item in kubearmor-client for handling this. Resolving this conversation.

nyrahul (approved):
LGTM 👍🏼

PR #699, opened 2022-05-18
Enable HostVisibility alone and not pod visibility on k8s=false mode

seswarrajan:


nam-jaehyun:
what is this PR for?

seswarrajan:
The changes are valid now.

PR #702, opened 2022-05-19
update helm

nam-jaehyun:
Updated helm files as they are outdated.


nyrahul (approved):
LGTM

PR #708, opened 2022-05-20
fix CleanUpExitedHostPids

nam-jaehyun:

### DIFF
--- a/.github/workflows/latest-release.yml
+++ b/.github/workflows/latest-release.yml
@@ -34,7 +34,6 @@ jobs:
           else
             echo ::set-output name=tag::${GITHUB_REF#refs/*/}
           fi
-
           ./contribution/k3s/install_k3s.sh
 
       - name: Generate KubeArmor artifacts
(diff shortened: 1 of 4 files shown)


KubeArmor/monitor/processTree.go
```diff
@@ -231,32 +231,40 @@ func (mon *SystemMonitor) DeleteActivePid(containerID string, ctx SyscallContext
...
+			if len(pidMap) == 0 {
+				delete(ActiveHostPidMap, containerID)
+			}
 		}
 
 		ActivePidMapLock.Unlock()
+
+		if mon.Status {
+			time.Sleep(10 * time.Second)
+		}
```
nyrahul:
```suggestion
		if !mon.Status {
		        break
		}
		time.Sleep(10 * time.Second)
```
quit the subroutine if destroy is called.

KubeArmor/monitor/systemMonitor.go
```diff
@@ -406,6 +403,9 @@ func (mon *SystemMonitor) InitBPF() error {
 
 // DestroySystemMonitor Function
 func (mon *SystemMonitor) DestroySystemMonitor() error {
+	mon.Status = false
+	time.Sleep(1 * time.Second)
```
nyrahul:
```suggestion
```
This sleep wont help since the sleep in SystemMonitor is of 10 sec. Better to remove the sleep completely.

nyrahul (changes requested):

nyrahul (commented):
The changes look good to me. Can you pls squash the commits into a single one and it should be good to merge? Thanks

PR #709, opened 2022-05-20
KubeArmor support for un-orchestrated containers

### LINKED ISSUE #693: supporting non-orchestrated containerized workloads
## Feature Request

**Short Description**

KubeArmor is now getting deployed on non-orchestrated containerized workloads. Kubearmor has requirement to operate in this mode for [LFEdge OpenHorizon](https://www.lfedge.org/projects/openhorizon/).

Kubearmor currently creates security profiles by:
* listening to events from k8s
* In systemd mode, it operates directly on the processes running on the host.

The new mode to be supported is to:
* listen to events from docker or containerd directly and apply policies by specifying apparmor policy directly to the docker daemon.

Containerized workloads are increasingly used for the edge scenarios. In most cases, these containers are not orchestrated but they need the same level of protection that is offered to orchestrated workloads by kubearmor.

Docker/containerd allows one to specify the apparmor/selinux policy directly as a command line option. Utilize this feature for kubearmor to specify the policy. When the user applies the policy, the kubearmor will create the apparmor policy in the context for the given container and then [ask docker](https://docs.docker.com/engine/security/apparmor/) to apply the policy.

Note that currently we do not have a way to specify a container within the kubearmor security policy. However, with #675 we had identified on how this can be specified. We can leverage the same policy spec in this case.

**Tasks**
- [x] kubearmor (operating in systemd mode) to auto identify if there is a docker daemon running
- [x] wait for events on docker/containerd
- [x] get docker containers telemetry events and stream it through kubearmor grpc
- [x] create apparmor policy at runtime based on ksp (kubearmor security policy)
- [x] specify apparmor policy to the docker daemon and restart the workload
- [x] ability to refresh the apparmor policy on policy updates
- [x] changes to ksp to identify container for applying the policy
- [ ] karmor cli changes to specify the policies and to get the policies.

**Tests**
- [ ] System tests to be added as part of CI
- [ ] Execute multiple containers and apply policies in the context of a single container
- [x] Ensure that policies are reapplied on container restart
- [ ] If the container is already started and the policy is applied, the container has to be restarted.

**Docs**
- [ ] "Deploying Kubearmor to protect containerized workloads"
- [ ] Blog explaining IBM OpenHorizon POC with edge based containerized workloads

Ankurk99:
Ref: #693


nam-jaehyun:
@Ankurk99 once this PR is updated, please let me know.

KubeArmor/policy/policy.go
```diff
@@ -34,3 +35,19 @@ func (p *ServiceServer) HostPolicy(c context.Context, data *pb.Policy) (*pb.Resp
 
 	return res, nil
 }
+
+func (p *ServiceServer) ContainerPolicy(c context.Context, data *pb.Policy) (*pb.Response, error) {
```
nyrahul: Please handle the new code lint issues.

KubeArmor/core/kubeUpdate.go
```diff
@@ -1168,6 +1168,460 @@ func (dm *KubeArmorDaemon) WatchSecurityPolicies() {
...
+			if len(path.Message) == 0 {
+				if len(secPolicy.Spec.Process.Message) > 0 {
+					secPolicy.Spec.Process.MatchPaths[idx].Message = secPolicy.Spec.Process.Message
+				} else {
+					secPolicy.Spec.Process.MatchPaths[idx].Message = secPolicy.Spec.Message
+				}
+			}
+
+			if len(path.Action) == 0 {
+				if len(secPolicy.Spec.Process.Action) > 0 {
```
nyrahul, line 1260: Is it possible to refactor this code?

nyrahul (approved):
Excellent PR. LGTM .. Few minor comments in the review.

KubeArmor/core/kubeArmor.go
```diff
@@ -578,13 +583,21 @@ func KubeArmor() {
...
 		reflection.Register(dm.Logger.LogServer)
 
+		dm.Logger.Print("Started to monitor host security policies on gRPC")
+
+	}
+
+	if !dm.K8sEnabled && cfg.GlobalCfg.ContainerPolicy {
+		policyService.UpdateContainerPolicy = dm.ParseAndUpdateContainerSecurityPolicy
+
 		dm.Logger.Print("Started to monitor host security policies on gRPC")
```
daemon1024:
```suggestion

		dm.Logger.Print("Started to monitor container security policies on gRPC")
```

KubeArmor/config/config.go
```diff
@@ -211,6 +217,7 @@ func LoadConfig() error {
 
 	GlobalCfg.Policy = viper.GetBool(ConfigKubearmorPolicy)
 	GlobalCfg.HostPolicy = viper.GetBool(ConfigKubearmorHostPolicy)
+	GlobalCfg.ContainerPolicy = viper.GetBool(ConfigKubearmorContainerPolicy)
```
daemon1024: Any particular reason for this flag? Since I see in the code we seem to be using `GlobalCfg.Policy`

KubeArmor/core/kubeArmor.go
```diff
@@ -449,6 +449,11 @@ func KubeArmor() {
 
 	// == //
 
+	if !dm.K8sEnabled && cfg.GlobalCfg.Policy {
```
daemon1024, line 457: Since we are using `cfg.GlobalCfg.Policy` I don't think there's any need for ContainerPolicy. 
No k8s and enabling KubeArmorPolicy should be suffice to indiciate it is for Containers right? WDYT?

KubeArmor/core/kubeUpdate.go
```diff
@@ -1168,6 +1168,460 @@ func (dm *KubeArmorDaemon) WatchSecurityPolicies() {
...
+			dm.Logger.UpdateSecurityPolicies("ADDED", newPoint)
+
+			if dm.RuntimeEnforcer != nil && newPoint.PolicyEnabled == tp.KubeArmorPolicyEnabled {
+				// enforce security policies
+				dm.RuntimeEnforcer.UpdateSecurityPolicies(newPoint)
+			}
+		}
+	} else {
+		dm.RuntimeEnforcer.UpdateAppArmorProfiles(containername, "DELETED", appArmorAnnotations)
+
```
daemon1024: We Add Policies to Feeder Map as well ( `dm.Logger.UpdateSecurityPolicies("ADDED", newPoint)`)

But we are only removing from AppArmor Profiles 🤔 Would this impact anything?

KubeArmor/core/kubeUpdate.go
```diff
@@ -1168,6 +1168,460 @@ func (dm *KubeArmorDaemon) WatchSecurityPolicies() {
...
+				dm.EndPoints = append(dm.EndPoints[:idx], dm.EndPoints[idx+1:]...)
+				break
+			}
+		}
+		dm.EndPointsLock.Unlock()
+	}
+}
+
+// WatchContainerSecurityPolicies Function
+func (dm *KubeArmorDaemon) WatchContainerSecurityPolicies() {
```
daemon1024: Do we need this function?

KubeArmor/config/config.go
Ankurk99: Ah right, so the original plan was to use GlobalCfg.ContainerPolicy for running KubeArmor in container mode, but GlobalCfg.Policy with -k8s=false will do the job.

KubeArmor/core/kubeArmor.go
Ankurk99: Yes, that would be simpler. The idea was that it would be implicit for the user to know that it's a container based enforcement/observability when the -enableKubeArmorContainerPolicy is set. But I think we can be good without it.

KubeArmor/core/kubeUpdate.go
Ankurk99: No, it was added as a wrapper function for testing purpose. Will remove it :+1:

Ankurk99:
Ref documentation: [Deploying KubeArmor to support un-orchestrated workloads](https://github.com/kubearmor/KubeArmor/wiki/Deploying-KubeArmor-to-protect-un-orchestrated-containerized-workloads).

nyrahul:
@Ankurk99 , pls handle the DCO and should be good to merge.

Ankurk99:
PR good to merge, PTAL. Thanks @daemon1024 for helping to debug it.

nyrahul (approved):
LGTM 👍🏼

KubeArmor/core/kubeArmor.go
```diff
@@ -311,6 +311,25 @@ func GetOSSigChannel() chan os.Signal {
...
 }
 
+// ======================= //
+// == Container runtime == //
+// ======================= //
+
+// host container runtime available in case of containerized workloads
+func hostContainerRuntime() string {
+	var sockFile string
+	if _, err := os.Stat("/var/run/docker.sock"); err == nil {
```
nyrahul: Lets leverage CRI_SOCKET env var that we have to identify the sock file path.
Point is, if CRI_SOCKET env var is set, just use that.
CC: @DelusionalOptimist

KubeArmor/core/kubeArmor.go
```diff
@@ -448,6 +469,64 @@ func KubeArmor() {
...
+			if _, err := os.Stat(trimmedSocket); err != nil {
+				dm.Logger.Warnf("Error while looking for CRI socket file: %s", err.Error())
+
+				// destroy the daemon
+				dm.DestroyKubeArmorDaemon()
+				return
+			}
+		}
+
+		if hostContainerRuntime() == "docker" {
```
nyrahul: lets keep a local var and then match it
```suggestion
                cri := hostContainerRuntime()
		if cri == "docker" {
```

KubeArmor/core/kubeArmor.go
```diff
@@ -448,6 +469,64 @@ func KubeArmor() {
...
+			if cfg.GlobalCfg.CRISocket == "" {
+
+				socketFile := kl.GetCRISocket("docker")
```
nyrahul: Better to get the socket path as an output from the hostContainerRuntime() API.

KubeArmor/core/kubeArmor.go
```diff
@@ -448,6 +469,64 @@ func KubeArmor() {
...
+				socketFile := kl.GetCRISocket("crio")
+
+				if socketFile != "" {
+					cfg.GlobalCfg.CRISocket = "unix://" + socketFile
+				}
+			}
+			// monitor cri-o events
+			go dm.MonitorCrioEvents()
+
+		} else if hostContainerRuntime() == "unavailable" {
```
nyrahul: Note that with this check if the k8s is disabled and if the policy is enabled but there is no container runtime available then the kubearmor will quit. Imo, kubearmor should still continue and apply policies at the host level and reject container based policies.

nyrahul (changes requested):

KubeArmor/core/kubeArmor.go
Ankurk99: @nyrahul Yes, doing the same. For eg, if you see the function call:  
``` 
if hostContainerRuntime() == "docker" {
			if cfg.GlobalCfg.CRISocket == "" {
				socketFile := kl.GetCRISocket("docker")
				if socketFile != "" {
					cfg.GlobalCfg.CRISocket = "unix://" + socketFile
				}
			}
			// update already deployed containers
			dm.GetAlreadyDeployedDockerContainers()
			// monitor docker events
			go dm.MonitorDockerEvents()
} 
```
Here if hostContainerRuntime is docker and -criSocket var is not set (cfg.GlobalCfg.CRISocket == "") then only we are setting the globalCfg.Crisocket to the docker sock file path

KubeArmor/core/kubeArmor.go
Ankurk99: Okay, makes sense. I will update it.

KubeArmor/Makefile
```diff
@@ -35,6 +35,11 @@ run: build
 	cd $(CURDIR); sudo rm -f /tmp/kubearmor.log
 	cd $(CURDIR); sudo -E ./kubearmor -logPath=/tmp/kubearmor.log -enableKubeArmorPolicy=true -enableKubeArmorHostPolicy=true -hostVisibility=process,file,network,capabilities
```
nam-jaehyun, line 36: can we remove =true for the consistency with line 41?

KubeArmor/Makefile
```diff
@@ -35,6 +35,11 @@ run: build
...
 
+.PHONY: run-container
+run-container: build
+	cd $(CURDIR); sudo rm -f /tmp/kubearmor.log
+	cd $(CURDIR); sudo -E ./kubearmor -logPath=/tmp/kubearmor.log -enableKubeArmorHostPolicy -enableKubeArmorPolicy -k8s=false
```
nam-jaehyun, line 51:
```suggestion
	cd $(CURDIR); sudo -E ./kubearmor -logPath=/tmp/kubearmor.log -enableKubeArmorPolicy -enableKubeArmorHostPolicy -k8s=false
```

KubeArmor/Makefile
nam-jaehyun:
```suggestion
	cd $(CURDIR); sudo -E ./kubearmor -logPath=/tmp/kubearmor.log -enableKubeArmorPolicy -enableKubeArmorHostPolicy -hostVisibility=process,file,network,capabilities
```

KubeArmor/core/dockerHandler.go
```diff
@@ -237,6 +237,18 @@ func (dm *KubeArmorDaemon) GetAlreadyDeployedDockerContainers() {
 					continue
 				}
 
+				// check for unorchestrated docker containers
+				if !dm.K8sEnabled {
+					dm.ContainersLock.Lock()
+					container.ProcessVisibilityEnabled = true
```
nam-jaehyun: do we simply enable all visibility flags rather than using given visibility flags?

KubeArmor/core/kubeArmor.go
nam-jaehyun: Can we handle the socket like this?
```
if globalCfg. CRISocket != "" {
  if _, err := os.Stat(globalCfg.CRISocket); err == nil {
    sockFile = "xxx"
    sockPath = globalCfg.CRISocket
  } else {
    sockFile = "unavailable"
    sockPath = "unavailable"
  }
} else {
   if _, err := os.Stat("/var/run/docker.sock") ...
   ...
   }
}
```

KubeArmor/policy/policy.go
```diff
@@ -15,10 +15,11 @@ import (
 // ServiceServer provides structure to serve Policy gRPC service
 type ServiceServer struct {
 	pb.PolicyServiceServer
-	UpdateHostPolicy func(tp.K8sKubeArmorHostPolicyEvent)
+	UpdateHostPolicy      func(tp.K8sKubeArmorHostPolicyEvent)
+	UpdateContainerPolicy func(tp.K8sKubeArmorPolicyEvent)
```
nam-jaehyun, line 18: ContainerPolicy comes first for better readability?

KubeArmor/policy/policy.go
```diff
@@ -15,10 +15,11 @@ import (
...
 }
 
-// HostPolicy accepts host policy event on gRPC service and updates host security polcies. It responds with 1 if success else 0.
+// HostPolicy accepts host policy event on gRPC service and updates host security policies. It responds with 1 if success else 0.
```
nam-jaehyun: ContainerPolicy comes first for better readability?

protobuf/policy.proto
```diff
@@ -14,4 +14,5 @@ message policy {
 
 service PolicyService {
     rpc hostPolicy (policy) returns (response);
+    rpc containerPolicy (policy) returns (response);
```
nam-jaehyun, line 16: ContainerPolicy comes first for better readability?

KubeArmor/core/kubeArmor.go
```diff
@@ -412,7 +435,9 @@ func KubeArmor() {
 
 	// == //
 
+	// Containerized workloads with Host
 	if cfg.GlobalCfg.Policy || cfg.GlobalCfg.HostPolicy {
+
```
nam-jaehyun: remove an empty line?

KubeArmor/core/kubeArmor.go
```diff
@@ -311,6 +311,29 @@ func GetOSSigChannel() chan os.Signal {
 	return c
 }
 
+// ======================= //
+// == Container runtime == //
+// ======================= //
+
+// host container runtime available in case of containerized workloads
+func hostContainerRuntime() (string, string) {
```
nam-jaehyun: what is the purpose of this function?

KubeArmor/core/kubeArmor.go
```diff
@@ -448,6 +473,44 @@ func KubeArmor() {
 
 	// == //
 
+	if !dm.K8sEnabled && cfg.GlobalCfg.Policy {
+		// check if the CRI socket set while executing kubearmor exists
+		if cfg.GlobalCfg.CRISocket != "" {
+			trimmedSocket := strings.TrimPrefix(cfg.GlobalCfg.CRISocket, "unix://")
```
nam-jaehyun: do you know when we use "trimmedSocket"?

KubeArmor/core/kubeArmor.go
```diff
@@ -448,6 +473,44 @@ func KubeArmor() {
...
+			if _, err := os.Stat(trimmedSocket); err != nil {
+				dm.Logger.Warnf("Error while looking for CRI socket file: %s", err.Error())
+			}
+		}
+		cri, criPath := hostContainerRuntime()
```
nam-jaehyun: we don't consider the case when a user gives KubeArmor a specific CRI path?

KubeArmor/core/kubeArmor.go
nam-jaehyun: what is different with kl.GetCRISocket()?

KubeArmor/core/kubeArmor.go
```diff
@@ -448,6 +473,44 @@ func KubeArmor() {
...
+			go dm.MonitorCrioEvents()
+
+		} else if cri == "unavailable" {
+			dm.Logger.Warnf("Failed to monitor containers. Unsupported container runtime")
+		}
+	}
+
 	if dm.K8sEnabled && cfg.GlobalCfg.Policy {
 		// check if the CRI socket set while executing kubearmor exists
 		if cfg.GlobalCfg.CRISocket != "" {
```
nam-jaehyun, line 491: Can you also check where trimmedSocket is used?

KubeArmor/core/kubeArmor.go
```diff
@@ -577,14 +640,22 @@ func KubeArmor() {
...
 		policyService.UpdateHostPolicy = dm.ParseAndUpdateHostSecurityPolicy
 
 		pb.RegisterPolicyServiceServer(dm.Logger.LogServer, policyService)
 		reflection.Register(dm.Logger.LogServer)
 
 		dm.Logger.Print("Started to monitor host security policies on gRPC")
+
+	}
+
+	if !dm.K8sEnabled && cfg.GlobalCfg.Policy {
```
nam-jaehyun: can you put this block above 645?

nam-jaehyun (changes requested):
I just added some comments. Please check them.

KubeArmor/Makefile
Ankurk99: Added -enableKubeArmorPolicy and -k8s=false flag together to imply that the it's a case of an un-orchestrated workload.

KubeArmor/core/dockerHandler.go
Ankurk99: Can you please mention what all other flags need to be enabled?

KubeArmor/core/kubeArmor.go
Ankurk99: To get the sock file path from the -crioSocket flag and to check whether the sock file exist or not?

KubeArmor/core/kubeArmor.go
Ankurk99: Makes sense, I missed that. Thanks

KubeArmor/core/kubeArmor.go
Ankurk99: I have made the required changes, PTAL

KubeArmor/core/kubeArmor.go
```diff
@@ -577,8 +631,17 @@ func KubeArmor() {
 		dm.Logger.Print("Started to monitor host security policies")
 	}
 
+	policyService := &policy.ServiceServer{}
+
+	if !dm.K8sEnabled && cfg.GlobalCfg.Policy {
+		trimmedSocket := strings.TrimPrefix(cfg.GlobalCfg.CRISocket, "unix://")
```
nyrahul: Lets use a function to handle this and set a var at the function level.

KubeArmor/core/dockerHandler.go
nyrahul: Lets set the visibility based on the env var like we do in the case of k8s.

nyrahul (changes requested):

KubeArmor/core/dockerHandler.go
```diff
@@ -237,6 +237,25 @@ func (dm *KubeArmorDaemon) GetAlreadyDeployedDockerContainers() {
...
+					}
+					if strings.Contains(cfg.GlobalCfg.Visibility, "file") {
+						container.FileVisibilityEnabled = true
+					}
+					if strings.Contains(cfg.GlobalCfg.Visibility, "network") {
+						container.NetworkVisibilityEnabled = true
+					}
+					if strings.Contains(cfg.GlobalCfg.Visibility, "capabilities") {
+						container.CapabilitiesVisibilityEnabled = true
+					}
```
nyrahul: Lets convert this into a function call and call it below as well.

KubeArmor/common/common.go
```diff
@@ -383,9 +383,15 @@ var ContainerRuntimeSocketMap = map[string][]string{
...
 func GetCRISocket(ContainerRuntime string) string {
-	for _, candidate := range ContainerRuntimeSocketMap[ContainerRuntime] {
-		if _, err := os.Stat(candidate); err == nil {
-			return candidate
+	for k := range ContainerRuntimeSocketMap {
+		if ContainerRuntime != "" && k != ContainerRuntime {
+			continue
+		}
+		criruntime := k
+		for _, candidate := range ContainerRuntimeSocketMap[criruntime] {
```
nyrahul, line 391:
```suggestion
		for _, candidate := range ContainerRuntimeSocketMap[k] {
```

KubeArmor/core/kubeArmor.go
```diff
@@ -446,12 +451,44 @@ func KubeArmor() {
...
+	if !dm.K8sEnabled && cfg.GlobalCfg.Policy {
+
+		// Check if cri socket set, if not then auto detect
+		if cfg.GlobalCfg.CRISocket == "" {
+			if kl.GetCRISocket("") == "" {
+				dm.Logger.Warnf("Error while looking for CRI socket file")
+			} else {
+				cfg.GlobalCfg.CRISocket = "unix://" + kl.GetCRISocket("")
+				// update the value of trimmed socket path when the cfg.GlobalCfg.CRISocket is not set
+				trimmedSocket = trimmedSocketPath()
```
nyrahul, line 466: What is the use of trimmedSocket below?
All we need is a flag saying `enableContainerMonitoring := true` and use it in place of `os.Stat(trimmedSocket)` below. Note that `GetCRISocket()` returns successfully only when the path is available. Hence setting a flag should be enough.

nyrahul, line 466:
```suggestion
				enableContainerPolicy = true
```

KubeArmor/core/kubeArmor.go
```diff
@@ -577,8 +614,16 @@ func KubeArmor() {
...
 	}
 
+	policyService := &policy.ServiceServer{}
+
+	if !dm.K8sEnabled && cfg.GlobalCfg.Policy {
+		if _, err := os.Stat(trimmedSocket); err == nil {
+			policyService.UpdateContainerPolicy = dm.ParseAndUpdateContainerSecurityPolicy
+			dm.Logger.Print("Started to monitor container security policies on gRPC")
+		}
+	}
```
nyrahul, line 624:
```suggestion
	if enableContainerPolicy {
		policyService.UpdateContainerPolicy = dm.ParseAndUpdateContainerSecurityPolicy
		dm.Logger.Print("Started to monitor container security policies on gRPC")
	}
```

KubeArmor/core/kubeArmor.go
```diff
@@ -446,12 +451,44 @@ func KubeArmor() {
 		}
 	}
 
-	// == //
+	trimmedSocket := trimmedSocketPath()
```
nyrahul, line 454: I am not sure why we are calling `trimmedSocketPath()` at this point?
`trimmedSocket` seems to be used only `if !dm.K8sEnabled && cfg.GlobalCfg.Policy`.

nyrahul (changes requested):

KubeArmor/core/kubeArmor.go
Ankurk99: Actually we are also using it here: https://github.com/Ankurk99/KubeArmor/blob/main/KubeArmor/core/kubeArmor.go#L492

KubeArmor/core/kubeArmor.go
Ankurk99: Don't we need this check? Added it for the condition when we don't set the criSocket flag and the cri socket file is also not available on the system.

KubeArmor/core/dockerHandler.go
Ankurk99: comment seems to be outdated

KubeArmor/core/kubeArmor.go
Ankurk99: trimmedSocket := trimmedSocketPath() on line 454 checks if the sock file exist, but if the GlobalCfg.CRISocket is not set then we are setting it here, but the trimmedSocket value set above won't be updated, that's why it's called here to reflect the updated value of CRISocket flag.

nyrahul (approved):

PR #715, opened 2022-05-23
add debug info

nam-jaehyun:
- add debug info

### DIFF
--- a/KubeArmor/core/kubeArmor.go
+++ b/KubeArmor/core/kubeArmor.go
@@ -336,10 +336,16 @@ func KubeArmor() {
 		dm.Node.Annotations = map[string]string{}
 		dm.HandleNodeAnnotations(&dm.Node)
 
+		hostInfo := kl.GetCommandOutputWithoutErr("hostnamectl", []string{})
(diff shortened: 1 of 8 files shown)


KubeArmor/core/kubeArmor.go
```diff
@@ -336,11 +336,12 @@ func KubeArmor() {
 		dm.Node.Annotations = map[string]string{}
 		dm.HandleNodeAnnotations(&dm.Node)
 
+		dm.Node.OSImage = kl.GetCommandOutputWithoutErr("lsb_release", []string{"-d", "-s"})
```
nyrahul: `lsb_release` is not available on most platforms. Mostly debian/ubuntu it is available.
A more portable way to check this is using `hostnamectl`.

nyrahul (changes requested):

KubeArmor/core/kubeArmor.go
nam-jaehyun: used hostnamectl instead of lsb_release

nyrahul (approved):

PR #716, opened 2022-05-23
fix parentProcessName and processName

nam-jaehyun:
additional fix on #703.


nyrahul (approved):
LGTM

PR #717, opened 2022-05-24
fix github action workflow to create stable release

Ankurk99:
updated latest-release.yml to create both the latest release and stable version release.


nyrahul (approved):
LGTM

PR #724, opened 2022-05-30
Update default deny/audit behaviour

daemon1024:
Reintroduce default deny/audit changes [`f383637` (#602)](https://github.com/kubearmor/KubeArmor/pull/602/commits/f383637327c4f440c82316584e9bdeacaa3f9a10)
Revert default deny/audit changes in [`30c8f54` (#668)](https://github.com/kubearmor/KubeArmor/pull/668/commits/30c8f54e68c4600f54c90fc14ba6b4458c475ee1)

TODO
- [x] Add more test cases
- [x] Verify Audit Mode behaviour

### DIFF
--- a/KubeArmor/enforcer/appArmorProfile.go
+++ b/KubeArmor/enforcer/appArmorProfile.go
@@ -675,21 +675,20 @@ func (ae *AppArmorEnforcer) GenerateProfileHead(numProcessWhiteList, numFileWhit
 	profileHead := "  #include <abstractions/base>\n"
 	profileHead = profileHead + "  umount,\n"
 
-	if defaultPosture.FileAction == "block" && !(numProcessWhiteList > 0 || numFileWhiteList > 0 || !fromSourceFile) {
-		profileHead = profileHead + "  file,\n"
-	} else if numProcessWhiteList == 0 && numFileWhiteList == 0 {
+	// Allow Access to Resource when
+	// -> Default Posture is not Block
+	// OR
+	// -> No allow policy AND No from source allow policy
+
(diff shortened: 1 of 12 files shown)


KubeArmor/enforcer/appArmorProfile.go
```diff
@@ -675,21 +675,20 @@ func (ae *AppArmorEnforcer) GenerateProfileHead(numProcessWhiteList, numFileWhit
...
 
-	if defaultPosture.FileAction == "block" && !(numProcessWhiteList > 0 || numFileWhiteList > 0 || !fromSourceFile) {
-		profileHead = profileHead + "  file,\n"
-	} else if numProcessWhiteList == 0 && numFileWhiteList == 0 {
+	// Allow Access to Resource when
+	// -> Default Posture is not Block
+	// OR
+	// -> No allow policy AND No from source allow policy
+
+	if !(defaultPosture.FileAction == "block") || !(numProcessWhiteList > 0 || numFileWhiteList > 0 || !fromSourceFile) {
```
nam-jaehyun, line 683: it doesn't make sense since it will make all logs to alerts regardless of given actions if defaultPosture is not 'block'. In other words, even though we have an allow policy, the policy would be meaningless if defaultPosture is not "block".

daemon1024: 1. This logic is just for enforcement and not telemetry, so doesn't impact conversion of logs to alerts.
2. Yes if the defaultPosture is not set to block i.e. it's in Audit mode, our enforcer shouldn't be active.

For 
> make all logs to alerts regardless of given actions

We explicitly check if a default posture is in Audit/Block mode then we convert logs to alerts.
We rely on our policyMatcher to match Allow policies and discard them so they don't come out as neither log nor alerts.

daemon1024:
Handled in #727

PR #727, opened 2022-06-06
fix default posture and audit mode

nam-jaehyun:
- update default posture and audit mode
in this fix, we have some redundant code in order to improve readability.

```
$ make vagrant-up RUNTIME=k3s
```

```
ubuntu $ cd KubeArmor/tests
ubuntu $ DEFAULT_POSTURE=allow ./test-scenarios-local.sh
ubuntu $ DEFAULT_POSTURE=audit ./test-scenarios-local.sh
ubuntu $ DEFAULT_POSTURE=block ./test-scenarios-local.sh
```

- update vagrant (k3s)
we can test the audit mode of KubeArmor in CentOS.

```
$ make vagrant-up RUNTIME=k3s OS=centos
```

```
centos $ cd KubeArmor/tests
centos $ DEFAULT_POSTURE=allow ./test-scenarios-local.sh
centos $ DEFAULT_POSTURE=audit ./test-scenarios-local.sh
centos $ DEFAULT_POSTURE=block ./test-scenarios-local.sh
```

- update test scripts
in order to test default posture, we can use "default" in result and action.
please refer to tests/scenarios/github_test_12/cmd3.

- fix failed test scenarios
github_test_09, github_test_10


nam-jaehyun:
the way to enforce policies

- default posture = block

    - container level

        if there is at least one allow policy or fromSource-based allow policy

            if matched -> allow / otherwise -> block

        if there is no allow policy and fromSource-based allow policy

            allow by default

    - fromSource level

        if there is at least one allow policy or fromSource-based allow policy

            under the same source -> if matched -> allow / otherwise -> block
            under different sources -> block

        if there is no allow policy and fromSource-based allow policy

            allow by default

- default posture = audit

    - container level

        if there is at least one allow policy or fromSource-based allow policy

            if matched -> allow / otherwise -> audit (= allow)

        if there is no allow policy and fromSource-based allow policy

            allow by default

    - fromSource level

        if there is at least one allow policy or fromSource-based allow policy

            under the same source -> if matched -> allow / otherwise -> audit (= allow)
            under different sources -> audit (= allow)

        if there is no allow policy and fromSource-based allow policy

            allow by default

- default posture = allow

   - disable whitelist-based policy enforcement
    - skip 'allow' policies while enforcing 'audit' and 'block' policies

nyrahul:
> * default posture = allow (required since host policies follow this way)
>   
>   * container level
>     if there is at least one allow policy or fromSource-based allow policy
>     ```
>       if matched -> allow / otherwise -> block
>     ```
>       
>     if there is no allow policy and fromSource-based allow policy
>     ```
>       allow by default
>     ```
>   * fromSource level
>     if there is at least one allow policy or fromSource-based allow policy
>     ```
>       under the same source -> if matched -> allow / otherwise -> block
>       under different sources -> allow
>     ```
>       
>     if there is no allow policy and fromSource-based allow policy
>     ```
>       allow by default
>     ```

For default posture of Block/Audit, the rules seem ok. But the rule for Allow seem problematic!!

Assuming `defaultPosture: Allow`, lets take an example:
```yaml
spec:
  process:
    matchPaths:
    - fromSource:
       - path: /bin/bash
       path: /bin/ls
  action:
    Allow
```
Now if I execute `sleep` from `bash`, it will be permission denied, based on the rule mentioned!!
If I run `sleep` as a child process of another shell `sh`, then it will be allowed.

nam-jaehyun:
@nyrahul yes, you're right. sleep from bash will be blocked while sleep from sh will be allowed.
this is how host policies work since we do not have global allow policies for hosts.

nyrahul:
@nam-jaehyun , I have the following proposal.

Assume `defaultPosture: Block`:
```yaml
spec:
  process:
    matchPaths:
    - fromSource:
       - path: /bin/bash
       path: /bin/ls
   - path: {}
  action:
    Allow
```
With the above policy, all sub-processes of `/bin/bash` will be blocked except `ls`. The `-path: {}` specifies that allow all other processes. Thus, all other processes will be allowed spawned outside of bash. Even if any policy has this rule with Allow action then all the processes will be allowed.

The auto-generated policies will not have `-path: {}` as part of it.

WDYT?

nyrahul:
There is one more scenario we may have to consider:
How to allow all sub-processes from a particular parent process and block everything else?
I dont think we support this today. Do we?

nyrahul (approved):
LGTM. Thanks for handling this.

PR #740, opened 2022-06-20
[bugfix] fix security profile clash 

### LINKED ISSUE #700: security profile clash when namespace and container-name is same
## Bug Report

kubearmor does not handle properly the situation where there is multiple containers with the same name in the same namespace ( this situation is common when using charts e.g gitlab autodevops) which results in unexpected behavior (policies are not enforced, policies are being applied to wrong containers etc...).

<details><summary>Sample Deployment with same container name (deploy.yaml)</summary>
<p>

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: myapp1
spec:
  selector:
    matchLabels:
      app: myapp1
  template:
    metadata:
      labels:
        app: myapp1
    spec:
      containers:
      - name: ubuntu
        image: ubuntu
        command:
          - "bash"
        args:
          - "-c"
          - "sleep infinity"
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: myapp2
spec:
  selector:
    matchLabels:
      app: myapp2
  template:
    metadata:
      labels:
        app: myapp2
    spec:
      containers:
      - name: ubuntu
        image: ubuntu
        command:
          - "bash"
        args:
          - "-c"
          - "sleep infinity"
```

</p>
</details>

<details><summary>Security policy for myapp1 deployment (myapp1-policy.yaml)</summary>
<p>

```yaml
kind: KubeArmorPolicy
metadata:
  name: policy1
spec:
  selector:
    matchLabels:
      app: myapp1
  process:
    matchPaths:
    - path: /usr/bin/ls
  action:
    Block
```
</p>
</details>

<details><summary>Security policy for myapp2 deployment (myapp2-policy.yaml)</summary>
<p>

```yaml
kind: KubeArmorPolicy
metadata:
  name: policy2
spec:
  selector:
    matchLabels:
      app: myapp2
  process:
    matchPaths:
    - path: /usr/bin/cat
  action:
    Block
```
</p>
</details>

**General Information**

- Environment description: Any
- Kernel version: ANY
- Orchestration system version in use (e.g. `kubectl version`, ...): ANY
- Link to relevant artifacts (policies, deployments scripts, ...): Given in the description above

**To Reproduce**

1. `kubectl apply -f deploy.yaml`
2. `kubectl apply -f myapp1-policy.yaml`

This policy should impact only myapp1, but currently this will also apply the rules to myapp2.

**What is the root cause?**
Currently, we create an apparmor profile using following template `kubearmor-NAMESPACE-CONTAINERNAME`. Thus if there are multiple deployments in the same namespace having same containername then it would apply to all the similar deployments.

**Expected behavior**

Applying `myapp1-policy.yaml` should not impact myapp2 deployment.

**Possible Solution**

Add pod name (without instance id) to the security profile template name for eg. `kubearmor-NAMESPACE-PODNAME-CONTAINERNAME`.

Credits to @achrefbensaad for discovering this issue and reporting it in detail on slack. Thanks

DelusionalOptimist:
Signed-off-by: Rudraksh Pareek <rudraksh@accuknox.com>

Scenarios:
- [x] Deploy kubearmor then deploy pods
- [x] Deploy workloads then deploy kubearmor
- [x] The profile name has deployment's name and is unaffected by changing template hash of pods
- [x]  Workloads annotated by older version of kubearmor comply with these changes. (Handling update)

TODO:
- [x] Update annontations controller to use similar naming convention

This PR fixes #700

### DIFF
--- a/KubeArmor/core/kubeUpdate.go
+++ b/KubeArmor/core/kubeUpdate.go
@@ -6,6 +6,7 @@ package core
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
@@ -452,17 +453,27 @@ func (dm *KubeArmorDaemon) WatchK8sPods() {
 
 				pod := tp.K8sPod{}
 
+				// need this for apparmor profile
+				var podOwnerName string
+
 				pod.Metadata = map[string]string{}
 				pod.Metadata["namespaceName"] = event.Object.ObjectMeta.Namespace
 				pod.Metadata["podName"] = event.Object.ObjectMeta.Name
 
-				if len(event.Object.ObjectMeta.OwnerReferences) > 0 {
-					if event.Object.ObjectMeta.OwnerReferences[0].Kind == "ReplicaSet" {
-						deploymentName := K8s.GetDeploymentNameControllingReplicaSet(pod.Metadata["namespaceName"], event.Object.ObjectMeta.OwnerReferences[0].Name)
+				ownerRef := kl.GetControllingPodOwner(event.Object.ObjectMeta.OwnerReferences)
+				if ownerRef != nil {
+					podOwnerName = ownerRef.Name
+					if ownerRef.Kind == "ReplicaSet" {
+						deploymentName := K8s.GetDeploymentNameControllingReplicaSet(pod.Metadata["namespaceName"], podOwnerName)
 						if deploymentName != "" {
 							pod.Metadata["deploymentName"] = deploymentName
 						}
+						// if it belongs to a replicaset, we also remove the pod template hash
+						podOwnerName = strings.TrimSuffix(podOwnerName, fmt.Sprintf("-%s", event.Object.ObjectMeta.Labels["pod-template-hash"]))
 					}
+				} else {
+					// static pod
+					podOwnerName = event.Object.ObjectMeta.Name
 				}
 
 				pod.Annotations = map[string]string{}
@@ -592,7 +603,7 @@ func (dm *KubeArmorDaemon) WatchK8sPods() {
 
 					for _, container := range event.Object.Spec.Containers {
 						if _, ok := appArmorAnnotations[container.Name]; !ok {
-							appArmorAnnotations[container.Name] = "kubearmor-" + pod.Metadata["namespaceName"] + "-" + container.Name
+							appArmorAnnotations[container.Name] = "kubearmor-" + pod.Metadata["namespaceName"] + "-" + podOwnerName + "-" + container.Name
 							updateAppArmor = true
 						}
 					}
--- a/pkg/KubeArmorAnnotation/handlers/pod_mutation.go
+++ b/pkg/KubeArmorAnnotation/handlers/pod_mutation.go
@@ -6,6 +6,7 @@ package handlers
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"strings"
 
@@ -113,6 +114,27 @@ func (a *PodAnnotator) InjectDecoder(d *admission.Decoder) error {
 // == Add AppArmor annotations == //
 func appArmorAnnotator(pod *corev1.Pod) {
 	podAnnotations := map[string]string{}
+	var podOwnerName string
+
+	// podOwnerName is the pod name for static pods and parent object's name
+	// in other cases
+	for _, ownerRef := range pod.ObjectMeta.OwnerReferences {
+		// pod is owned by a replicaset, daemonset etc thus we use the managing
+		// controller's name
+		if *ownerRef.Controller {
+			podOwnerName = ownerRef.Name
+
+			if ownerRef.Kind == "ReplicaSet" {
+				// if it belongs to a replicaset, we also remove the pod template hash
+				podOwnerName = strings.TrimSuffix(podOwnerName, fmt.Sprintf("-%s", pod.ObjectMeta.Labels["pod-template-hash"]))
+			}
+		}
+	}
+
+	if podOwnerName == "" {
+		// pod is standalone, name remains constant
+		podOwnerName = pod.ObjectMeta.Name
+	}
 
 	// Get existant kubearmor annotations
 	for k, v := range pod.Annotations {
@@ -130,7 +152,7 @@ func appArmorAnnotator(pod *corev1.Pod) {
 	// Get the remaining containers / not addressed explecitly in the annotation
 	for _, container := range pod.Spec.Containers {
 		if _, ok := podAnnotations[container.Name]; !ok {
-			podAnnotations[container.Name] = "kubearmor-" + pod.Namespace + "-" + container.Name
+			podAnnotations[container.Name] = "kubearmor-" + pod.Namespace + "-" + podOwnerName + "-" + container.Name
 		}
 	}
 
--- a/KubeArmor/common/common.go
+++ b/KubeArmor/common/common.go
@@ -18,6 +18,7 @@ import (
 
 	kc "github.com/kubearmor/KubeArmor/KubeArmor/config"
 	kg "github.com/kubearmor/KubeArmor/KubeArmor/log"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // ============ //
@@ -388,6 +389,16 @@ func GetCRISocket(ContainerRuntime string) string {
 		}
 	}
 	return ""
+}
+
+// GetControllingPodOwner Function returns the pod's Controlling OnwerReference
+func GetControllingPodOwner(ownerRefs []metav1.OwnerReference) *metav1.OwnerReference {
+	for _, ownerRef := range ownerRefs {
+		if *ownerRef.Controller {
+			return &ownerRef
+		}
+	}
+	return nil
 }
 
 // ==================== //
--- a/contribution/self-managed-k8s/k8s/initialize_kubernetes.sh
+++ b/contribution/self-managed-k8s/k8s/initialize_kubernetes.sh
@@ -9,8 +9,9 @@ fi
 
 # use docker as default CRI
 if [ "$CRI_SOCKET" == "" ]; then
+    # if docker, let kubeadm figure it out
     if [ -S /var/run/docker.sock ]; then
-        CRI_SOCKET=unix:///var/run/docker.sock
+        CRI_SOCKET=""
     elif [ -S /var/run/containerd/containerd.sock ]; then
         CRI_SOCKET=unix:///var/run/containerd/containerd.sock
     elif [ -S /var/run/crio/crio.sock ]; then


DelusionalOptimist:
To be addressed later. Closing for now.

KubeArmor/core/kubeUpdate.go
```diff
@@ -456,36 +457,42 @@ func (dm *KubeArmorDaemon) WatchK8sPods() {
...
+
 					if event.Object.ObjectMeta.OwnerReferences[0].Kind == "ReplicaSet" {
 						deploymentName := K8s.GetDeploymentNameControllingReplicaSet(pod.Metadata["namespaceName"], event.Object.ObjectMeta.OwnerReferences[0].Name)
 						if deploymentName != "" {
 							pod.Metadata["deploymentName"] = deploymentName
 						}
+
+						// if it belongs to a replicaset, we also remove the pod template hash
+						podTemplateHashSuffix := fmt.Sprintf("-%s", pod.Labels["pod-template-hash"])
+						pod.Metadata["ownerName"] = strings.TrimSuffix(event.Object.ObjectMeta.OwnerReferences[0].Name, podTemplateHashSuffix)
```
nyrahul: When I checked my sample deployment, I have a pod that was instantiated with name `wordpress-54dd4449df-wq7ql` ... but the `pod-template-hash` was only `54dd4449df`. `TrimSuffix` won't be able to trim in this case.

```
vagrant@kubearmor-dev:~/KubeArmor/tests$ kubectl describe pod -n wordpress-mysql wordpress-54dd4449df-wq7ql
Name:         wordpress-54dd4449df-wq7ql
Namespace:    wordpress-mysql
Priority:     0
Node:         kubearmor-dev/10.0.2.15
Start Time:   Wed, 06 Jul 2022 07:25:24 +0000
Labels:       app=wordpress
              pod-template-hash=54dd4449df
Annotations:  container.apparmor.security.beta.kubernetes.io/wordpress: localhost/kubearmor-wordpress-mysql-wordpress
              kubearmor-policy: enabled
Status:       Running
IP:           10.0.0.69
IPs:
  IP:           10.0.0.69
Controlled By:  ReplicaSet/wordpress-54dd4449df
Containers:
  wordpress:
    Container ID:   docker://b503dfed67c4ec54c01b9f91437a089bf44913092bd1da66b38f43c1a7f94b32
    Image:          wordpress:4.8-apache
    Image ID:       docker-pullable://wordpress@sha256:6216f64ab88fc51d311e38c7f69ca3f9aaba621492b4f1fa93ddf63093768845
    Port:           80/TCP
    Host Port:      0/TCP
    State:          Running
      Started:      Wed, 06 Jul 2022 07:25:32 +0000
    Ready:          True
    Restart Count:  0
    Environment:
      WORDPRESS_DB_HOST:      mysql
      WORDPRESS_DB_PASSWORD:  root-password
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-b4j9b (ro)
Conditions:
  Type              Status
  Initialized       True 
  Ready             True 
  ContainersReady   True 
  PodScheduled      True 
Volumes:
  kube-api-access-b4j9b:
    Type:                    Projected (a volume that contains injected data from multiple sources)
    TokenExpirationSeconds:  3607
    ConfigMapName:           kube-root-ca.crt
    ConfigMapOptional:       <nil>
    DownwardAPI:             true
QoS Class:                   BestEffort
Node-Selectors:              <none>
Tolerations:                 node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
                             node.kubernetes.io/unreachable:NoExecute op=Exists for 300s
Events:                      <none>
```

nyrahul (changes requested):
Thanks for handling this on prio. Please find my review comment.

KubeArmor/core/kubeUpdate.go
nyrahul: I am mistaken here. The name used is from replicaset which is in the form of `wordpress-54dd4449df`.

DelusionalOptimist: :white_check_mark:

DelusionalOptimist:
@nyrahul @achrefbensaad About the annotation controller.
I was looking back and in #360  we wanted to move the pod annotation logic entirely to the annotations controller, however we still have some of it in kubearmor. Need to know this so as to decide where these changes belong :thinking:

daemon1024 (commented):
LGTM

> Workloads annotated by older version of kubearmor comply with these changes. (Handling update)

Thhis would be a breaking change right? 

> I was looking back and in https://github.com/kubearmor/KubeArmor/issues/360 we wanted to move the pod annotation logic entirely to the annotations controller, however we still have some of it in kubearmor.

AFAIK Annotation handling inside KubeArmor is fallback if the controller is not present.

DelusionalOptimist:
> Thhis would be a breaking change right?

Yes @daemon1024, a breaking change. Described more in - https://github.com/kubearmor/KubeArmor/issues/700#issuecomment-1175797617.

> AFAIK Annotation handling inside KubeArmor is fallback if the controller is not present.

Then it needs to go in both the places ig :thinking:

KubeArmor/core/kubeUpdate.go
```diff
@@ -456,36 +457,42 @@ func (dm *KubeArmorDaemon) WatchK8sPods() {
...
+				pod.Labels = map[string]string{}
+				for k, v := range event.Object.Labels {
+					if k == "pod-template-generation" {
+						continue
+					}
+
+					if k == "controller-revision-hash" {
+						continue
+					}
+					pod.Labels[k] = v
```
nyrahul: The labels that are added as part of pod.Labels are shipped in every alert/telemetry. Sending `pod-template-hash` in such telemetry events does not make sense.
Can you check with these changes if the pod-labels (in json format) now contains `pod-template-hash` as well?

nyrahul (commented):
Please check one inline comment.

KubeArmor/core/kubeUpdate.go
DelusionalOptimist: Yeah they do 
```
{"Timestamp":1657109207,"UpdatedTime":"2022-07-06T12:06:47.678678Z","ClusterName":"default","HostName":"pingu-vm","NamespaceName":"default","PodName":"myapp1-5dddbf7478-8p8r8","Labels":"app=myapp1,pod-template-hash=5dddbf7478","ContainerID":"b529bd2f301f6308334761c792851f8adbe7dae2cffc560159e5bce2cf9649a9","ContainerName":"ubuntu","ContainerImage":"ubuntu:latest@sha256:b6b83d3c331794420340093eb706a6f152d9c1fa51b262d9bf34594887c2c7ac","HostPPID":2843709,"HostPID":2843839,"PPID":7,"PID":15,"ParentProcessName":"/usr/bin/bash","ProcessName":"/usr/bin/ls","PolicyName":"policy1","Severity":"1","Type":"MatchedPolicy","Source":"/usr/bin/bash","Operation":"Process","Resource":"/usr/bin/ls --color=auto","Data":"syscall=SYS_EXECVE","Enforcer":"AppArmor","Action":"Block","Result":"Permission denied"}
```

DelusionalOptimist: No worries, we only need it once anyways. Won't use this map :+1:

DelusionalOptimist:
Folks, I've tested this out with deployments (with multiple replicas), daemonsets and static pods. Lmk if some resource other than these might behave differently.
Also, as we discussed, I tried the update scenario with `karmor uninstall --force` and installing kubearmor after that, we do get profiles with the new convention and enforcement works as expected. However, one thing to note is that the `--force` flag only removes annotations on pods owned by deployments.

nyrahul:
> Folks, I've tested this out with deployments (with multiple replicas), daemonsets and static pods. Lmk if some resource other than these might behave differently. Also, as we discussed, I tried the update scenario with `karmor uninstall --force` and installing kubearmor after that, we do get profiles with the new convention and enforcement works as expected. However, one thing to note is that the `--force` flag only removes annotations on pods owned by deployments.

I guess your point is that the old apparmor policies with old names are still retained in the `/etc/apparmor.d` even though the new ones are created. I believe that should be fine.

pkg/KubeArmorAnnotation/handlers/pod_mutation.go
```diff
@@ -113,6 +114,21 @@ func (a *PodAnnotator) InjectDecoder(d *admission.Decoder) error {
...
+		podOwnerName = pod.ObjectMeta.OwnerReferences[0].Name
+
+		if pod.ObjectMeta.OwnerReferences[0].Kind == "ReplicaSet" {
+			podTemplateHashSuffix := fmt.Sprintf("-%s", pod.ObjectMeta.Labels["pod-template-hash"])
+			podOwnerName = strings.TrimSuffix(podOwnerName, podTemplateHashSuffix)
+		}
+	} else {
+		podOwnerName = pod.ObjectMeta.Name
+	}
+	fmt.Println("podOwnerName is : ", podOwnerName)
```
nyrahul: Lets use the normal logging APIs.

KubeArmor/core/kubeUpdate.go
```diff
@@ -486,6 +476,29 @@ func (dm *KubeArmorDaemon) WatchK8sPods() {
 					pod.Labels[k] = v
 				}
 
+				if len(event.Object.ObjectMeta.OwnerReferences) > 0 {
+					// pod is owned by a replicaset, daemonset etc thus we use parent object's name
+					pod.Metadata["ownerName"] = event.Object.ObjectMeta.OwnerReferences[0].Name
```
nyrahul: Is it necessary to keep the `ownerName` as part of the `pod.Metadata[]`?
Can we keep it as a local var in the function? I am worried someone might start using `pod.Metadata["ownerName"]` it in the future and it might not be filled up in all cases.

nyrahul (changes requested):

KubeArmor/core/kubeUpdate.go
```diff
@@ -486,6 +476,29 @@ func (dm *KubeArmorDaemon) WatchK8sPods() {
 					pod.Labels[k] = v
 				}
 
+				if len(event.Object.ObjectMeta.OwnerReferences) > 0 {
```
nyrahul: what happens if there are multiple event.Object.ObjectMeta.OwnerReferences .. currently we are using ref at index 0 .. I am just worried that we may not know which index holds the right ref that we may be interested in.

DelusionalOptimist: @nyrahul please take a look now.

DelusionalOptimist: We go through all OwnerReferences and use https://github.com/kubernetes/design-proposals-archive/blob/main/api-machinery/controller-ref.md#api to conclude whether a reference is to the pod's managing controller.
There can be only one controller reference per pod.

pkg/KubeArmorAnnotation/handlers/pod_mutation.go
```diff
@@ -113,6 +114,27 @@ func (a *PodAnnotator) InjectDecoder(d *admission.Decoder) error {
 // == Add AppArmor annotations == //
 func appArmorAnnotator(pod *corev1.Pod) {
 	podAnnotations := map[string]string{}
+	var podOwnerName string
+
+	// podOwnerName is the pod name for static pods and parent object's name
+	// in other cases
+	for _, ownerRef := range pod.ObjectMeta.OwnerReferences {
```
nyrahul, line 121: Can we use the newly created `GetControllingPodOwner()` to replace this loop?

nyrahul (changes requested):

nyrahul (approved):

pkg/KubeArmorAnnotation/handlers/pod_mutation.go
DelusionalOptimist: I couldn't find any example where we were using something from kubearmor's packages in any of the controllers so doing it here kind of felt like an anti-pattern.
Also, the annotations controller and kubearmor use different k8s API versions so we'll have to update the API version used in the controller if we import something or we'll face bugs while building. I thought it would further complicate this PR :sweat_smile:

PR #743, opened 2022-06-24
fix k3s installation script to use proper kubectl command

Ankurk99:


nam-jaehyun:
@Ankurk99 I used the absolute path because kubectl is not in $PATH during 'make vagrant-up OS=centos'. Can you check the centos-based vagrant env?

Ankurk99:
> @Ankurk99 I used the absolute path because kubectl is not in $PATH during 'make vagrant-up OS=centos'. Can you check the centos-based vagrant env?

I see. But some distro installs kubectl at /usr/bin/kubectl, wouldn't that be an issue?

nyrahul (approved):

PR #741, opened 2022-06-20
KubeArmor BPF LSM Enforcer

daemon1024:
- [x] Setup eBPF objects for BPF LSM program
- [x] Handle eBPF Maps on Container Events
- [x] Handle Default Posture Decisions and Eliminate Conflicting rules
- [x] Extract Path and Source in each set of Hooks
    - Currently Limited to 256 byte length
- [x] Basic Enforcement and Decision Making
	- [x] Process
	- [x] File
	- [x] Network
- [x] Update Dev Env
- Testing - Deferred till #760 

[Design Doc](https://github.com/kubearmor/KubeArmor/wiki/KubeArmor-BPF-LSM-Integration)

Ref #484

PR #747, opened 2022-06-30
cri: handle bugs and edge cases

daemon1024:
- It is possible that cri socket be named something different and cri runtime be different like EKS uses dockershim as socket but the runtime is containerd. So we decide based on the runtime and not configured socket

- Recent changes introduced docker getting instantiated later at runtime and not init time due to which we were not fetching already deployed containers, so added a check to instantiate docker handler if nil

### DIFF
--- a/KubeArmor/core/kubeArmor.go
+++ b/KubeArmor/core/kubeArmor.go
@@ -461,15 +461,15 @@ func KubeArmor() {
 			}
 
 			// monitor containers
-			if strings.Contains(cfg.GlobalCfg.CRISocket, "docker") {
+			if strings.Contains(dm.Node.ContainerRuntimeVersion, "docker") {
 				// update already deployed containers
 				dm.GetAlreadyDeployedDockerContainers()
 				// monitor docker events
 				go dm.MonitorDockerEvents()
-			} else if strings.Contains(cfg.GlobalCfg.CRISocket, "containerd") {
+			} else if strings.Contains(dm.Node.ContainerRuntimeVersion, "containerd") {
 				// monitor containerd events
 				go dm.MonitorContainerdEvents()
-			} else if strings.Contains(cfg.GlobalCfg.CRISocket, "crio") {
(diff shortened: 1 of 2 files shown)


nyrahul (approved):
LGTM

DelusionalOptimist (approved):
LGTM :rocket:

nam-jaehyun:
@daemon1024 can we double-check possible cases for each runtime (docker, containerd, and cri-o / not just for docker)?

nam-jaehyun:
@daemon1024 is it a part of the updates with the BPFLSM enforcer (#741)?

daemon1024:
@nam-jaehyun just checked again.

In case of docker we have separate functions for getting previously deployed containers and new container, so the need to validate initialization. 

In containerd and cri-o we have a single channel handling them both so it should be fine as it is.

daemon1024:
@nam-jaehyun It should be independent of https://github.com/kubearmor/KubeArmor/pull/741 but I need this fix for KubeArmor to work with EKS(regardless of BPF LSM), I will rebase #741 once this is merged.

KubeArmor/core/kubeArmor.go
```diff
@@ -461,15 +461,15 @@ func KubeArmor() {
 			}
 
 			// monitor containers
-			if strings.Contains(cfg.GlobalCfg.CRISocket, "docker") {
+			if strings.Contains(dm.Node.ContainerRuntimeVersion, "docker") {
```
nam-jaehyun, line 464: if CRI socket is not set, KubeArmor automatically detects CRI based on dm.Node.ContainerRuntimeVersion. do you think we still need to use dm.Node.ContainerRuntimeVersion even with the user-defined CRI socket?

nam-jaehyun (commented):
if CRI socket is not set, KubeArmor automatically detects CRI based on dm.Node.ContainerRuntimeVersion. do you think we still need to use dm.Node.ContainerRuntimeVersion even with the user-defined CRI socket?

PR #752, opened 2022-07-01
ginkgo based test framework

nyrahul:
* aim is to replace current test framework based on bash scripts which
  is difficult to maintain and extend.
* handling flakey tests is difficult
* ginkgo and gomega provides better control over executions and
  assertions
* it now directly uses k8s client go apis and kubearmor-client apis to
  handle k8s tasks and observing alerts/telemetry from kubearmor

Signed-off-by: Rahul Jadhav <nyrahul@gmail.com>


nyrahul:
Will first make the GH actions work in the local repo then raise a PR.

PR #760, opened 2022-07-06
ginkgo based test framework

nyrahul:
* aim is to replace current test framework based on bash scripts which
  is difficult to maintain and extend.
* handling flakey tests is difficult
* ginkgo and gomega provides better control over executions and
  assertions
* it now directly uses k8s client go apis and kubearmor-client apis to
  handle k8s tasks and observing alerts/telemetry from kubearmor

Signed-off-by: Rahul Jadhav <nyrahul@gmail.com>


tests/go.mod
```diff
@@ -0,0 +1,103 @@
+module github.com/kubearmor/karts
+
+go 1.17
```
daemon1024:
```suggestion
go 1.18
```

tests/util/karmorlog.go
```diff
@@ -0,0 +1,81 @@
...
+	for eventChan != nil {
+		select {
+		case evtin := <-eventChan:
+			switch evt := evtin.(type) {
+			case pb.Alert:
+				alerts = append(alerts, evt)
+			case pb.Log:
+				logs = append(logs, evt)
+			default:
+				log.Errorf("UNKNOWN EVT type %v", reflect.TypeOf(evt))
```
daemon1024: We changed the behaviour in https://github.com/kubearmor/kubearmor-client/pull/96, will need to update here too.

I am wondering if we should create a structure like
```go
struct {
	Data []byte  // byte data for log/alert
	Type string // "Alert"/"Log"
} 
```

Then we can directly marshal according to the type.

tests/util/kartutil.go
```diff
@@ -0,0 +1,273 @@
...
+		if flag {
+			return nil
+		}
+		if timeout == 0 {
+			break
+		}
+		time.Sleep(1 * time.Second)
+	}
+	return errors.New("condition not met")
+}
```
daemon1024: What do you think about using [context](https://pkg.go.dev/context) Library to provide timeout here?

For [ref](https://www.digitalocean.com/community/tutorials/how-to-use-contexts-in-go)

Implementation would something in the lines of

```go
func waitForCondition(ctx context.Context, cf ConditionFunc) error {
	for {
		select {
			case <-ctx.Done():
            			return ctx.Err()
            default:
            	flag, err := cf(ctx)
				if err != nil {
					return err
				}
				if flag {
					return nil
				}
				time.Sleep(1 * time.Second)
		}
	}
	return errors.New("condition not met")
}
```

And we instantiate the context in the main program.

This would preserve the control flow in the main program.

tests/util/kartutil.go
```diff
@@ -0,0 +1,273 @@
...
+*/
+
+// Kubectl execute
+func Kubectl(cmdstr string) (string, error) {
+	cmdf := strings.Fields(cmdstr)
+	cmd := exec.Command("kubectl", cmdf...)
+	sout, err := cmd.Output()
+	return string(sout), err
+}
```
daemon1024, line 259: Any particular reasoning to exec `kubectl` instead of directly using client-go 🤔

tests/smoke/smoke_test.go
```diff
@@ -0,0 +1,186 @@
...
+	pods := []string{}
+	for len(pods) != 1 && cnt < 30 {
+		pods, err = K8sGetPods(name, "wordpress-mysql", 20)
+		Expect(err).To(BeNil())
+		cnt++
+		time.Sleep(1 * time.Second)
+	}
+	Expect(len(pods)).To(Equal(1))
+	return pods[0]
+}
```
daemon1024: We can maybe avoid this long sleep check by checking the annotations 🤔 Like one of them would have the AppArmor annotation and other would not. And return the one with the apparmor annotation.

tests/util/kartutil.go
```diff
@@ -0,0 +1,273 @@
...
+		log.Error("failed to install kubearmor err=%s", err)
+		return err
+	}
+	return nil
+}
+
+func k8sUninstallKubearmor() {
+	err := kins.K8sUninstaller(k8sClient, getOptions())
+	if err != nil {
+		log.Error("failed to install kubearmor err=%s", err)
```
daemon1024:
```suggestion
		log.Error("failed to uninstall kubearmor err=%s", err)
```

tests/util/kartutil.go
```diff
@@ -0,0 +1,273 @@
...
+		return false
+	}
+	k8sClient = cli
+	return true
+}
+
+func getOptions() kins.Options {
+	return kins.Options{
+		"kube-system",
+		"kubearmor/kubearmor:stable",
```
daemon1024, line 44: We would need to use locally built test images to test things right 👀

daemon1024 (changes requested):
<details>
  <summary>Finally it's here 🚀 </summary>

  image
</details>

We would need to modify `ci-go.yaml` to run linter as part of `tests` too, currently it's limited to `KubeArmor` directory.


Shared some concerns and feedback inline 😁

tests/util/karmorlog.go
nyrahul: you mean send this structure from kubearmor-client itself?

tests/util/kartutil.go
nyrahul: Thanks. NEver used contexts before! What do you mean by "preserving the control flow"? How would that be different with contexts vs existing code? Both of them block on `waitForCondition`.

tests/util/kartutil.go
nyrahul: Just wanted to ensure for some reason if the client-go APIs are not ready then kubectl exec is available and the tester should not add exec themselves. currently apply/delete are kubectl execed because client-go code to handle them is complex and hence pending.

tests/util/kartutil.go
nyrahul: We are currently executing `make run` and using the test framework against that. But yes, we will need a way to test with local images.

tests/util/karmorlog.go
daemon1024: Yes!

tests/util/kartutil.go
daemon1024: It is possible that cf() never returns i.e. K8s API never returns for some reason. In that case, the current timeout wouldn't be effective and our program would still hang up.

With contexts, the parent context is passed on to functions and callback here, and canceling the execution of events due to timeout or some other reason is preserved in the main program where the context is initalised.

All the client-go APIs accept context and can have our timeouts passed through them.

tests/smoke/smoke_test.go
nyrahul: This is an excellent idea. Was able to remove all hardcoded sleeps with this change.

daemon1024 (approved):
🚀

Issue #221, opened 2021-07-23
Support OpenShift


nam-jaehyun:
KubeArmor will eventually need to support OpenShift too.

anutosh491:
Hello @nam-jaehyun , I recently got onboard with KubeArmor and started contributing to good first issues . Along with other interesting projects that KubeArmor has to offer for the LFX mentorship program this summer ..... I am also interested in this project . I am comfortable with Go and K8s as of now . Could you share any relevant links , resources which I could look up for this ? Also what would you suggest me to do to get more familiar with this project !

Ankurk99:
Done as a part of LFX mentorship

PR #761, opened 2022-07-07
Add crio volume mount configuration

vikasvr:
This Pull requests adds volume mounts for cri-o
Reference Issue: https://github.com/kubearmor/KubeArmor/issues/221 

Supporting Pull request to provide required env for daemon set [here](https://github.com/kubearmor/kubearmor-client/pull/98)

Signed-off-by: Vikas Verma <vikasvr23@gmail.com>

### DIFF
--- a/deployments/get/defaults.go
+++ b/deployments/get/defaults.go
@@ -154,6 +154,48 @@ var defaultConfigs = map[string]DaemonSetConfig{
 			},
 		},
 	},
+	"oke": {
+		Args: []string{
+			"-enableKubeArmorHostPolicy",
+		},
+		Envs: envVar,
+		VolumeMounts: []corev1.VolumeMount{
+			hostUsrVolMnt,
+			apparmorVolMnt,
+			{
+				Name:      "crio-sock-path", // crio socket
+				MountPath: "/var/run/crio/crio.sock",
+				ReadOnly:  true,
+			},
+			{
+				Name:      "crio-storage-path", // crio storage - stores all of its data, including containers images, in this directory.
+				MountPath: "/var/lib/containers/storage",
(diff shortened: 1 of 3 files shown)


nyrahul (approved):
LGTM

Ankurk99 (approved):

DelusionalOptimist (approved):

deployments/get/defaults.go
```diff
@@ -150,6 +150,48 @@ var defaultConfigs = map[string]DaemonSetConfig{
 			},
 		},
 	},
+	"cri-o": {
```
DelusionalOptimist: @Ankurk99 @nyrahul we should target `oke` here, right? So that deployment manifests for OpenShift Kubernetes can be generated? :thinking:

Ankurk99: >we should target oke here

@vikasvr I think it would make more sense. Please also update the main.go environments here: https://github.com/kubearmor/KubeArmor/blob/37002bf115c2a614fe8b643c8eb249ac9a98faf2/deployments/main.go#L23

vikasvr: @DelusionalOptimist  Since It's a map of container runtime and associated configuration, I added `cri-o`, is it ok to have provider specific info `oke`.

DelusionalOptimist (commented):
@vikasvr can you also generate and add the deployment yaml by running `make`? It should generate a new kubearmor.yaml in `deployments/OKE/`.
Also, please squash your commits and we should be good to go.

DelusionalOptimist (approved):
Great work! Thanks @vikasvr for testing KubeArmor on OKE! :raised_hands:

PR #774, opened 2022-07-18
get all required data for karmor probe when kubearmor is running

Essietom:
Signed-off-by: essietom <essietom@gmail.com>

### DIFF
--- a/KubeArmor/core/karmorprobedata.go
+++ b/KubeArmor/core/karmorprobedata.go
@@ -1,0 +1,92 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 Authors of KubeArmor
+
+package core
+
+import (
+	"encoding/json"
+	"log"
+	"os"
+	"path/filepath"
+
+	cfg "github.com/kubearmor/KubeArmor/KubeArmor/config"
+)
+
+type KarmorData struct {
+	OSImage             string
+	KernelVersion       string
+	KubeletVersion      string
+	ContainerRuntime    string
+	SupportedLSMs       string
+	KernelHeaderPresent bool
+	HostSecurity        bool
+	ContainerSecurity   bool
+	KubeArmorPosture    KubeArmorPostures
+}
+
+type KubeArmorPostures struct {
+	DefaultFilePosture             string
+	DefaultNetworkPosture          string
+	DefaultCapabilitiesPosture     string
+	HostDefaultFilePosture         string
+	HostDefaultNetworkPosture      string
+	HostDefaultCapabilitiesPosture string
+}
+
+func SetKarmorData(dm *KubeArmorDaemon) KarmorData {
+	var kd KarmorData
+	var postures KubeArmorPostures
+
+	lsm := []byte{}
+	lsmPath := "/sys/kernel/security/lsm"
+
+	if _, err := os.Stat(filepath.Clean(lsmPath)); err == nil {
+		lsm, err = os.ReadFile(lsmPath)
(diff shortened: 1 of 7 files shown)


KubeArmor/core/kubeArmor.go
```diff
@@ -635,3 +640,37 @@ func KubeArmor() {
 	// destroy the daemon
 	dm.DestroyKubeArmorDaemon()
 }
+
+// StrToFile Function
+func KarmorDataToFile(val KarmorData, destFile string) {
```
nyrahul: this function should go in `karmorprobedata.go` file.

KubeArmor/core/kubeArmor.go
```diff
@@ -635,3 +640,37 @@ func KubeArmor() {
...
+	j, err := json.Marshal(val)
+	if err != nil {
+		log.Println(err)
```
nyrahul: return in case of this failure.

KubeArmor/core/kubeArmor.go
```diff
@@ -635,3 +640,37 @@ func KubeArmor() {
...
+	}
+	if _, err := os.Stat(destFile); err != nil {
+		newFile, err := os.Create(filepath.Clean(destFile))
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Failed to create a file (%s, %s)\n", destFile, err.Error())
```
nyrahul: Why are we using `fmt` in someplaces and `log` in other cases?

KubeArmor/core/kubeArmor.go
```diff
@@ -635,3 +640,37 @@ func KubeArmor() {
...
+		}
+		if err := newFile.Close(); err != nil {
+			fmt.Fprintf(os.Stderr, "Failed to close the file (%s, %s)\n", destFile, err.Error())
+		}
+	}
+
+	// #nosec
+	file, err := os.OpenFile(destFile, os.O_RDWR, 0644)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to open a file (%s, %s)\n", destFile, err.Error())
```
nyrahul: return in case of failure.

KubeArmor/core/kubeArmor.go
nyrahul: I think this whole function can be reduced to couple of lines. Something like this:

```go
j, err := json.Marshal(val)
// failure handling
err := ioutil.WriteFile(destFile, j, 0644)
// failure handlling
```

Why use `os.Stat()`, `os.Create()`, `os.Close()` etc?

nyrahul (changes requested):

nyrahul:
As discussed on call, please handle:
- [ ] squash the commits
- [ ] resolve conflicts
- [ ] license headers .. change to 2022
- [ ] documentation change in kubearmor ([getting started guide](https://github.com/kubearmor/KubeArmor/blob/main/getting-started/deployment_guide.md)) to suggest using probe in case users want to understand compatibility.

KubeArmor/core/karmorprobedata.go
```diff
@@ -0,0 +1,93 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2021 Authors of KubeArmor
+
+package core
+
+import (
+	"encoding/json"
+	"io/ioutil"
```
daemon1024: Please refer https://go.dev/doc/go1.16#ioutil for switching ioutil to suggested APIs.

PR #791, opened 2022-07-31
fix CI workflow

nam-jaehyun:
- fix CI workflow (including tests with Docker)
- match golang version to 1.18
- update vagrant setup
  - Ubuntu 21.10 is no longer supported; thus, we use Ubuntu 22.04 as NETNEXT=1


.github/workflows/ci-test-systemd.yml
```diff
@@ -18,46 +18,59 @@ on:
...
+            for tool in "clang" "llc" "llvm-strip"; do
+              sudo rm -f /usr/bin/$tool
+              sudo ln -s /usr/bin/$tool-14 /usr/bin/$tool
+            done
+          fi
+
+      - name: Compile libbpf
+        run: |
+          cd KubeArmor/BPF
+          make -C libbpf/src
```
nyrahul, line 60: Hey @nam-jaehyun should we keep this in script and invoke the script from all the actions? The script can be kept in `.github/scripts` folder.

nyrahul (changes requested):
The GH action has substantial copy-pasta .. can we keep the llvm install handling as part of some script in .github folder and call it from the action?

daemon1024 (commented):
LGTM, just as @nyrahul mentioned, if we can leverage [composite actions](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) or scripts. 

We can keep composite actions as a seperate work item/issue as well.

nyrahul:
> LGTM, just as @nyrahul mentioned, if we can leverage [composite actions](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) or scripts.
> 
> We can keep composite actions as a seperate work item/issue as well.

I am going to just merge this and take up refactoring in separate PR.

nyrahul (approved):

PR #794, opened 2022-08-03
add KubeArmorController

### LINKED ISSUE #241: Merge kubearmor-host-policy-manager into kubearmor-policy-manager
we're currently using two managers for KubeArmorPolicy and KubeArmorHostPolicy.
It would be better if we have a single manager for both policies.

achrefbensaad:


deployments/controller/kubearmor-controller-manager.yaml
```diff
@@ -0,0 +1,1261 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.4.1
+  creationTimestamp: null
+  name: kubearmorhostpolicies.security.kubearmor.com
```
nam-jaehyun, line 7: kubearmorhostpolicies?

deployments/main.go
```diff
@@ -34,12 +33,8 @@ func main() {
 			dp.GetRelayService(namespace),
 			dp.GetRelayDeployment(namespace),
 			dp.GenerateDaemonSet(strings.ToLower(env), namespace),
-			dp.GetPolicyManagerService(namespace),
```
nam-jaehyun, line 37: no controller? only CRDs?

pkg/KubeArmorController/api/v1/kubearmorhostpolicy_types.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// KubeArmorHostPolicySpec defines the desired state of KubeArmorHostPolicy
+type KubeArmorHostPolicySpec struct {
+	NodeSelector NodeSelectorType `json:"nodeSelector"`
+
+	Process      ProcessType          `json:"process,omitempty"`
```
nam-jaehyun, line 14: hsp doesn't support the global allow policy, which means that hsp with allow action must have fromSource.
this is the big difference between hsp and ksp. did you consider this point?

nam-jaehyun (changes requested):
Please check if there are wrong names that you missed to change while merging three controllers together. Also, please check if you correctly bring the functions and intents for each controller to this unified controller. I can still see some incorrect points.

deployments/controller/kubearmor-controller-manager.yaml
achrefbensaad: thats the plural name of the CRD.

deployments/main.go
achrefbensaad: Yes as the controller will require certificates pair to operate and we cannot store  those on git.

However we can update the helm chart to generate them at the during the install process.

pkg/KubeArmorController/api/v1/kubearmorhostpolicy_types.go
achrefbensaad: I am not aware of this detail. However, the original host policy controller have the same structure. 
https://github.com/kubearmor/KubeArmor/blob/main/pkg/KubeArmorHostPolicy/api/security.kubearmor.com/v1/kubearmorhostpolicy_types.go#L240-L258

achrefbensaad:
> Please check if there are wrong names that you missed to change while merging three controllers together. Also, please check if you correctly bring the functions and intents for each controller to this unified controller. I can still see some incorrect points.

Can you please point me to the incorrect points, I ran the `make manifests` and the results are a carbone copy of the original.

nyrahul (approved):
LGTM

nyrahul:
Hey @nam-jaehyun , can you please revisit your comments and @achrefbensaad responses ... and provide a signal to go ahead/no-go with this?

PR #797, opened 2022-08-03
hotfix/disable-security-unlink-path-for-unsupported-platforms

achrefbensaad:


achrefbensaad:
~~We need to compile a new stable version of karmor also.~~ No longer needed

KubeArmor/build/compile.sh
```diff
@@ -18,10 +18,19 @@ fi
 cd /KubeArmor/BPF
 make clean
 
+touch ignore.lst
+if ! grep -q "^CONFIG_SECURITY_PATH=y$"  /boot/config-$(uname -r) ; then 
+    echo "Disabeling security_path_unlink ..."
+    sed -i "s#// security_path_unlink start#/*#" system_monitor.c
```
nyrahul: it would be better to compile a sample code in the init-container to check if this support is available, than to depend on /boot/config-KERNEL-VERSION file for two reasons:
1. we should not be mounting /boot in the kubearmor pod for security reasons. /boot has sensitive information such as kmem.
2. config file could be present in different paths depending on distro ([ref](https://github.com/nyrahul/linux-kernel-configs/blob/88082861ad2a9a0542aa80cc9f2e8d10fb0c009a/lk-config-get.sh#L64))

KubeArmor/monitor/systemMonitor.go
```diff
@@ -184,6 +186,37 @@ func NewSystemMonitor(node *tp.Node, logger *fd.Feeder, containers *map[string]t
 	return mon
 }
 
+func LoadIgnoreList(bpfPath string) ([]string, error) {
```
nyrahul: Lint fix needed

deployments/generic/kubearmor.yaml
```diff
@@ -190,6 +196,10 @@ spec:
           path: /etc/os-release
           type: File
         name: os-release-path
+      - hostPath:
+          path: /boot
+          type: DirectoryOrCreate
```
nyrahul:
```suggestion
          type: Directory
```
should not create the directory.

deployments/AKS/kubearmor.yaml
```diff
@@ -120,6 +120,9 @@ spec:
         - mountPath: /media/root/etc/os-release
           name: os-release-path
           readOnly: true
+        - mountPath: /boot
```
nyrahul: why do we need to mount `/boot`?

deployments/BottleRocket/kubearmor.yaml
```diff
@@ -191,6 +197,10 @@ spec:
           path: /etc/os-release
           type: File
         name: os-release-path
+      - hostPath:
+          path: /boot
+          type: DirectoryOrCreate
```
nyrahul:
```suggestion
          type: Directory
```

nyrahul (changes requested):
thanks for handling this at this speed. Please check the review comments.

KubeArmor/build/compile.sh
achrefbensaad: Done

KubeArmor/BPF/tests/checksyscall.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2021 Authors of KubeArmor
```
nyrahul:
```suggestion
# Copyright 2022 Authors of KubeArmor
```

KubeArmor/BPF/tests/Makefile
```diff
@@ -0,0 +1,104 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2021 Authors of KubeArmor
```
nyrahul:
```suggestion
# Copyright 2022 Authors of KubeArmor
```

KubeArmor/BPF/tests/Makefile
```diff
@@ -0,0 +1,104 @@
+# SPDX-License-Identifier: Apache-2.0
```
nyrahul, line 1: This makefile is duplicated. Can we do better here? May be we can leverage makefile's `include` construct.

nyrahul (changes requested):

KubeArmor/BPF/tests/Makefile
achrefbensaad: Done

nyrahul (approved):
LGTM 👍🏼

KubeArmor/BPF/tests/checksyscall.sh
```diff
@@ -0,0 +1,36 @@
...
+fi
+
+SYSCALL=${1#*/}
+
+if ! $SUDO syscheck $1 ; then
+    echo "Disabeling $SYSCALL ..."
+    sed -i "s#// $SYSCALL start#/*#" $BPF_HOME/system_monitor.c
+    sed -i "s#// $SYSCALL end#*/#"  $BPF_HOME/system_monitor.c
+    echo "$SYSCALL" >> ignore.lst
+fi
```
daemon1024: Can we use [if preprocessor](https://www.cs.auckland.ac.nz/references/unix/digital/AQTLTBTE/DOCU_078.HTM) for conditional compilation instead of brute removing contents?

KubeArmor/BPF/system_monitor.c
```diff
@@ -761,6 +761,7 @@ static __always_inline int events_perf_submit(struct pt_regs *ctx)
 
 // == Full Path == //
 
+// security_path_unlink start
```
daemon1024:
```suggestion
#if defined(SECURE_PATH_UNLINK)
```

KubeArmor/BPF/system_monitor.c
```diff
@@ -781,6 +782,7 @@ int kprobe__security_path_unlink(struct pt_regs *ctx){
 
     return 0;
 }
+// security_path_unlink end
```
daemon1024:
```suggestion
#endif
```

KubeArmor/BPF/tests/Makefile
```diff
@@ -0,0 +1,34 @@
...
+files = $(Cfiles:.c=)
+
+.PHONY: all
+all: compile-all check
+
+.PHONY: compile-all
+compile-all: checks/*.c
+	@for i in $(files) ; do \
+	echo "Compiling eBPF bytecode: $(GREEN)$$i.c$(NC) ..." && \
+	$(CL) $(KF) -DMONITOR_HOST_AND_CONTAINER -c $$i.c -o -| llc -march=bpf -mcpu=probe -filetype=obj -o $$i.bpf.o;\
```
daemon1024, line 23:
```suggestion
	$(CL) $(KF) -DMONITOR_HOST_AND_CONTAINER -DSECURE_PATH_UNLINK -c $$i.c -o -| llc -march=bpf -mcpu=probe -filetype=obj -o $$i.bpf.o;\
```

We add the flag to compilation if supported else not.

daemon1024 (commented):
Suggesting an alternate approach, what do you think?

KubeArmor/BPF/tests/checksyscall.sh
daemon1024: That's how things are in the Kernel as well so...
Ref https://elixir.bootlin.com/linux/latest/source/include/linux/security.h#L1769

KubeArmor/build/compile.sh
```diff
@@ -15,6 +15,10 @@ if [ -f "$FILE" ]; then
     fi
 fi
 
+cd /KubeArmor/BPF/tests
+make clean
+make
```
daemon1024: Can we have a quick check by deploying the init container on GKE?
Not sure if it would find linux/bpf.h without the right KRNDIR supplied.

achrefbensaad: @daemon1024 I updated the PR to pass down the correct value of KRNDIR.

nyrahul (approved):
LGTM

daemon1024 (approved):

KubeArmor/BPF/tests/checksyscall.sh
```diff
@@ -0,0 +1,37 @@
...
+
+if command -v sudo &> /dev/null
+then
+    SUDO="sudo"
+fi
+
+SYSCALL=${1#*/}
+
+if ! $SUDO ./syscheck $1 ; then
+    echo "Disabeling $SYSCALL ..."
```
Ankurk99: echo "Disabling $SYSCALL ..."

KubeArmor/BPF/tests/checksyscall.sh
```diff
@@ -0,0 +1,37 @@
...
+    echo "$SYSCALL" >> ignore.lst
+else
+    echo "Enabeling syscall $SYSCALL"
```
Ankurk99: echo "Enabling syscall $SYSCALL"

KubeArmor/BPF/tests/main.go
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 Authors of KubeArmor
+
+package main
+
```
Ankurk99, line 6: please address this go-lint error

Ankurk99 (changes requested):
All good, just some minor nits.

nam-jaehyun:
I feel this approach is a bit complicated, requiring many changes to only check security_path_unlink().
Is this the best way to check the support of security_path_unlink()?
what if we have some other hooks, which are not consistently supported in different OSs?
do we need more init-containers and the check code for each hook?

achrefbensaad:
> I feel this approach is a bit complicated, requiring many changes to only check security_path_unlink(). Is this the best way to check the support of security_path_unlink()? what if we have some other hooks, which are not consistently supported in different OSs? do we need more init-containers and the check code for each hook?

No, you will only need to create a .c under checks and contains the function in system_monitor.c with ifdef. 
All checks will be run on the same init container that compiles system_monitor.c to ebpf code.

nam-jaehyun (approved):
LGTM

PR #799, opened 2022-08-04
update SELinux enforcer

nam-jaehyun:
- fix the SELinux module to allow the KubeArmor Systemd daemon to enforce host policies.
- update the dependencies of the rpm package.


KubeArmor/.goreleaser.yaml
```diff
@@ -58,22 +62,18 @@ nfpms:
...
         dependencies:
-          - make
-          - libelf-dev
+          - clang
+          - llvm
           - linux-headers-generic
       rpm:
         dependencies:
-          - make
-          - elfutils-libelf-devel
```
daemon1024, line 76: Do we not need these?
Not sure but ig we need libelf for compiling the monitor code 🤔

nyrahul: @nam-jaehyun , can you respond to this comment? Lets get this one merged so that we can create a new systemd release. Thanks

nyrahul (approved):
LGTM

KubeArmor/.goreleaser.yaml
nam-jaehyun: When I tested the rpm package, KubeArmor worked without those dependencies. However, I also had the same question. We may see if those are required later.

daemon1024: It might be possible that these exist already on the system? (which is most likely the case). If the dependencies already exist it wouldn't reinstall so should be fine to add them 👀

nam-jaehyun: added make and libelf-dev (elfutils-libelf-devel)

PR #806, opened 2022-08-10
add syscalls

### LINKED ISSUE #824: add Syscall  based policy constructs
We want to our policies to alert for syscalls made on the system.

```
syscalls:
    matchSyscalls: # for all syscalls
    - syscall:
      - syscallX
      - syscallY
      fromSource:                          # --> optional
      - path: [absolute exectuable path]
      - dir: [directory] # recurive = true?
    matchPaths: # for the syscalls that contain file paths
    - path: [absolute file path]
      syscall:
      - syscallX
      - syscallY
      fromSource:                          # --> optional
      - path: [absolute exectuable path]
      - dir: [directory] # recurive = true?
```

achrefbensaad:
* ability to detect system calls that are security sensitive
* ability to provide a policy construct to check use of specific syscall in specific context ... (use-case: if setuid() is invoked in the container, send an alert with the contextual info).
* [Future] Ability to generate seccomp profile based on this


KubeArmor/types/types.go
```diff
@@ -424,6 +424,34 @@ type CapabilitiesType struct {
...
+	Syscalls   []string                `json:"syscall,omitempty"`
+	FromSource []SyscallFromSourceType `json:"fromSource,omitempty"`
+}
+
+type SyscallMatchDirectoriesType struct {
+	Dir        string                  `json:"dir,omitempty"`
+	Recursive  bool                    `json:"recursive,omitempty"`
+	Syscalls   []string                `json:"syscall,omitempty"`
+	FromSource []SyscallFromSourceType `json:"fromSource,omitempty"`
+}
```
nyrahul: Lets not have Paths and Directories separately, like we have for policy rules today.
If a path ends with `/` then it is a directory.

Ankurk99:
Can you please add an extra line at the end of all the yaml files? Also please address the go-lint warnings.

KubeArmor/types/types.go
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2021 Authors of KubeArmor
 
+//Types package is an extention of the crd types.
```
daemon1024:
```suggestion
// Package ypes is an extension of the CRD types.
```
Pleasing the linter :)

KubeArmor/types/types.go
```diff
@@ -424,6 +425,45 @@ type CapabilitiesType struct {
...
+
+	Severity int      `json:"severity,omitempty"`
+	Tags     []string `json:"tags,omitempty"`
+	Message  string   `json:"message,omitempty"`
+}
+
+// SyscallMatchPathType Structure
+type SyscallMatchPathType struct {
+	Path       string                  `json:"path,omitempty"`
+	Recursive  bool                    `json:"recursive,omitempty"`
```
daemon1024, line 448: We using Path as both path and directory now?

KubeArmor/monitor/syscallParser.go
```diff
@@ -1108,3 +1108,14 @@ func GetArgs(dataBuff *bytes.Buffer, Argnum int32) ([]interface{}, error) {
 
 	return args, nil
 }
+
+var auditedSyscalls = map[int]string{
+	87: "unlink",
+}
```
daemon1024, line 1114: We extend to add further syscalls here in a subsequent PR. But do we plan to Audit existing syscalls?

Like we classify sockert/connect/accept/bind as Network, but what if I want to audit a a particular syscall here. This is out of scope of the PR but wanted to understand future plans.

daemon1024 (approved):
LGTM 🚀  Some questions inline :)

KubeArmor/types/types.go
achrefbensaad: Yes, we merged matchpaths and matchdirectories in syscalls for now, we are going to merge them for all other types in the upcoming releases.

KubeArmor/monitor/syscallParser.go
achrefbensaad: Yes we are planing to, but we will not audit all system calls. We will audit only for a subset of them.

PR #807, opened 2022-08-10
added tag to hsp to make it cluster-scoped resource

rksharma95:
`KubeArmorHostPolicy` is not a namespace scoped resource, therefore marked the type `KubeArmorHostPolicy` with tag `+genclient:nonNamespaced`.

Managing `hsp` with client-go API was an issue earlier. This PR fixes that issue.

Signed-off-by: rk <ramakant@accuknox.com>

### DIFF
--- a/pkg/KubeArmorHostPolicy/Makefile
+++ b/pkg/KubeArmorHostPolicy/Makefile
@@ -51,7 +51,6 @@ delete: manifests
 # Generate manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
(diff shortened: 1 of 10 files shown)


nyrahul (approved):

achrefbensaad (approved):
LGTM

nam-jaehyun:
don't we need to also change Makefile?

nam-jaehyun:
also, you upgrade the version of kubebuilder for hsp.
then, it would be better to upgrade the version of kubebuilder for ksp too.

rksharma95:
> don't we need to also change Makefile?

@nam-jaehyun what are the changes to be expected in the Makefile?

nam-jaehyun:
> > don't we need to also change Makefile?
> 
> @nam-jaehyun what are the changes to be expected in the Makefile?

removing the out-dated line in Makefile

https://github.com/kubearmor/KubeArmor/blob/64635aca1105ab85ea63cd167e85cee21cbfcf20/pkg/KubeArmorHostPolicy/Makefile#L54

nam-jaehyun (changes requested):
Let's merge it after updating Makefile.

PR #817, opened 2022-08-15
Refactor BPFLSM Enforcement

### LINKED ISSUE #787: Lenient Whitelisting Policies
Let's consider a scenario,

We want to allow service account token access to only `/bin/cat`. But everything else should function as it is.
Can also be interpreted as 
Block access to service account token to everyone except for `/bin/cat`

A sample policy for this could be
```yaml
  file:
    matchDirectories:
    - dir: /run/secrets/kubernetes.io/serviceaccount/
      fromSource:
      - path: /bin/cat
      recursive: true
    - dir: /run/secrets/kubernetes.io/serviceaccount/
      recursive: true
      action: Block # Denied in Global 
    - dir: /
      recursive: true # This Allow rule is helping us introduce the leniency 
  process:
    matchDirectories:
    - dir: /
      recursive: true
  action:
    Allow
```

But in our current behavior we copy the deny rule to subprofiles, so this policy blocks service account token from `/bin/cat` as well.
We need a form of conflict resolution to not copy the deny rule to the subprofile here.

Consider another policy.

```yaml
  file:
    matchDirectories:
    - dir: /run/secrets/kubernetes.io/serviceaccount/
      fromSource:
      - path: /bin/cat
      recursive: true
    - dir: /run/secrets/kubernetes.io/serviceaccount/
      recursive: true
      action: Block # Denied in Global 
    - dir: /
      recursive: true
    matchPaths:
    - path: /run/secrets/kubernetes.io/serviceaccount/token
      fromSource:
      - path: /bin/more
    - path: /run/secrets/kubernetes.io/serviceaccount/token
      action: Block
  process:
    matchDirectories:
    - dir: /
      recursive: true
  action:
    Allow
```

We decide to take the superset rule take precedence.

Add a conflict resolution mechanism in KubeArmor to add the allow rule in case of matching allow and deny, 
Also in case a set of rules are conflicting in the sense that one is a parent to another, we add the rule which is the superset,
Like in this case
`/run/secrets/kubernetes.io/serviceaccount/**` is a superset to `/run/secrets/kubernetes.io/serviceaccount/token`. So we decide to choose the superset rule everytime.

**Action Item**

Implement the above described conflict resolution. 

// @nyrahul @nam-jaehyun

daemon1024:
Fixes #787


nyrahul (approved):
Awesome as usual. Great work with refactoring. LGTM. 👍🏼

PR #822, opened 2022-08-18
Write karmor probe data

Essietom:
Ref https://github.com/kubearmor/kubearmor-client/issues/19

Signed-off-by: Esther Adenekan <essietom@gmail.com>

### DIFF
--- a/KubeArmor/core/karmorprobedata.go
+++ b/KubeArmor/core/karmorprobedata.go
@@ -1,0 +1,61 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 Authors of KubeArmor
+
+package core
+
+import (
+	kl "github.com/kubearmor/KubeArmor/KubeArmor/common"
+	cfg "github.com/kubearmor/KubeArmor/KubeArmor/config"
+	tp "github.com/kubearmor/KubeArmor/KubeArmor/types"
+)
+
+type KarmorData struct {
+	OSImage                 string
+	KernelVersion           string
+	KubeletVersion          string
+	ContainerRuntime        string
+	ActiveLSM               string
+	KernelHeaderPresent     bool
+	HostSecurity            bool
+	ContainerSecurity       bool
+	ContainerDefaultPosture tp.DefaultPosture
+	HostDefaultPosture      tp.DefaultPosture
+}
+
+// SetKarmorData generates runtime configuration for KubeArmor to be consumed by kArmor
+func (dm *KubeArmorDaemon) SetKarmorData() {
+	var kd KarmorData
+
+	kd.ContainerDefaultPosture = tp.DefaultPosture{
+		FileAction:         cfg.GlobalCfg.DefaultFilePosture,
+		NetworkAction:      cfg.GlobalCfg.DefaultNetworkPosture,
+		CapabilitiesAction: cfg.GlobalCfg.DefaultCapabilitiesPosture,
+	}
+	kd.HostDefaultPosture = tp.DefaultPosture{
+		FileAction:         cfg.GlobalCfg.HostDefaultFilePosture,
+		NetworkAction:      cfg.GlobalCfg.HostDefaultNetworkPosture,
+		CapabilitiesAction: cfg.GlobalCfg.HostDefaultCapabilitiesPosture,
+	}
+
+	kd.OSImage = dm.Node.OSImage
+	kd.ContainerRuntime = dm.Node.ContainerRuntimeVersion
+	kd.KernelVersion = dm.Node.KernelVersion
+	kd.KubeletVersion = dm.Node.KubeletVersion
+	kd.ContainerRuntime = dm.Node.ContainerRuntimeVersion
+	if dm.RuntimeEnforcer != nil {
+		kd.ActiveLSM = dm.RuntimeEnforcer.EnforcerType
+
+		if cfg.GlobalCfg.Policy {
+			kd.ContainerSecurity = true
+		}
+		if cfg.GlobalCfg.HostPolicy {
+			kd.HostSecurity = true
+		}
+	}
+	kd.KernelHeaderPresent = true //this is always true since KubeArmor is running
+	err := kl.WriteToFile(kd, "/tmp/karmorProbeData.cfg")
+	if err != nil {
+		dm.Logger.Errf("Error writing karmor config data", err)
+	}
+
+}
--- a/KubeArmor/common/common.go
+++ b/KubeArmor/common/common.go
@@ -430,3 +430,16 @@ func MatchIdentities(identities []string, superIdentities []string) bool {
 	// otherwise, return true
 	return matched
 }
+
+// WriteToFile writes given string to file as JSON
+func WriteToFile(val interface{}, destFile string) error {
+	j, err := json.Marshal(val)
+	if err != nil {
+		return err
+	}
+	err = os.WriteFile(destFile, j, 0600)
+	if err != nil {
+		return err
+	}
+	return nil
+}
--- a/KubeArmor/core/kubeArmor.go
+++ b/KubeArmor/core/kubeArmor.go
@@ -631,7 +631,7 @@ func KubeArmor() {
 	dm.Logger.Print("Started to serve gRPC-based log feeds")
 
 	// == //
-
+	go dm.SetKarmorData()
 	dm.Logger.Print("Initialized KubeArmor")
 
 	// == //


daemon1024:
Ref https://github.com/kubearmor/kubearmor-client/issues/19

KubeArmor/core/karmorprobedata.go
```diff
@@ -0,0 +1,92 @@
...
+	lsms := string(lsm)
+
+	if cfg.GlobalCfg.Policy && !cfg.GlobalCfg.HostPolicy {
+		kd.ContainerSecurity = true
+	} else if !cfg.GlobalCfg.Policy && cfg.GlobalCfg.HostPolicy {
+		kd.HostSecurity = true
+	} else if cfg.GlobalCfg.Policy && cfg.GlobalCfg.HostPolicy {
+		kd.ContainerSecurity = true
+		kd.HostSecurity = true
+	}
```
nyrahul: I think we should not have such logic here.
We have plans to keep the config in the configmap anyways. 
I think we can remove this code.
```suggestion
```

nyrahul: @daemon1024 , WDYT?

nyrahul (dismissed):

KubeArmor/core/karmorprobedata.go
daemon1024: All the KubeArmor daemons in a cluster would have the same config map right?

In case of hybrid node setup, it is possible that one of daemons is working and one may not be. How will we differentiate the status in that case?

nyrahul: If the daemon is not working then we should not read the tmp file anyways. The configmap would look same. 

Are you implying that different daemons might have different runtime behaviour. And in that case we show the actual runtime behaviour in the tmp file. Is that what is happening in that code snippet?

daemon1024: > Are you implying that different daemons might have different runtime behaviour. And in that case we show the actual runtime behaviour in the tmp file.

Right.

> Is that what is happening in that code snippet?

Not sure, like we are printing all the supported LSM instead of active LSM. So needs some updates..

Essietom: i pushed the update for this, kindly review

KubeArmor/core/karmorprobedata.go
```diff
@@ -0,0 +1,107 @@
...
+	err = os.WriteFile(destFile, j, 0600)
+	if err != nil {
+		log.Println(err)
+		return
+	}
+
+}
+
+// detect the enforcer on the node
+func getActiveLSM() string {
```
daemon1024: We should be extracting the information from `re.Enforcer` and not make decisions in the function. It is possible that BPFLSM fails to initalise and fallsback to AppArmor, or it may be that the logic might change.

daemon1024 (changes requested):

KubeArmor/core/karmorprobedata.go
```diff
@@ -0,0 +1,81 @@
...
+	var postures KubeArmorPostures
+
+	if cfg.GlobalCfg.Policy && !cfg.GlobalCfg.HostPolicy {
+		kd.ContainerSecurity = true
+	} else if !cfg.GlobalCfg.Policy && cfg.GlobalCfg.HostPolicy {
+		kd.HostSecurity = true
+	} else if cfg.GlobalCfg.Policy && cfg.GlobalCfg.HostPolicy {
+		kd.ContainerSecurity = true
+		kd.HostSecurity = true
+	}
```
daemon1024:
```suggestion
	if cfg.GlobalCfg.Policy {
		kd.ContainerSecurity = true
	}
	if cfg.GlobalCfg.HostPolicy {
		kd.HostSecurity = true
	}
```

We can simplify the logic here!

Also let's move this logic inside if RuntimeEnforcer != nil. If there were trouble initalising the enforcer then no enforcement is involved.

daemon1024 (changes requested):

KubeArmor/core/karmorprobedata.go
```diff
@@ -0,0 +1,79 @@
...
+type KarmorData struct {
+	OSImage             string
+	KernelVersion       string
+	KubeletVersion      string
+	ContainerRuntime    string
+	ActiveLSM           string
+	KernelHeaderPresent bool
+	HostSecurity        bool
+	ContainerSecurity   bool
+	KubeArmorPosture    KubeArmorPostures
```
daemon1024:
```suggestion
	ContainerDefaultPosture    tp.DefaultPosture
	HostDefaultPosture    tp.DefaultPosture
```
Ref https://pkg.go.dev/github.com/kubearmor/KubeArmor/KubeArmor/types#DefaultPosture

KubeArmor/core/karmorprobedata.go
```diff
@@ -0,0 +1,79 @@
...
+func SetKarmorData(dm *KubeArmorDaemon) KarmorData {
+	var kd KarmorData
+	var postures KubeArmorPostures
+
+	postures.DefaultFilePosture = cfg.GlobalCfg.DefaultFilePosture
+	postures.DefaultNetworkPosture = cfg.GlobalCfg.DefaultNetworkPosture
+	postures.DefaultCapabilitiesPosture = cfg.GlobalCfg.DefaultCapabilitiesPosture
+	postures.HostDefaultFilePosture = cfg.GlobalCfg.HostDefaultFilePosture
+	postures.HostDefaultNetworkPosture = cfg.GlobalCfg.HostDefaultNetworkPosture
+	postures.HostDefaultCapabilitiesPosture = cfg.GlobalCfg.HostDefaultCapabilitiesPosture
```
daemon1024:
```suggestion
	kd.ContainerDefaultPosture = tp.DefaultPosture{
		FileAction:         cfg.GlobalCfg.DefaultFilePosture,
		NetworkAction:      cfg.GlobalCfg.DefaultNetworkPosture,
		CapabilitiesAction: cfg.GlobalCfg.DefaultCapabilitiesPosture,
	}
	kd.HostDefaultPosture = tp.DefaultPosture{
		FileAction:         cfg.GlobalCfg.HostDefaultFilePosture,
		NetworkAction:      cfg.GlobalCfg.HostDefaultNetworkPosture,
		CapabilitiesAction: cfg.GlobalCfg.HostDefaultCapabilitiesPosture,
	}
```

KubeArmor/core/karmorprobedata.go
```diff
@@ -0,0 +1,79 @@
...
+			kd.HostSecurity = true
+		}
+	}
+	kd.KernelHeaderPresent = true //this is always true since kube armor is running
+	kd.KubeArmorPosture = postures
+	return kd
+}
+
+// StrToFile Function
+func KarmorDataToFile(val KarmorData, destFile string) {
```
daemon1024:
```suggestion
// WriteToFile writes given string to file as JSON
func WriteToFile(val interface{}, destFile string) error {
```

We should move this to `common` directory where other helper functions reside.

KubeArmor/core/karmorprobedata.go
```diff
@@ -0,0 +1,79 @@
...
+type KubeArmorPostures struct {
+	DefaultFilePosture             string
+	DefaultNetworkPosture          string
+	DefaultCapabilitiesPosture     string
+	HostDefaultFilePosture         string
+	HostDefaultNetworkPosture      string
+	HostDefaultCapabilitiesPosture string
+}
+
+func SetKarmorData(dm *KubeArmorDaemon) KarmorData {
```
daemon1024:
```suggestion
func (dm *KubeArmorDaemon) SetKarmorData {
```

Let's add this as a struct method and not pass the address to the daemon structure here.

KubeArmor/core/karmorprobedata.go
```diff
@@ -0,0 +1,79 @@
...
+		if cfg.GlobalCfg.Policy {
+			kd.ContainerSecurity = true
+		}
+		if cfg.GlobalCfg.HostPolicy {
+			kd.HostSecurity = true
+		}
+	}
+	kd.KernelHeaderPresent = true //this is always true since kube armor is running
+	kd.KubeArmorPosture = postures
+	return kd
```
daemon1024:
```suggestion
	err := WriteToFile(kd,"/tmp/karmorProbeData.cfg")
	dm.Logger.Errf("Error writing karmor config data",err)
```
I think it should be fine to write the config in this function itself and eliminate the need to return the data, since we are not using it again anywhere else.

daemon1024 (changes requested):
Some refactoring suggestions, Lemme know WDYT bout em.

KubeArmor/core/karmorprobedata.go
```diff
@@ -0,0 +1,61 @@
...
+	ActiveLSM               string
+	KernelHeaderPresent     bool
+	HostSecurity            bool
+	ContainerSecurity       bool
+	ContainerDefaultPosture tp.DefaultPosture
+	HostDefaultPosture      tp.DefaultPosture
+}
+
+// func SetKarmorData(dm *KubeArmorDaemon) KarmorData {
+func (dm *KubeArmorDaemon) SetKarmorData() {
```
daemon1024:
```suggestion
// SetKarmorData generates runtime configuration for KubeArmor to be consumed by kArmor
func (dm *KubeArmorDaemon) SetKarmorData() {
```

Please add relevant comment

daemon1024 (approved):
Let's squash the commits

Ankurk99 (changes requested):
LGTM

KubeArmor/core/karmorprobedata.go
```diff
@@ -0,0 +1,61 @@
...
+		kd.ActiveLSM = dm.RuntimeEnforcer.EnforcerType
+
+		if cfg.GlobalCfg.Policy {
+			kd.ContainerSecurity = true
+		}
+		if cfg.GlobalCfg.HostPolicy {
+			kd.HostSecurity = true
+		}
+	}
+	kd.KernelHeaderPresent = true //this is always true since kube armor is running
```
Ankurk99: //this is always true since KubeArmor is running

Ankurk99 (approved):

PR #827, opened 2022-08-18
Supporting Rancher RKE

### LINKED ISSUE #771: supporting Rancher RKE
## Feature Request

Kubearmor has not be tested with [Rancher RKE](https://rancher.com/products/rke). The task involves setting up the sample cluster, verifying kubearmor operations on the cluster and making any fixes if necessary.

## Tasks
- [ ] setting up a sample cluster using Rancher RKE
- [ ] installing kubearmor on the sample cluster
- [ ] execute the automated tests on the sample cluster
- [ ] if kubearmor does not work, diagnose the issue
- [ ] discuss with the community and fix the issue
- [ ] update [kubearmor support matrix](https://github.com/kubearmor/KubeArmor/blob/main/getting-started/support_matrix.md)

PrimalPimmy:
Fixes #771 .

Not filling up the support matrix till all tests are passed.

### DIFF
--- a/deployments/RKE/kubearmor.yaml
+++ b/deployments/RKE/kubearmor.yaml
@@ -1,0 +1,1186 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: kubearmor
+  namespace: kube-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: kubearmor
+  namespace: kube-system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
+subjects:
+- kind: ServiceAccount
+  name: kubearmor
+  namespace: kube-system
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: kubearmor
+  namespace: kube-system
+spec:
+  ports:
+  - port: 32767
+    protocol: TCP
+    targetPort: 32767
+  selector:
+    kubearmor-app: kubearmor-relay
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    kubearmor-app: kubearmor-relay
+  name: kubearmor-relay
+  namespace: kube-system
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      kubearmor-app: kubearmor-relay
+  template:
+    metadata:
+      annotations:
+        kubearmor-policy: audited
+      labels:
+        kubearmor-app: kubearmor-relay
+    spec:
+      containers:
+      - image: kubearmor/kubearmor-relay-server:latest
+        name: kubearmor-relay-server
+        ports:
+        - containerPort: 32767
+      nodeSelector:
+        kubernetes.io/os: linux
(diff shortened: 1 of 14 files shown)


deployments/RKE/kubearmor.yaml
```diff
@@ -0,0 +1,1186 @@
...
+          type: Directory
+        name: sys-kernel-debug-path
+      - hostPath:
+          path: /etc/os-release
+          type: File
+        name: os-release-path
+      - hostPath:
+          path: /usr/src
+          type: Directory
+        name: usr-src-path
```
nyrahul, line 192: I thought the whole problem with supporting RKE was to use `DirectoryOrCreate` here. Isn't it?
```suggestion
      - hostPath:
          path: /usr/src
          type: DirectoryOrCreate
        name: usr-src-path
```

nyrahul (changes requested):
Hey @PrimalPimmy , can you test whether the manifest generated in this PR can be used to install on RKE? Based on what I see, it won't work.

deployments/RKE/kubearmor.yaml
PrimalPimmy: I thought it only applied to /lib/modules, and not /usr/src?

PrimalPimmy: we only changed it to DirectoryOrCreate only for /lib/modules path as far as I remember. rest were working fine.

PrimalPimmy:
I have confirmed that RKE works with this configuration. Anyone else can test it if you want.

achrefbensaad:
@nyrahul @PrimalPimmy Ill re-test this PR.

achrefbensaad (changes requested):
This manifest does not work on rke.
I tried installing using kamor install (it worked)
I tried using the generic manifest and it worked also.

image

PrimalPimmy:
@achrefbensaad I'm looking into it.. btw, this manifest is generated when we do karmor install --save, right?

achrefbensaad:
No, We generate it by running, `make gen` under `deployments`

PrimalPimmy:
I did that after editing main.go to add RKE. Perhaps because I didn't make changes to defaults,go, is that why it's not working?

achrefbensaad:
are there any changes that require updating `default.go` ?

PrimalPimmy:
I changed 
image

from Directory to DirectoryOrCreate. 

But now for some reason, it even works without this change. Before it would get stuck on init if this change is not done

daemon1024:
> But now for some reason, it even works without this change.

Because your first installation created the Directory ( due to OrCreate ), so for subsequent installs it's available.

daemon1024:
If we are needing DirectoryOrCreate, that means it's not serving any purpose right? It should fine to just not mount it then? Can we try that?

PrimalPimmy:
hey @achrefbensaad , can you test this in a new rke environment? did the changes, should work now hopefully

nyrahul:
> hey @achrefbensaad , can you test this in a new rke environment? did the changes, should work now hopefully

Thanks @achrefbensaad for supporting this ticket with such rigor.

achrefbensaad:
We opted for directory or create for all plateforms. https://github.com/kubearmor/KubeArmor/pull/872

PR #778, opened 2022-07-25
Added k8s test scenarios to ginkgo-based test framework

rksharma95:
* Added test scenarios to test `Kubearmor` in k8s env by applying `KubeArmorSecurityPolicy` for process, files and network operations.
* Added utility function `func K8sApplyFile(fileName string) error{}`
   supported k8s resources can be created with the YAML file if doesn't already exist.
   currently supported k8s resources:
   - Deployment
   - Service
   - Namespace
   - KubeArmorPolicy
   - KubeArmorHostPolicy   

- [x] validate if all the `local.sh` scenarios are covered.

Signed-off-by: rk <ramakant@accuknox.com>

PR #832, opened 2022-08-18
Use configmap for configuration handling 

### LINKED ISSUE #754: using configmap and grpc for configuration handling
## Feature Request

**Short Description**

Kubearmor is provided with the [configuration at invoke time](https://github.com/kubearmor/KubeArmor/blob/8e74a245981a1ad53f4bba014705a331f578efe4/deployments/generic/kubearmor.yaml#L82). This means changing the default configuration is not possible without updating the container image.
Most of [the configuration options](https://github.com/kubearmor/KubeArmor/blob/8e74a245981a1ad53f4bba014705a331f578efe4/KubeArmor/config/config.go#L18) are currently using default values but some may need to be changed depending upon deployment basis.

Similarly in systemd mode, the only way for the user to change the configuration is to edit `/opt/kubearmor/kubearmor.yaml` file.

A lot of times it is required that the `defaultPosture` of Kubearmor be changed between `Audit/Deny`. Currently, it is not possible to change it in the context of k8s deployment unless the user changes the deployment yamls manually.

Use grpc endpoint to save/update the configuration setting.
Use k8s `configmap` for saving the configuration. Observe changes to configmap and invoke the corresponding grpc handler.

For e.g., `watchConfigMap() -> updateDefaultPosture() -> grpcUpdateDefaultPosture() -> updateDefaultPosture()`

s1ntaxe770r:
**Purpose of PR?**:
handled `defaultPosture` configuration changes dynamically using k8s configmap. An informer instance is added to watch for configmap changes. 
It also bring changes to defaultPosture map updation logic. Now map will contain `defaultPosture` info for only annotated namespaces.  

Fixes #754

No

**Checklist:**
- [x] Bug fix. Fixes #754


nyrahul:
Hey @daemon1024 , Can you please review this PR? Can you check if it maps to all the configuration params we have and the implementation is as expected? Thanks

nyrahul:
@s1ntaxe770r , please fix the CI failure.

KubeArmor/core/kubeArmor.go
```diff
@@ -611,6 +612,11 @@ func KubeArmor() {
 
 	policyService := &policy.ServiceServer{}
 
+	postureService := &posture.ServiceServer{}
+	postureService.UpdateDefaultPosture = dm.UpdateDefaultPosture
```
daemon1024: dm.UpdateDefaultPosture is for namespace default posture and not the Global Default Posture. 

Namespace Default Posture depends on namespace annotations. But Global Default Posture is the default value for unannotated namespaces.

Currently we use cfg.GlobalCfg.DefaultPosture* as default values as can be seen at
https://github.com/kubearmor/KubeArmor/blob/7ff5ebdb23176cda21762f8e409f3f141090155e/KubeArmor/core/kubeUpdate.go#L387-L390

So ideally, the process to update global default posture would be
- Updating the global cfg,
- Loop through the default posture map and updating the respective defaults of unannotated pods (this might need an improvement of logic since updating individuals defaults might not be feasible)

daemon1024 (dismissed):
Implementation for Watching Config changes and gRPC Service LGTM.

Find comment about DefaultPosture Inline, needs updates.

Regarding Watching Config changes, it shouldn't be just limited to Default Posture,

We should include other configurations too like visibility?

KubeArmor/core/kubeArmor.go
s1ntaxe770r: Thanks @daemon1024 would make changes

s1ntaxe770r:
Hey @daemon1024 updated the PR however I am unsure how to proceed here:
> Loop through the default posture map and updating the respective defaults of unannotated pods (this might need an improvement of logic since updating individuals defaults might not be feasible)

daemon1024:
Let me frame the problem statement, You already have a whitelist policy enforced, What happens when you update the Default Posture using Config Map today. Does the block posture automatically update to audit? or vice versa?

You can reproduce this using https://github.com/kubearmor/KubeArmor/blob/main/getting-started/default_posture.md

s1ntaxe770r:
@daemon1024  how's this looking now?

KubeArmor/core/kubeUpdate.go
```diff
@@ -384,10 +384,12 @@ func (dm *KubeArmorDaemon) UpdateEndPointWithPod(action string, pod tp.K8sPod) {
 		if val, ok := dm.DefaultPostures[newEndPoint.NamespaceName]; ok {
 			newEndPoint.DefaultPosture = val
 		} else {
+			// fetch default posture from k8s configmap
+			cm := dm.GetConfigMap()
 			globalDefaultPosture := tp.DefaultPosture{
```
daemon1024: The problem remains the same imo.

Consider this.

1. DefaultPosture = Block (In config Map)
2. We set dm.DefaultPosture[samplens] = Block
3. Update DefaultPosture = Audit
4. We read dm.DefaultPosture[samplens] and it already has a value Block, we would never update the value since it already exists.

daemon1024: I would suggest you to create a testcase about this process and validate things with that testcase once when you make updates :)

s1ntaxe770r:
@daemon1024  could you test this out? It works in my testing but pod updates takes more than a few seconds

KubeArmor/core/kubeUpdate.go
```diff
@@ -2375,3 +2390,60 @@ func (dm *KubeArmorDaemon) WatchDefaultPosture() {
...
+					CapabilitiesAction: validateDefaultPosture("kubearmor-capabilities-posture", ns, cm.Data[cfg.ConfigDefaultCapabilitiesPosture]),
+				}
+				cfg.GlobalCfg.HostVisibility = cm.Data[cfg.ConfigVisibility]
+				dm.UpdateGlobalPosture(posture)
+				// update default posture for endpoints
+
+				for _, endpoint := range dm.EndPoints {
+					ns := endpoint.NamespaceName
+					dm.Logger.Printf("updating default posture for %s in %s", ns, endpoint.EndPointName)
+					dm.UpdateDefaultPosture(string(resp.Type), ns, posture)
```
daemon1024: If we have a namespace annotated, This would override that annotation tho?

Suppose we have namespace  "test" annotated with default posture block, and then we update config map to have audit posture. Namespace test should still be blocking.

daemon1024:
Can we add the testing you are doing  as testcase to our suite as well? This would be an important change I believe we should support it with test if possible.

KubeArmor/core/kubeUpdate.go
s1ntaxe770r: oof , how would I update unannotated pods ?

s1ntaxe770r: 🤔  I could loop through `dm.k8spods` and check annotations ?

daemon1024: Ig maybe we should try some major refactor here? Like we don't fill up the posture map struct if they are not annotated. And if no annotation is found we check the global defaults.

If this is not feasible then yes Looping with specific conditions should do.

rksharma95:
`ci-test-runtime` passing locally :thinking: 
```[INFO] Deleting github
namespace "github" deleted
deployment.apps "ubuntu-1-deployment" deleted
[DBG] Deleted github
[DBG] Finished Container Scenarios
[WARN] Skipped Host Scenarios
[INFO] Stopping KubeArmor
[DBG] Stopped KubeArmor
[PASS] Successfully tested KubeArmor```

File: deployments/BottleRocket/kubearmor.yaml [hunk 4]
```diff
@@ -389,6 +389,19 @@ spec:
...
+data:
+  cluster: default
+  defaultFilePosture: block
+  gRPC: "3276"
+  hostDefaultCapabilitiesPosture: block
+  hostDefaultNetworkPosture: block
+  visibility: process,file,network,capabilities
+kind: ConfigMap
+metadata:
+  name: kuberamor-config
```
nyrahul:
```suggestion
  name: kubearmor-config
```

deployments/get/defaults.go
```diff
@@ -25,6 +25,8 @@ var (
 	KubeArmorControllerServiceName      = "kubearmor-controller-metrics-service"
 	KubeArmorControllerDeploymentName   = "kubearmor-controller"
 	KubeArmorControllerSecretName       = "kubearmor-webhook-server-cert"
+	AnnotationsControllerSecretName     = "kubearmor-webhook-server-cert"
+	KubeArmorConfigMapName              = "kuberamor-config"
```
nyrahul:
```suggestion
	KubeArmorConfigMapName              = "kubearmor-config"
```

KubeArmor/core/kubeUpdate.go
```diff
@@ -2382,12 +2393,88 @@ func (dm *KubeArmorDaemon) removeBackUpPolicy(name string) {
...
+
+	for idx, endpoint := range dm.EndPoints {
+		ns, err := K8s.K8sClient.CoreV1().Namespaces().Get(context.Background(), endpoint.NamespaceName, metav1.GetOptions{})
+		if err != nil {
+			kg.Err("unable to fetch requested namespace")
+		}
+		fp, fa := validateDefaultPosture("kubearmor-file-posture", ns, cm.Data[cfg.ConfigDefaultFilePosture])
+		np, na := validateDefaultPosture("kubearmor-network-posture", ns, cm.Data[cfg.ConfigDefaultNetworkPosture])
+		cp, ca := validateDefaultPosture("kubearmor-capabilities-posture", ns, cm.Data[cfg.ConfigDefaultCapabilitiesPosture])
+		annotated := fa || na || ca      // if namespace is annotated for atleast one posture
```
daemon1024: This would result in API calls for the number of endpoints in a particular namespace even though it wouldn't change after 1 call. Not sure if it's the right approach here.

KubeArmor/core/kubeUpdate.go
```diff
@@ -2382,12 +2393,88 @@ func (dm *KubeArmorDaemon) removeBackUpPolicy(name string) {
...
+			if endPoint.PolicyEnabled == tp.KubeArmorPolicyEnabled {
+				// enforce security policies
+				dm.RuntimeEnforcer.UpdateSecurityPolicies(*endPoint)
+			}
+		}
+	}
+
+}
+
+func validateDefaultPosture(key string, ns *corev1.Namespace, defaultPosture string) (string, bool) {
```
daemon1024, line 2483: Let's document what's the return value intend to do.

KubeArmor/core/kubeUpdate.go
```diff
@@ -2438,6 +2531,34 @@ func (dm *KubeArmorDaemon) UpdateDefaultPosture(action string, namespace string,
...
+	dm.DefaultPosturesLock.Lock()
+	defer dm.DefaultPosturesLock.Unlock()
+
+	cfg.GlobalCfg.DefaultFilePosture = validateGlobalDefaultPosture(posture.FileAction)
+	cfg.GlobalCfg.DefaultNetworkPosture = validateGlobalDefaultPosture(posture.NetworkAction)
+	cfg.GlobalCfg.DefaultCapabilitiesPosture = validateGlobalDefaultPosture(posture.CapabilitiesAction)
+
+	dm.Logger.Printf("updated gloabal defaultFilePoster as %v", cfg.GlobalCfg.DefaultFilePosture)
+	dm.Logger.Printf("updated gloabal defaultCapabilitiesPoster as %v", cfg.GlobalCfg.DefaultCapabilitiesPosture)
+	dm.Logger.Printf("updated gloabal defaultNetworkPoster as %v", cfg.GlobalCfg.DefaultNetworkPosture)
```
daemon1024:
```suggestion
	dm.Logger.Printf("updated global defaultFilePoster as %v", cfg.GlobalCfg.DefaultFilePosture)
	dm.Logger.Printf("updated global defaultCapabilitiesPoster as %v", cfg.GlobalCfg.DefaultCapabilitiesPosture)
	dm.Logger.Printf("updated global defaultNetworkPoster as %v", cfg.GlobalCfg.DefaultNetworkPosture)
```

daemon1024 (commented):
Hey @rksharma95, Can you create a high level flow diagram to showcase how and when what gets updated w.r.t to Default Posture.

It is quite confusing I couldn't understand the flow from code. We watch for config map, we watch for annotations.... A flow diagram would help us be in sync here. Thanks.

KubeArmor/core/kubeUpdate.go
rksharma95: yes it can be optimized to one API call, i will make the changes.

rksharma95:
@daemon1024 here's an example flow, let me know if anything need to be more detailed. thanks
flow(1)

KubeArmor/core/kubeUpdate.go
```diff
@@ -2382,12 +2393,103 @@ func (dm *KubeArmorDaemon) removeBackUpPolicy(name string) {
...
+func (dm *KubeArmorDaemon) updatEndpointsWithCM(cm *corev1.ConfigMap, action string) {
+	dm.EndPointsLock.Lock()
+	defer dm.EndPointsLock.Unlock()
+
+	dm.DefaultPosturesLock.Lock()
+	defer dm.DefaultPosturesLock.Unlock()
+
+	// get all namespaces
+	nsList, err := K8s.K8sClient.CoreV1().Namespaces().List(context.Background(), metav1.ListOptions{})
+	if err != nil {
```
achrefbensaad, line 2418: we need to exit if err != nil as nsList will be nil.
This will cause a sigfault

KubeArmor/core/kubeUpdate.go
```diff
@@ -2382,12 +2393,103 @@ func (dm *KubeArmorDaemon) removeBackUpPolicy(name string) {
...
+		kg.Err("unable to fetch requested namespace")
+	}
+
+	// for each namespace if needed change endpoint depfault posture
+	for _, ns := range nsList.Items {
+		ns := ns
```
achrefbensaad, line 2425: Why this line is needed ?

KubeArmor/core/kubeUpdate.go
```diff
@@ -2382,12 +2393,103 @@ func (dm *KubeArmorDaemon) removeBackUpPolicy(name string) {
...
+		np, na := validateDefaultPosture("kubearmor-network-posture", &ns, cm.Data[cfg.ConfigDefaultNetworkPosture])
+		cp, ca := validateDefaultPosture("kubearmor-capabilities-posture", &ns, cm.Data[cfg.ConfigDefaultCapabilitiesPosture])
+		annotated := fa || na || ca      // if namespace is annotated for atleast one posture
+		fullyannotated := fa && na && ca // if namespace is fully annotated
+		posture := tp.DefaultPosture{
+			FileAction:         fp,
+			NetworkAction:      np,
+			CapabilitiesAction: cp,
+		}
+		dm.Logger.Printf("namespace %v is fullyannotated %v", ns.Name, fullyannotated)
```
achrefbensaad: This should be in an if statement, as at this point there is no check if the ns is fully annotated

KubeArmor/core/kubeUpdate.go
```diff
@@ -2382,12 +2393,103 @@ func (dm *KubeArmorDaemon) removeBackUpPolicy(name string) {
...
+		}
+
+		for idx, endpoint := range dm.EndPoints {
+			// skip all endpoints not in current namespace
+			if endpoint.NamespaceName != ns.Name {
+				continue
+			}
+			dm.Logger.Printf("default posture for endopoint %v is %v", endpoint.EndPointName, endpoint.DefaultPosture)
+			dm.Logger.Printf("global default posture changed to %v", posture)
+			dm.Logger.Printf("[check] if comdition met: %v", endpoint.DefaultPosture == posture)
```
achrefbensaad:
```suggestion
			dm.Logger.Printf("[check] if condition is met: %v", endpoint.DefaultPosture == posture)
```

KubeArmor/core/kubeUpdate.go
```diff
@@ -2438,6 +2546,34 @@ func (dm *KubeArmorDaemon) UpdateDefaultPosture(action string, namespace string,
...
+	case "audit", "Audit":
+		return "audit"
+	case "block", "Block":
+		return "block"
+	default:
+		return "audit"
+	}
+}
+
+func (dm *KubeArmorDaemon) UpdateGlobalPosture(posture tp.DefaultPosture) {
```
achrefbensaad, line 2557: Can you please handle the lint issue

tests/kubearmor_config/kubearmor_config_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+	Describe("Unannotated", func() {
+
+		It("default posture will be set to global posture configs", Label("unannotated"), func() {
+			// apply a allow based policy
+			err := K8sApply([]string{"manifests/ksp-unannotated-allow.yaml"})
+			Expect(err).To(BeNil())
+
+			// wait for policy
+			time.Sleep(5 * time.Second)
```
achrefbensaad: we dont need sleep after apply anymore, apply will sleep for 5 seconds before returning

achrefbensaad (changes requested):

KubeArmor/core/kubeUpdate.go
rksharma95: this is to avoid security concerns as `ns` is loop variable and we are passing it as a reference in helping functions, that is why it's been copied to local variable and that is being used in subsequent functions.

KubeArmor/core/kubeUpdate.go
rksharma95: these are just part of over-commenting for debug/review purpose, i was intended to remove before merge.

tests/kubearmor_config/kubearmor_config_test.go
rksharma95: yes the tests are need to be refactored now as #778 merged. I'm on it

rksharma95:
issue #1038 needs to be closed first, for the CI tests to be passed.

KubeArmor/core/kubeUpdate.go
rksharma95: I've removed debugging logs

achrefbensaad (approved):
LGTM.

Ankurk99 (approved):
LGTM!

deployments/AKS/kubearmor.yaml
```diff
@@ -2719,3 +2723,18 @@ spec:
...
       status: {}
+---
+apiVersion: v1
+data:
+  cluster: default
+  defaultFilePosture: block
+  gRPC: "32767"
+  hostDefaultCapabilitiesPosture: block
+  hostDefaultNetworkPosture: block
+  visibility: process,file,network,capabilities
```
nyrahul: Default posture should be kept to audit.

nyrahul (dismissed):

deployments/AKS/kubearmor.yaml
rksharma95: handled :+1:

Ankurk99:
Please generate deployment yamls based on the object.go changes. The deployment still contains the block default posture

KubeArmor/core/kubeUpdate.go
```diff
@@ -2500,3 +2623,79 @@ func (dm *KubeArmorDaemon) WatchDefaultPosture() {
...
+	})
+
+	go factory.Start(wait.NeverStop)
+	factory.WaitForCacheSync(wait.NeverStop)
+	dm.Logger.Print("Started watching Configmap")
+
+}
+
+// Returns KubeArmor configmap
+func (dm *KubeArmorDaemon) GetConfigMapNS() string {
```
achrefbensaad, line 2691: Can you please handle the lint issue

tests/kubearmor_config/kubearmor_config_test.go
```diff
@@ -0,0 +1,226 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 Authors of KubeArmor
+
+package kubearmor_config
```
achrefbensaad: please handle lint

tests/kubearmor_config/manifests/ubuntu-deployment.yaml
```diff
@@ -0,0 +1,83 @@
...
+    matchLabels:
+      container: fullyannotated
+  template:
+    metadata:
+      labels:
+        container: fullyannotated
+    spec:
+      containers:
+      - name: ubuntu-1
+        image: kubearmor/ubuntu-w-utils:0.1
```
achrefbensaad: Can you please add a new line

achrefbensaad (changes requested):

achrefbensaad (approved):

achrefbensaad:
LGTM!

PR #839, opened 2022-08-22
Two test suits added for unlinkat syscall and tcp_connect / tcp_accept kernel functions

aishwarya25252:
Test suits for syscall unlinkat and kernel functions tcp_connect , tcp_accept added


achrefbensaad:
@aishwarya25252, can we merge the two test suits `syscallchk` and `tcpchk` into one?

aishwarya25252:
@achrefbensaad yeah sure we can have both the test suit as single , will do so then

aishwarya25252:
@achrefbensaad can you suggest a way to trigger un_link ? as both rm and rm -rf triggers un_linkat syscall . Thanks

achrefbensaad:
use `unlink` command directly

achrefbensaad:
We can make use  of this snippet to call arbitrary syscalls .
```
#define _GNU_SOURCE
#include <unistd.h>
#include <sys/syscall.h>
#include <sys/types.h>
#include <sys/stat.h> 
#include <fcntl.h> 
int
main(int argc, char *argv[])
{
    char path[] = "/proc/swaps";
    long tid;
    tid = syscall(SYS_open, path,O_RDONLY);
}
```

tests/unitest/unitest_test.go
```diff
@@ -0,0 +1,98 @@
...
+	It("Checks for UNLINKAT syscall in logs for rm", func() {
+		_, _, err := K8sExecInPod(ub, "ubuntu-pods", []string{"touch","test.tmp"})
+		Expect(err).To(BeNil())
+		err = KarmorLogStart("system", "ubuntu-pods", "File", ub)
+		Expect(err).To(BeNil())
+		_, _, err = K8sExecInPod(ub, "ubuntu-pods", []string{"rm","test.tmp"})
+		Expect(err).To(BeNil())
+		logs, _, err := KarmorGetLogs(5*time.Second, 5)
+		Expect(logs[len(logs)-1].Source).To(Equal("/bin/rm test.tmp"))
+		Expect(logs[len(logs)-1].Data).To(MatchRegexp("syscall=SYS_UNLINKAT*"))
```
achrefbensaad: it is not garenteed that the last event will be the syscall event.

aishwarya25252: yups , I got the issue there , also there isn't any specific way to specify filters for karmorlogs . Should I just apply some custom karmor policies for each of the specs and check alerts ? like its been done in syscalls test suit . Any other suggestions will also be appreciated .

tests/unitest/unitest_test.go
```diff
@@ -0,0 +1,159 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 Authors of KubeArmor
+
+package unitest
+
+import (
+        "time"
+	"regexp"
+
+        . "github.com/kubearmor/karts/util"
```
daemon1024:
```suggestion
        . "github.com/kubearmor/KubeArmor/tests/util"
```

Need to update the package name due to #916

PR #847, opened 2022-08-23
refactor getting logs and alerts logic

rksharma95:
refactored the way, logs/alerts was matched before to pass the test. with this PR the dependency on the position of the telemetry events will be eliminated. Some utility functions are also added. 

Signed-off-by: rk <ramakant@accuknox.com>


rksharma95:
closing this PR as this fix handled in PR #778

PR #849, opened 2022-08-23
run containerized mode without host mode enabled

Ankurk99:
Currently KubeArmor when run only in containerized mode (without host mode) -enableKubeArmorPolicy is not receiving policies due to dependency on -enableKubeArmorHostPolicy. This fix will enable us to run containerized mode without -enableKubeArmorHostPolicy


KubeArmor/core/kubeArmor.go
```diff
@@ -616,7 +616,7 @@ func KubeArmor() {
 		dm.Logger.Print("Started to monitor container security policies on gRPC")
 	}
 
-	if !cfg.GlobalCfg.K8sEnv && cfg.GlobalCfg.HostPolicy {
+	if !dm.K8sEnabled {
```
daemon1024: Wouldn't this enable Host Policy even if HostPolicy is disabled?

Ankurk99: No, we will receive the host policy but it won't be enforced.

daemon1024: But we shouldn't be receiving the host policy :)

Ankurk99: Actually we are checking again if hostPolicy is enabled when creating the apparmor profile and applying it. Here: https://github.com/kubearmor/KubeArmor/blob/main/KubeArmor/enforcer/appArmorEnforcer.go#L588

daemon1024:
Idk y did this close 😓 Follow up #852

PR #852, opened 2022-08-23
run containerized mode without host mode enabled

daemon1024:
> Currently KubeArmor when run only in containerized mode (without host mode) -enableKubeArmorPolicy is not receiving policies due to dependency on -enableKubeArmorHostPolicy. This fix will enable us to run containerized mode without -enableKubeArmorHostPolicy 
~ @Ankurk99 

Ref https://github.com/kubearmor/KubeArmor/pull/849

### DIFF
--- a/KubeArmor/core/kubeArmor.go
+++ b/KubeArmor/core/kubeArmor.go
@@ -609,22 +609,22 @@ func KubeArmor() {
 		dm.Logger.Print("Started to monitor host security policies")
 	}
 
-	policyService := &policy.ServiceServer{}
-
-	if enableContainerPolicy {
-		policyService.UpdateContainerPolicy = dm.ParseAndUpdateContainerSecurityPolicy
-		dm.Logger.Print("Started to monitor container security policies on gRPC")
-	}
-
-	if !cfg.GlobalCfg.K8sEnv && cfg.GlobalCfg.HostPolicy {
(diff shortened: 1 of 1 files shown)


KubeArmor/core/kubeArmor.go
```diff
@@ -609,23 +609,23 @@ func KubeArmor() {
...
-
-	if enableContainerPolicy {
-		policyService.UpdateContainerPolicy = dm.ParseAndUpdateContainerSecurityPolicy
-		dm.Logger.Print("Started to monitor container security policies on gRPC")
-	}
-
-	if !cfg.GlobalCfg.K8sEnv && cfg.GlobalCfg.HostPolicy {
-		policyService.UpdateHostPolicy = dm.ParseAndUpdateHostSecurityPolicy
-		dm.Node.PolicyEnabled = tp.KubeArmorPolicyEnabled
+	if !dm.K8sEnabled && (enableContainerPolicy || cfg.GlobalCfg.HostPolicy) {
```
Ankurk99, line 612: do we need the & condition (enableContainerPolicy || cfg.GlobalCfg.HostPolicy) when we are already checking it inside the if condition?

daemon1024: Yes, we only register the service if either of them are enabled.

Ankurk99 (approved):
LGTM

PR #853, opened 2022-08-23
Compile BPF code using vmlinux.h

### LINKED ISSUE #803: Debian 11 - `linux-headers` package does not have kconfig file
**Environment** - Self-managed k8s with Debain 11 (Bullseye)

**Kernel Version**
```
root@k8s-master:~# uname -a
Linux k8s-master 5.10.0-8-amd64 #1 SMP Debian 5.10.46-4 (2021-08-03) x86_64 GNU/Linux
```

**Problem**
The code ([KubeArmor/BPF](https://github.com/kubearmor/KubeArmor/tree/main/KubeArmor/BPF)) depends on files like `Kconfig` and `kconfig.h` which we get from the `linux-headers` package. Without the presence of these `kconfig` files, the `kubearmor-init` container will fail. 

**Issue** - The `kconfig` files are removed from the `linux-headers` package in the [upstream kernel](https://github.com/torvalds/linux/commit/9945722afdc3443eab826b2da1122509a13a50a5). In some distros (which uses same script as upstream kernel to building their `linux-headers` package) you won't find the `kconfig` files. In Debain 11, even if the `linux-headers` package are installed, the init containers are failing because the `kconfig` files are not present in the kernel header directory.

wazir-ahmed:
If the host has kernel BTF information (`/sys/kernel/btf/vmlinux`),
then generate `vmlinux.h` using `bpftool` and compile the BPF code
using `vmlinux.h`.

This removes dependency on kernel headers to be present in the host
system if it uses a kernel > v5.4.

Fixes #803 

Signed-off-by: Wazir Ahmed <wazir@accuknox.com>

### DIFF
--- a/KubeArmor/BPF/Makefile
+++ b/KubeArmor/BPF/Makefile
@@ -1,43 +1,58 @@
 include Makefile.vars
 
 .PHONY: all
-all: chkdir syscall_test system_monitor.bpf.o system_monitor.container.bpf.o system_monitor.host.bpf.o
+all: kernel_headers libbpf syscall_test system_monitor.bpf.o system_monitor.container.bpf.o system_monitor.host.bpf.o
 
-.PHONY: chkdir
-chkdir:
-ifeq (,$(wildcard $(KRNDIR)/Kconfig))
-	@echo "Your kernel path[$(RED)$(KRNDIR)$(NC)] is incorrect. Use 'make KRNDIR=[KERNEL-SRC-PATH]'."
-	Quitting 
+.PHONY: kernel_headers
+kernel_headers:
+ifeq ($(BTF_SUPPORTED),1)
+	@echo "$(GREEN)Kernel BTF information found$(NC)"
+	@echo "Generating vmlinux.h for kernel $(KRNV_X).$(KRNV_Y).$(KRNV_Z)"
+	$(Q)bpftool btf dump file $(KRBTF) format c > $(VMLINUX)/vmlinux.h
 else
-	@echo "Using kernel path[$(GREEN)$(KRNDIR)$(NC)]"
+	@echo "$(RED)Kernel BTF information not found$(NC)"
+	@echo "Trying to use kernel headers present in the host"
+ifeq (,$(wildcard $(KRNDIR)/Kconfig))
+	@echo "$(RED)Auto-detected kernel headers path [$(KRNDIR)] is incorrect.$(NC) Use 'make KRNDIR=[KERNEL-SRC-PATH]'."
+	Quitting
+else
+	@echo "$(GREEN)Using kernel headers at $(KRNDIR)$(NC)"
 endif
+endif
+
+.PHONY: libbpf
+libbpf:
 ifeq (,$(wildcard $(LIBBPF)/src/libbpf.c))
-	git submodule update --init --recursive
+	$(Q)git submodule update --init --recursive
 endif
 ifeq (,$(wildcard $(LIBBPF)/src/libbpf.a))
-	make -C $(LIBBPF)/src
+	$(Q)make -C $(LIBBPF)/src
 endif
 
 .PHONY: syscall_test
 syscall_test: tests/Makefile
-	@make -C tests KRNDIR="$(KRNDIR)"
+	$(Q)make -C tests KRNDIR="$(KRNDIR)"
 
-system_monitor.bpf.o: $(SYSMONITOR) syscall_test
+system_monitor.bpf.o: $(SYSMONITOR) $(VMLINUX_H) syscall_test
 	@echo "Compiling eBPF bytecode: $(GREEN)$@$(NC) ..."
 	@echo "Using Compiler flags: $(CFlags)"
 	$(Q)$(CL) $(KF) -DMONITOR_HOST_AND_CONTAINER $(CFlags) -c $< -o -| llc -march=bpf -mcpu=probe -filetype=obj -o $@
 
-system_monitor.container.bpf.o: $(SYSMONITOR) syscall_test
+system_monitor.container.bpf.o: $(SYSMONITOR) $(VMLINUX_H) syscall_test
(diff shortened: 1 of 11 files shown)


KubeArmor/BPF/enforcer.bpf.c
```diff
@@ -13,6 +13,10 @@ char LICENSE[] SEC("license") = "Dual BSD/GPL";
 #define MAX_BUFFERS 1
 #define PATH_BUFFER 0
 
+/* In some of the distro, the generated
+ * vmlinux.h does not contain this value */
+#define PROC_PID_INIT_INO   0xEFFFFFFCU
+
```
daemon1024: Wouldn't this overwrite if present?
Perhaps should be something like:
```suggestion
#ifndef  PROC_PID_INIT_INO
/* In some of the distro, the generated
 * vmlinux.h does not contain this value */
#define PROC_PID_INIT_INO   0xEFFFFFFCU
#endif

```

wazir-ahmed: In [<linux/proc_ns.h>](https://elixir.bootlin.com/linux/latest/source/include/linux/proc_ns.h#L46), `PROC_PID_INIT_INO` is defined as an `enum`. We cannot use `ifdef` or `ifndef` with enum constants.

daemon1024: Okay! Then the question is would this overwrite? the enum definition?

wazir-ahmed: Yes. We cannot check for its existence, but we can overwrite enum constants with `#define`.

daemon1024: It should be fine, unless Some future version of the kernel updates the version 👍🏽

daemon1024:
Let's remove 18.04 tests, they are anyway deprecated by Github. Else the fix for failing tests would be to install latest bpftool.

daemon1024:
Have you tested it out on Ubuntu 18.04? With vagrant machine?
https://github.com/kubearmor/KubeArmor/runs/7976043831?check_suite_focus=true 
The test failing here seems to be weird.

wazir-ahmed:
> Have you tested it out on Ubuntu 18.04? With vagrant machine?

@daemon1024 No, I haven't tested it in 18.04.

daemon1024:
Failing tests are unrelated to the PR, Can y'all please review otherwise? Thanks.

KubeArmor/BPF/Makefile
```diff
@@ -33,20 +33,27 @@ endif
...
+# with native target, e.g., x64, arm64, etc. 'opt' does bpf CORE IR builtin
+# processing (llvm12) and IR optimizations. 'llvm-dis' converts
+# 'opt' output to IR, and finally 'llc' generates bpf byte code.
+# Ref https://elixir.bootlin.com/linux/v5.19.5/source/samples/bpf/Makefile#L439
+
 system_monitor.bpf.o: $(SYSMONITOR) $(VMLINUX_H) syscall_test
 	@echo "Compiling eBPF bytecode: $(GREEN)$@$(NC) ..."
 	@echo "Using Compiler flags: $(CFlags)"
-	$(Q)$(CL) $(KF) -DMONITOR_HOST_AND_CONTAINER $(CFlags) -c $< -o -| llc -march=bpf -mcpu=probe -filetype=obj -o $@
+	$(Q)$(CL) $(KF) -DMONITOR_HOST_AND_CONTAINER $(CFlags) -Xclang -disable-llvm-passes -c $< -o - | opt -O2 -mtriple=bpf-pc-linux | llvm-dis | llc -march=bpf -mcpu=probe -filetype=obj -o $@
```
wazir-ahmed, line 46: We need to use the same compilation command in `tests/Makefile` as well. Check the `syscall_test` target above.

daemon1024: We don't use any CORE reads so it should be fine imo. This is needed to process CORE IR.

Ankurk99:
LGTM, but I haven't tried compiling the code. Do you want me to test the compilation?

achrefbensaad (approved):
LGTM, @wazir-ahmed  Can you please add the installation of bpftool to the setup script contribution/self-managed-k8s/setup.sh ?

contribution/self-managed-k8s/setup.sh
```diff
@@ -21,7 +21,7 @@ export DEBIAN_FRONTEND=noninteractive
 echo 'debconf debconf/frontend select Noninteractive' | sudo debconf-set-selections
 
 # install dependencies and llvm--toolchain
-sudo apt-get -y install build-essential libelf-dev pkg-config net-tools linux-headers-generic
+sudo apt-get -y install build-essential libelf-dev pkg-config net-tools linux-headers-$(uname -r) linux-tools-$(uname -r)
```
wazir-ahmed, line 24: `linux-*-generic` packages always points to the default kernel the OS comes with. If the user updates the kernel, then it still points to older kernel version.

wazir-ahmed: @achrefbensaad Updated `setup.h`. The `linux-tools` package in Ubuntu has `bpftool`.

achrefbensaad:
Tested on:
- ubuntu 18.04 kernel 4.15.0-163-generic
- ubuntu 22.04 kernel 5.15.0-30-generic

daemon1024:
And Github tested on Ubuntu 20.04 :D

daemon1024 (approved):

wazir-ahmed:
Tested on:
- Debian 11 (Bullseye) w/ kernel 5.10.0-17-amd64
- Ubuntu 20.04 (Focal) w/ kernel 5.4.0-125-generic (v5.4 is the default kernel Ubuntu 20.04 ships with. CI uses the updated kernel v5.15)

wazir-ahmed:
The PR looks good to me 👍

PR #860, opened 2022-08-25
disable fallback annotation logic when the controller is healthy

### LINKED ISSUE #826: Containers restart after Kuberarmor install
## Bug Report

**General Information**

As far as I know, prior to version 0.5, deployments were edited manually to apply the required annotation for KubeArmor. But this method is not applicable, for example, when using a database inside a container. The data was frayed, as a result, the database was corrupted. After version 0.5, annotations were added using the admission controller, which in fact should not lead to restarting the pod and corrupting the database in the future. Until some time, everything worked for me as it should, but after updating 0.5.5, the containers began to restart again, which entailed the destruction of the database.

### LINKED ISSUE #840: Disable kubearmor annotation fallback logic when annotation controller is active.
## Feature Request

**Short Description**

We want to prevent kubearmor fallback logic from executing when the annotation controller is present.

~~The annotation controller will add to any new deployment the label `kubearmor-policy: enabled` and add a check in the fallback logic to not run if that annotation exist at the deployment level.~~

KubeArmor fallback logic will run only if the pod does not have all the required annotations. (one annotation per container).

achrefbensaad:
fixes #826, fixes #840 

We want to disable kubearmor annotation patch logic (fallback logic) when the annotations controller is working as expected.
Currently the fallback logic causes deployments restarts regardless of whether the annotation controller is present or not.

### DIFF
--- a/KubeArmor/core/kubeUpdate.go
+++ b/KubeArmor/core/kubeUpdate.go
@@ -452,6 +452,8 @@ func (dm *KubeArmorDaemon) WatchK8sPods() {
 				// create a pod
 
 				pod := tp.K8sPod{}
+				appArmorAnnotationsCount := 0
+				containersCount := 0
 
 				// need this for apparmor profile
 				var podOwnerName string
@@ -598,6 +600,7 @@ func (dm *KubeArmorDaemon) WatchK8sPods() {
 								containerName := strings.Split(k, "/")[1]
 								appArmorAnnotations[containerName] = strings.Split(v, "/")[1]
 							}
+							appArmorAnnotationsCount += 1
 						}
 					}
 
@@ -606,6 +609,7 @@ func (dm *KubeArmorDaemon) WatchK8sPods() {
 							appArmorAnnotations[container.Name] = "kubearmor-" + pod.Metadata["namespaceName"] + "-" + podOwnerName + "-" + container.Name
 							updateAppArmor = true
 						}
+						containersCount += 1
 					}
 
 					if event.Type == "ADDED" {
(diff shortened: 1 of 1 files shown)


daemon1024:
Can you briefly explain what the code change is doing. I don't really understand anything seeing the changes 😅

KubeArmor/core/kubeUpdate.go
```diff
@@ -606,14 +609,15 @@ func (dm *KubeArmorDaemon) WatchK8sPods() {
...
+						containersCount += 1
 					}
 
 					if event.Type == "ADDED" {
 						// update apparmor profiles
 						dm.RuntimeEnforcer.UpdateAppArmorProfiles(pod.Metadata["podName"], "ADDED", appArmorAnnotations)
 
 						if updateAppArmor && pod.Annotations["kubearmor-policy"] == "enabled" {
-							if deploymentName, ok := pod.Metadata["deploymentName"]; ok {
+							if deploymentName, ok := pod.Metadata["deploymentName"]; ok && containersCount != appArmorAnnotationsCount {
```
daemon1024, line 620: We only patch deployment if annotation don't exist already?

achrefbensaad: yes. If the annotation does not exist or it is uncomplete then we need to patch the deployment as reaching this condition means that the annotation controller is not healthy.

achrefbensaad:
> Can you briefly explain what the code change is doing. I don't really understand anything seeing the changes 😅

@daemon1024, I have updated the PR description.

Ankurk99 (approved):
LGTM!

daemon1024 (approved):

PR #861, opened 2022-08-25
gracefully terminate kubearmor container

### LINKED ISSUE #851: Handle KubeArmor container termination gracefully
## Bug Report

KubeArmor container doesn't clean up resources properly when uninstalled. This only happens when KubeArmor runs in containerised environment and not in systemd/process mode.

**General Information**

- Environment description (GKE, VM-Kubeadm, k3s, docker ...)
- Target containers/pods - KubeArmor

**To Reproduce**

With AppArmor:
1. Install KubeArmor in k8s
2. Setup Policies
3. karmor uninstall ( don't remove CRD)

With BPFLSM:
1. Install KubeArmor in k8s
2. karmor uninstall 

**Expected behavior**

No artifacts remain for KubeArmor.

**Actual Behavior**

AppArmor:
- AppArmor Profiles stay intact and actions are getting denied even without KubeArmor

BPFLSM:
- Presence of pinned map in /sys/fs/bpf

daemon1024:
When k8s or CRI stop containers, they send SIGTERM to pid 1 in the container. pid 1 in this case is bash which executes our script and never really passed the signal onto child processes for it to gracefully handle and waited for the termination timeout to get abruptly killed and hence not cleaning up resources.

We now leverage exec, to put KubeArmor process as PID 1 so KubeArmor directly handles the termination signals.

Now KubeArmor will clean up resources while being terminated, also the process will be quicker since we don't wait for the entire period for it to get abruptly killed later.

Fixes #851


daemon1024:
I took the liberty to remove DEBUG flag, since I don't think it was being leverage anyway. Lemme know if you think otherwise.

nyrahul (approved):
LGTM 👍🏼

PR #872, opened 2022-09-02
use DirectoryOrCreate for lib modules

daemon1024:
It's possible that the lib modules directory is not available, this will just create an empty directory if it's not available and we have other relevant checks in the init container already.

Ref #771 #827 

// @PrimalPimmy


achrefbensaad (approved):
LGTM!

PR #874, opened 2022-09-02
update kubearmor-test yamls

achrefbensaad:
fixes latest release failures when we create a new release branch


daemon1024:
Isn't the default value latest?

daemon1024 (approved):

achrefbensaad:
We need to explicitly set the value to latest so that we can update it later using sed to the tag we are releasing. 
see here https://github.com/kubearmor/KubeArmor/blob/main/.github/workflows/ci-latest-release.yml#L58

Issue #885, opened 2022-09-08
match container-name field in security policy


Ankurk99:
## Bug Report

**General Information**

In containerized un-orchestrated mode, currently we are checking only the Spec.Selector.MatchLabels field for the container name. But we should instead check for some specific field under MatchLabels (kubearmor.io/container.name) to match the container name.

**Current Behavior**
Suppose a sample policy contains:
```
  selector:
    matchLabels:
      abcd: nginx
```
then also KubeArmor will match nginx as the container name.

**Expected Behavior**
Only match container name with kubearmor.io/container.name field value 
```
  selector:
    matchLabels:
      kubearmor.io/container.name: nginx
```
Ref: https://github.com/kubearmor/KubeArmor/blob/main/KubeArmor/core/kubeUpdate.go#L1286

HariVamsiK:
Hi
I want to work on this issue. Could you please assign it to me, it would be beneficial for my academics too.
Thanks

Ankurk99:
Hello @HariVamsiK glad to see your interest. 
Sorry but this issue is already addressed in #886.
There are many other open issues, you might want to take a look at them. We would be more than happy to help.

DelusionalOptimist:
@Ankurk99 since this has been handled in #886, let's close the issue?

PR #886, opened 2022-09-08
fix container-name field matching (un-orchestrated)

Ankurk99:
Ref: https://github.com/kubearmor/KubeArmor/issues/885

### DIFF
--- a/KubeArmor/core/kubeUpdate.go
+++ b/KubeArmor/core/kubeUpdate.go
@@ -1283,7 +1283,12 @@ func (dm *KubeArmorDaemon) ParseAndUpdateContainerSecurityPolicy(event tp.K8sKub
 	containername := ""
 	for k, v := range secPolicy.Spec.Selector.MatchLabels {
 		secPolicy.Spec.Selector.Identities = append(secPolicy.Spec.Selector.Identities, k+"="+v)
-		containername = v // TODO: fix identity behavior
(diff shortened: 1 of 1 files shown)


KubeArmor/core/kubeUpdate.go
```diff
@@ -1283,7 +1283,12 @@ func (dm *KubeArmorDaemon) ParseAndUpdateContainerSecurityPolicy(event tp.K8sKub
 	containername := ""
 	for k, v := range secPolicy.Spec.Selector.MatchLabels {
 		secPolicy.Spec.Selector.Identities = append(secPolicy.Spec.Selector.Identities, k+"="+v)
-		containername = v // TODO: fix identity behavior
+		if k == "kubearmor.io/container.name" {
+			containername = v
+		} else {
+			dm.Logger.Warnf("Fail to apply policy. The MatchLabels container name key should be `kubearmor.io/container.name` ")
+			return
```
daemon1024, line 1290: In future we might need to support other labels too like https://docs.docker.com/engine/reference/commandline/run/#set-metadata-on-container--l---label---label-file

So should we Warn here or just simply ignore?

Ankurk99: It would be a bit confusing for a user to actually know the reason for the failure of policy application. I think it's fine for now, we can update it in the future.

nyrahul (approved):
LGTM

PR #887, opened 2022-09-08
fix stable ci

achrefbensaad:


.github/workflows/ci-stable-release.yml
```diff
@@ -23,7 +24,9 @@ jobs:
         run: |
           value=`cat STABLE-RELEASE`
           docker pull docker.io/kubearmor/kubearmor:$value
+          docker pull docker.io/kubearmor/kubearmor-init:latest
           docker image tag kubearmor/kubearmor:$value kubearmor/kubearmor:stable
+          docker image tag kubearmor/kubearmor-init:latest kubearmor/kubearmor-init:stable
```
daemon1024, line 29: We use https://github.com/kubearmor/KubeArmor/blob/ba0ade46a8c08315177a1ac518a6dbf71c1a5a12/deployments/get/objects.go#L527

Should we update init container to use stable as well

daemon1024 (approved):

.github/workflows/ci-stable-release.yml
achrefbensaad: yes, it makes sense, that will allow us to get a more stable containers. Lets do it in a seperate PR.

PR #892, opened 2022-09-12
enable multicontainer support

### LINKED ISSUE #675: Multicontainer support for KubeArmor
## Feature Request
Allow KubeArmor to apply policies at container level

Currently if we apply any policy to a pod running multiple containers, the policy get applied at pod level and we cannot create separate KubeArmor profile for different containers in the pod

Include container name field in the security policies and modify KubeArmor to apply policies to different containers based on this container name field.

For example: a pod is running 3 containers (container1/2/3) and we want to apply the policy to container1 and container2 then policy can be modified like:
```
spec:
  Selector:
    containerName:
        name: container1, container2
```

achrefbensaad:


Ankurk99 (approved):
LGTM

nyrahul (approved):
Thanks for adding the detailed test suite.

PR #900, opened 2022-09-18
Add code generation for Capabilties

### LINKED ISSUE #781: using non-privileged containers for kubearmor daemonset
## Feature Request

**Short Description**

Privileged containers are usually frowned upon. Almost every static scanning engine will flag this as an issue. In lot of cases, orgs also deploy admission controllers that would not allow containers to be installed in priviledged mode.

Kubearmor [currently uses](https://github.com/kubearmor/KubeArmor/blob/d6fef36897a751211e1c54780734bd02ee85a28d/deployments/get/objects.go#L362) privileged mode for its daemonset containers.

It is best to not use privileged mode, but to use specific capabilities for kubearmor.

Use specific capabilities in place of privileged mode. Cilium [recently added](https://isovalent.com/blog/post/cilium-release-112/#unprivileged-pod) this mode as well.

### Tasks
- [x] remove `privilege: true` from the deployment and drop all caps and enable specific required caps
- [x] Analysis of individual caps that are used and why that cap is needed
- [x] install kubearmor in non `kube-system` namespace
- [ ] install kubearmor with non clusteradmin role (identify least privileged role for applying apparmor annotations, watch pods/nodes and using that roles in the manifest)
- [x] drop caps from the code once the cap is used and no more required. For e.g., once the ebpf byte code is loaded we might want to drop BPF cap.

kranurag7:
This PR handles 
- remove `privileged: true` from Daemonset
- generation of manifest for Linux Capabilities in Daemonset

### DIFF
--- a/KubeArmor/build/kubearmor-test-containerd.yaml
+++ b/KubeArmor/build/kubearmor-test-containerd.yaml
@@ -51,7 +51,21 @@ spec:
         image: kubearmor/kubearmor-init:latest
         imagePullPolicy: Never
         securityContext:
-          privileged: true
+          privileged: false
+          capabilities:
+            drop: 
+              - ALL
+            add: 
+              - SETUID 
+              - SETGID
+              - SETPCAP
+              - SYS_ADMIN
+              - SYS_PTRACE 
+              - MAC_ADMIN
+              - SYS_RESOURCE 
+              - IPC_LOCK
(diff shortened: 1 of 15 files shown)


daemon1024 (approved):

nyrahul (approved):
LGTM

PR #905, opened 2022-09-19
Added an option to not log passed results

### LINKED ISSUE #904: Reduce logging by allowing to not log passed results
## Feature Request

**Short Description**

KubeArmor is very spammy in terms of logs, I'd want to disable logging of events with `result = "Passed"`

Amount of logs is enormous

Extra runtime flag which would control logging

michalschott:
resolves #904

### DIFF
--- a/KubeArmor/config/config.go
+++ b/KubeArmor/config/config.go
@@ -23,6 +23,7 @@ type KubearmorConfig struct {
 	LogPath           string // Log file to use
 	SELinuxProfileDir string // Directory to store SELinux profiles
 	CRISocket         string // Container runtime to use
+	LogPassedResults  string // Should passed results be logged
 
 	Visibility     string // Container visibility to use
 	HostVisibility string // Host visibility to use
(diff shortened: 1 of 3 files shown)


achrefbensaad:
Hi @michalschott , At what level did you notice that KubeArmor is spammy ?.

via `karmor log`, or a log collection tool or SIEM or an other tool ?

michalschott:
Logs from pod

achrefbensaad:
@michalschott , if you use `karmor logs` without `--logFilter` flag you will get only alerts.

michalschott:
@achrefbensaad indeed, but log aggregation pods are still forwarding that spam.

achrefbensaad:
So basicly, you want an on/off switch for passed events? if so, dropping events in the user space is not the best approach. It would be better to drop the events on the kernel space.

michalschott:
@achrefbensaad keen to agree with you, but my bpf knowledge is quite limited - reason I have added this simple switch.

Personally I think it would be good to have an option to not log passed events to stdou, so noise is reduced. It would be easier to spot violations.

PR #908, opened 2022-09-20
Fixes #769 Ignore empty policy in Kubeamor

### LINKED ISSUE #769: Ignore empty policy
## Bug Report

**Environment**
KubeArmor running in host	mode

**General Information**
Add check to KubeArmor to ignore accepting blank policies. If a policy ends with "---" KubeArmor currently takes it as an additional empty policy.
It will have issues for eg: If we run KubeArmor in containerized mode then it will create an empty apparmor profile (with name kubearmor_) for that empty policy.

**Expected behavior**
KubeArmor should ignore blank policies (policy yaml ending with "---").

Chinwendu20:
~~I wrote a check for host policy and container policy. I placed them in the kubeamor/common/common.go file and imported them in the kubeamor/policy/policy.go file. If the policy file is empty, res.status is assigned 0, container policy is not updated and then there would be a warning log with message, "Empty policy Event" . The function~~

Does not apply anymore

Fixes #769

### DIFF
--- a/KubeArmor/policy/policy.go
+++ b/KubeArmor/policy/policy.go
@@ -25,9 +25,22 @@ func (p *ServiceServer) ContainerPolicy(c context.Context, data *pb.Policy) (*pb
 	res := new(pb.Response)
 
 	err := json.Unmarshal(data.Policy, &policyEvent)
+
 	if err == nil {
-		p.UpdateContainerPolicy(policyEvent)
-		res.Status = 1
+
+		if policyEvent.Object.Metadata.Name != "" {
+
+			p.UpdateContainerPolicy(policyEvent)
+
+			res.Status = 1
+
+		} else {
+
+			kg.Warn("Empty Container Policy Event")
+
+			res.Status = 0
+		}
+
 	} else {
 		kg.Warn("Invalid Container Policy Event")
 		res.Status = 0
@@ -38,13 +51,27 @@ func (p *ServiceServer) ContainerPolicy(c context.Context, data *pb.Policy) (*pb
 
 // HostPolicy accepts host policy event on gRPC service and updates host security policies. It responds with 1 if success else 0.
 func (p *ServiceServer) HostPolicy(c context.Context, data *pb.Policy) (*pb.Response, error) {
+
 	policyEvent := tp.K8sKubeArmorHostPolicyEvent{}
 	res := new(pb.Response)
 
 	err := json.Unmarshal(data.Policy, &policyEvent)
 	if err == nil {
-		p.UpdateHostPolicy(policyEvent)
-		res.Status = 1
+
+		if policyEvent.Object.Metadata.Name != "" {
+
+			p.UpdateHostPolicy(policyEvent)
+
+			res.Status = 1
+
+		} else {
+
+			kg.Warn("Empty Host Policy Event")
+
+			res.Status = 0
+
+		}
+
 	} else {
 		kg.Warn("Invalid Host Policy Event")
 		res.Status = 0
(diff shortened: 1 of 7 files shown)


rksharma95:
@Chinwendu20 appreciate your contribution, Check the DCO failed, Please use git commit -s to sign-off in your commits. Thanks.

Chinwendu20:
Okay looks like my package import path was incorrect. I would fix that

KubeArmor/common/common.go
```diff
@@ -443,3 +444,100 @@ func WriteToFile(val interface{}, destFile string) error {
...
+			}
+		case reflect.Ptr:
+
+			if !struct_value.IsNil(){	
+
+				EmptyPolicy=false
+			
+			}
+		}
+	}
```
daemon1024: Why do we need to use the reflect package here?

Shouldn't the check be as simple as 
`if policyEvent.Object.Metadata &&  policyEvent.Object.Metadata.Name != "" {}`

daemon1024 (changes requested):
Hey @Chinwendu20, Thanks a lot for handling the changes!

As I iterated on Slack,
We should check for 
policyEvent.Object.Name != ""

And that should be good enough.

KubeArmor/common/common.go
Chinwendu20: Okay I saw that the Name field in Metadata is optional. Actually there is another field Generate_Name that one can use to generate metadata name. So a valid policy can do without that.

image

Chinwendu20: Thanks. You are right even if that were to be the case, then I should check for only Name and GenerateName field. Just when I looked at it I saw that all the fields were optional and perhaps it needed just one field.

daemon1024: > Is required when creating resources, some resources **may allow** a client to request the generation of an appropriate  automatically.

As it mentions may request client for auto generation, KubeArmor doesn't do that, so Ig we can deem it as required.

daemon1024:
I couldn't find it but can we check the apiVersion and Kind on KubeArmor side as well?

Chinwendu20:
> I couldn't find it but can we check the apiVersion and Kind on KubeArmor side as well?

This is the type of the policy Event

image
 
I cannot see any field for apiversion or kind

KubeArmor/common/common.go
Chinwendu20: Okay thanks a lot. I will implement the change

Chinwendu20:
Do you know of any other way that I can obtain the apiversion or kind from the kubeamor side?

daemon1024:
It should be fine, let's just add the Name check :)

Chinwendu20:
Hi @daemon1024 I would just check for the Metadata Name. As even empty polices have Metadata as the creation time is always present.

KubeArmor/policy/policy.go
```diff
@@ -38,12 +47,22 @@ func (p *ServiceServer) ContainerPolicy(c context.Context, data *pb.Policy) (*pb
...
+
+		if policyEvent.Object.Metadata.Name != ""{
+
 		p.UpdateHostPolicy(policyEvent)
+
+		}else{
+
+		kg.Warn("Empty Host Policy Event")
+
+		}
```
daemon1024, line 73: Let's add res,Status = 0 as well in the else condition and move the status = 1 to the if condition. Also I would suggest you to run `gofmt` once it will handle the formating for ya!

Chinwendu20: Okay thanks

Chinwendu20: okay thanks I have done that

daemon1024 (approved):
Nice 👌🏽

PR #912, opened 2022-09-23
use informers instead of watchers for default posture

daemon1024:
Watchers are simple to implement and lowest level of abstraction to monitor K8s events but it's not resilient. SharedInformer provides eventually consistent linkage to the objects we want to monitor.

This commit fixes the issue where if due to some connection error we stopped watching namespace changes thereby breaking the namespace based default posture updates.

Ref 
https://pkg.go.dev/k8s.io/client-go/tools/cache#SharedInformer
https://stackoverflow.com/questions/59544139/kubernetes-client-go-watch-interface-vs-cache-newinformer-vs-cache-newsharedi


DelusionalOptimist (approved):
LGTM. :+1:

achrefbensaad (approved):
LGTM!

PR #921, opened 2022-09-30
fix capabilities allow policy bug

daemon1024:
Capability whitelist rules disabled network instead of caps due to the typo

Thank You @aishwarya25252 for reporting this.


aishwarya25252 (approved):

Ankurk99 (approved):
LGTM

PR #926, opened 2022-10-05
enforcer: fix directory rule behaviour

daemon1024:
AppArmor `**` or `*` rule doesn't include the directory itself. Taking an example of ps and proc. ps tries to access the `proc` directory itself before reading the underlying directories. When we save allow ps -> /proc/**, it didn't include the /proc/ directory itself so we update the directory rule to include both directory itself and the recursive pattern identifier.
Now we would use /proc/{,**}. `{}` signifies alternation and we provide it with to either match recursively (**) or the dir itself (empty string) 

BPFLSM kernel space refers to /proc/ as /proc since everything is a file. So we added a additional file rule from userspace when we need to deal with Directory Rules.

### DIFF
--- a/KubeArmor/enforcer/appArmorTemplate.go
+++ b/KubeArmor/enforcer/appArmorTemplate.go
@@ -80,7 +80,7 @@ profile {{.Name}} flags=(attach_disconnected,mediate_deleted) {
 	## == PRE END == ##
 
 	## == POLICY START == ##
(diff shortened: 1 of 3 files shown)


Ankurk99:
Great finding @daemon1024! 
So now if we want to recursively include `proc` dir, then we use `proc/` in the policy rule and that will create proc/{**} apparmor rule?

daemon1024:
Yes @Ankurk99 

So the rule would have been
```
/proc/ rw,
/proc/** rw,
```

`{}` accepts what alternate strings can follow for matching
So
```
/proc/{,**} rw,
```

Noting that there's a comma `,` here, the empty string signifies we want to match the directory itself and `*`/`**` after the comma signifies how to match inside the directory based on recursive or not.

nyrahul (approved):
LGTM 👍

PR #930, opened 2022-10-10
enable policy enforcement namespace wide

achrefbensaad:
Allow policy enforcement for namespace wide resources when selector is not set.


nyrahul (approved):
LGTM 👍

PR #938, opened 2022-10-14
add test for security_path_rmdir

achrefbensaad:
Signed-off-by: Achref ben saad <achref@accuknox.com>

### DIFF
--- a/KubeArmor/BPF/tests/checksyscall.sh
+++ b/KubeArmor/BPF/tests/checksyscall.sh
@@ -30,7 +30,7 @@ SYSCALL=${1#*/}
 
 if ! $SUDO ./syscheck $1 ; then
     echo "Disabling $SYSCALL ..."
-    echo "$SYSCALL" >> ignore.lst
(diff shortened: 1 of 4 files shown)


daemon1024:
```sh

**Comment by daemon1024** [comment]
```sh
if ! $SUDO ./syscheck $1 ; then
    echo "Disabling $SYSCALL ..."
    echo "$SYSCALL" >> ignore.lst
else
    echo "Enabling syscall $SYSCALL"
    grep "// CFlag=.*$" "$1.c" | sed -E "s/.*?=(.*?)$/\1/" >> cflags.lst
fi
```

I believe this is the relevant code.
I think We shouldn't have tests like this, like needing to check and every syscall to ignore.lst.
Any bpf program is dependent on a certain kernel features.

Adding a test for each one won't make sense. As we proceed we might need to add 10 more hooks that are related to SECURITY_PATH flag. But in that case we would need to add 10 more test cases here right?

Ignore list check should be based on flags rather than tests for each syscall. WDYT?

achrefbensaad:
@daemon1024 , I agree. Changes has been made.

daemon1024 (approved):

PR #940, opened 2022-10-15
[WIP] Adds BTF CO:RE functionality to KubeArmor

### LINKED ISSUE #789: BTF and CORE Support
## Feature Request

We migrated away from BCC and now we are leveraging libbpf for our eBPF programs.
But we depend on kernel headers to leverage various kernel structures in our programs.
Our BPF System monitor is not portable and we leverage init container to compile our BPF Monitor at runtime.

Kernels where BTF (BPF Type Format) is available, allows us to Compile Once and Run Anywhere (CO-RE)
More about it at https://nakryiko.com/posts/bpf-portability-and-co-re/

But the restriction here is that BTF Information needs to be available which is in general added after 5.2 Kernel Version.

We have CO-RE programs as part of BPF LSM enforcer but those programs anyway run only on kernels >5.8 and BTF information is a prerequisite for BPF LSM so it's fine there,
But our eBPF Monitor needs to be supported on older kernel versions.

With the introduction of CORE Support, We can remove the dependency on Kernel Headers and the requirement to compile at runtime.

- Work with Kernels which support BTF without needing Kernel Headers leveraging CO-RE
- Leverage BTFGen to ship BTF Files with KubeArmor Code for kernels not supporting CO-RE : https://kinvolk.io/blog/2022/03/btfgen-one-step-closer-to-truly-portable-ebpf-programs/
- Generate BTF Information at runtime for kernels that have vmlinux but not BTF Information with pahole : https://lists.iovisor.org/g/iovisor-dev/message/1969

Support both the approaches like Tracee does.

AllMight2099:
Fixes #789 
Signed-off-by: Nishanth R <nishusheela2011@gmail.com>


KubeArmor/monitor/systemMonitor.go
```diff
@@ -118,6 +120,9 @@ type ContextCombined struct {
 // StopChan Channel
 var StopChan chan struct{}
 
+//go:embed system_monitor.*
```
daemon1024: We should nams these binaries different so they are differentiated from runtime compiled binaries in general

KubeArmor/monitor/systemMonitor.go
```diff
@@ -251,22 +256,44 @@ func (mon *SystemMonitor) InitBPF() error {
 		return err
 	}
 
-	mon.Logger.Print("Initializing eBPF system monitor")
+	var bpfFile io.ReaderAt
 
+	mon.Logger.Print("Initializing eBPF system monitor")
+	bpfPath = os.Getenv("PWD") + "/monitor/"
```
daemon1024: This is breaking, why is there a need to change bpfPath

KubeArmor/monitor/systemMonitor.go
```diff
@@ -251,22 +256,44 @@ func (mon *SystemMonitor) InitBPF() error {
...
 	} else if cfg.GlobalCfg.Policy && cfg.GlobalCfg.HostPolicy { // container and host
+		var tempBpfFile, err = embededdBPFFiles.ReadFile("system_monitor.bpf.o")
+		if err != nil {
+			return fmt.Errorf("unable to read file: system_monitor.bpf.o")
+		}
+		bpfFile = bytes.NewReader(tempBpfFile)
 		bpfPath = bpfPath + "system_monitor.bpf.o"
 	}
 	mon.Logger.Printf("eBPF system monitor object file path: %s", bpfPath)
-	mon.BpfModule, err = cle.LoadCollection(bpfPath)
```
daemon1024, line 245: We need to keep this for fall back?

KubeArmor/monitor/systemMonitor.go
```diff
@@ -251,22 +256,44 @@ func (mon *SystemMonitor) InitBPF() error {
...
+	spec, err := cle.LoadCollectionSpecFromReader(bpfFile)
+	if err != nil {
+		return fmt.Errorf("unable to load collection spec from %v", err)
+	}
+
+	mon.BpfModule, err = cle.NewCollection(spec)
```
daemon1024: We should load this spec only if BTF is available, we need to add a check for BTF information check before load embed binaries else need to look at runtime config.

daemon1024 (changes requested):
Thanks for the PR! Find review comments inline.

Noting existing functionality shouldn't be breaking ( it is right now as suggested by the failing tests ). This should be an incremental change.

KubeArmor/monitor/systemMonitor.go
AllMight2099: Oh shoot, I forgot to revert that, I was using that to test if I could use a pre-compiled system_monitor binary that I saved in a different location. I'll revert it

KubeArmor/monitor/systemMonitor.go
AllMight2099: I've made these changes

KubeArmor/monitor/systemMonitor.go
AllMight2099: I've changed the binary names to `embedded_system_monitor` to reflect that this would be embedded inside the Kubearmor binary

KubeArmor/monitor/systemMonitor.go
AllMight2099: I've made these changes too

Ankurk99:
Hey @AllMight2099, Can you please remove the pushed object binaries?

AllMight2099:
> Hey @AllMight2099, Can you please remove the pushed object binaries?

Hi @Ankurk99, I was actually hoping that we keep the pushed object binaries like how we're doing it for the bpflsm enforcer, so I changed the build Makefile to remove `go clean -testcache` when we run `make testall`, because of which the tests are able to run, so I'm fixing the test cases right now

daemon1024:
We need to make it work for systems which don't have SECURITY_PATH enabled. Can you setup a RHEL 9 system and validate things?
We currently do this handling in init container as part of https://github.com/kubearmor/KubeArmor/blob/7fdfe6da8d0218c735db1d41ad12b70409c07cba/KubeArmor/BPF/tests/checksyscall.sh#L36

daemon1024:
Please don't commit the BTF files. We should be generating it in our CI/inside docker builds.

btf-utils/btfhub-btfgen.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash -e
```
daemon1024, line 1: Where is this script called?

AllMight2099: It's not called anywhere, I made it to download btfhub-archive and generate the reduced files

daemon1024: How would we ship it with our container? It should be called at docker build time tho right?

AllMight2099: This was meant to run everytime someone changed system_monitor.c file. I was going to store the reduced files as a zip and unzip it in the container

daemon1024: Wouldn't generating it as part of container image cover the case for 
> run everytime someone changed system_monitor.c file

We can assume that no one would expect BTF files when using non CORE in development environment.

AllMight2099: Generating it in the container means having btfhub-archive, which isn't feasible cause it takes around 5-ish GB

daemon1024: Hey @AllMight2099 , We have a misunderstanding here.

I am saying it to generate it at build time, not runtime. Like how we compile the KubeArmor code to a single binary in our Dockerfile and ship only the binary.

Similarly we download and reduce the btfhub at build time and copy the reduced files from staging to main container using Dockerfile.

AllMight2099: Ohhh, got it

daemon1024: For reference

https://github.com/kubearmor/KubeArmor/blob/fc08819ea1efc8d3c68b399d7b73fe44859baab6/Dockerfile#L18 built in builder image

https://github.com/kubearmor/KubeArmor/blob/fc08819ea1efc8d3c68b399d7b73fe44859baab6/Dockerfile#L32
Copied from builder image to actual image.

AllMight2099: I'll make those changes

kranurag7:
@nyrahul @daemon1024 Please take a look at this comment. This is relevant to this PR. 
https://github.com/aws/containers-roadmap/issues/1027#issuecomment-1433911314

daemon1024:
That's fine, we already have CORE Support, this was mostly to streamline CORE and still make it work in older kernels. We do need some effort here but Fargate would anyway need a different implementation provided it might be a sidecar model. Let's discuss on it once the support lands. 

Thanks @kranurag7

PR #953, opened 2022-10-25
copy ignore list to BPF module

rksharma95:
kubearmor will require this information inside container image at the time of loading ebpf based system monitor. 

Signed-off-by: rksharma95 <ramakant@accuknox.com>

### DIFF
--- a/KubeArmor/build/compile.sh
+++ b/KubeArmor/build/compile.sh
@@ -12,3 +12,4 @@ else
 fi
 
 cp *.bpf.o /opt/kubearmor/BPF/
+cp *.bpf.o ignore.lst /opt/kubearmor/BPF/
--- a/KubeArmor/build/compile.sh
+++ b/KubeArmor/build/compile.sh
@@ -11,5 +11,4 @@ else
     make
 fi
 
-cp *.bpf.o /opt/kubearmor/BPF/
 cp *.bpf.o ignore.lst /opt/kubearmor/BPF/


KubeArmor/build/compile.sh
```diff
@@ -12,3 +12,4 @@ else
 fi
 
 cp *.bpf.o /opt/kubearmor/BPF/
+cp *.bpf.o ignore.lst /opt/kubearmor/BPF/
```
nyrahul, line 14: Few points:
1. `*.bpf.o` is getting copied in two places.
2. What is the reason to copy `ignore.lst` ?

nyrahul (changes requested):

KubeArmor/build/compile.sh
rksharma95: we are generating `ignore.lst` based on syscheck result and kubearmor will require this info while loading system monitor which is missing in current kubearmor images and resulting in failing for the environment with unavailable hooks i.e. on rhel9 5.14.0 lsm hooks for unlink,rmdir is unavailable. 
with this change kubearmor-init container will make available the dynamic list `ignore.lst` for the kubearmor to be used.

nyrahul: @achrefbensaad , can you check if this is inline to your understanding? I thought we generate `ignore.lst` dynamically in the init container.

achrefbensaad: Yes we generate ignore.lst dynamically in the init container, but we need to copy it over to the main container so that KubeArmor knows what hooks it should not load.

KubeArmor/build/compile.sh
```diff
@@ -12,3 +12,4 @@ else
 fi
 
 cp *.bpf.o /opt/kubearmor/BPF/
```
achrefbensaad: can you please remove this line.
we are copying *.bpf.o in line 15.

achrefbensaad (changes requested):

nyrahul (approved):

Ankurk99 (approved):

achrefbensaad (approved):

PR #985, opened 2022-11-14
feat(enforcer): update default posture to audit

Ankurk99:
**Purpose of PR?**:
It changes the default posture behavior of KubeArmor to be Audit instead of denying everything. The reason being, auto-discovering rules may not result in an exhaustive policy thus may deny actions which can cause the application to fail.

**Does this PR introduce a breaking change?**
Lenient whitelist policy might break.

**If the changes in this PR are manually verified, list down the scenarios covered:**:
Manually verified with allow based policy resulting in "Audit" logs for actions which are not allowed.

**Additional information for reviewer?** :
_Mention if this PR is part of any design or a continuation of previous PRs_
No

**Checklist:**
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update
- [x ] PR Title follows the convention of  `<type>(<scope>): <subject>`
- [ ] Commit has unit tests
- [ ] Commit has integration tests

PR #986, opened 2022-11-14
fix(enforcer): add allow redundant rules to aid conflict resolution

daemon1024:
**Purpose of PR?**:

We removed redundant allow rules when default posture = audit. But this breaks our conflict resolution in from source profiles which aid lenient whitelisting policies.

This change will help make lenient whitelisting policies work with default posture audit.

Ref #985

Lenient Whitelisting Test with DefaultPosture = audit

**Checklist:**
- [x] Bug fix.


nyrahul (approved):
LGTM 👍

Issue #928, opened 2022-10-06
Additional checks for LSM Support


daemon1024:
## Feature Request

We currently rely on `/sys/kernel/security/lsm` to check for LSMs. This is fine on actual Hosts. But containerized hosts may not have this file and yet they could support stuff. So we should check for other ways to get the supported LSMs.

**Describe the solution you'd like**

AppArmor:
- As mentioned in https://kubernetes.io/docs/tutorials/security/apparmor/
  we can check for 0/1 in `/sys/module/apparmor/parameters/enabled`
  
BPF LSM:
- Use https://pkg.go.dev/github.com/cilium/ebpf/features to check if we can load LSM programs

Check for other possibilities as well.

This could potentially help solve our issues with Kubernetes in Docker situation since they don't have the file present.

daemon1024:
@Chinwendu20 Would appreciate your help on this 🙌🏽  Thank You.

Chinwendu20:
We could check for kernel config on the host. I found various options for doing that [here](https://superuser.com/questions/287371/obtain-kernel-config-from-currently-running-linux-system). After going through that, I suggest we use this:

` cat /proc/config.gz | gunzip | grep CONFIG_BPF_LSM`

To check if the host supports bpf_lsm. I tried this on minikube vm with VirtualBox driver and it gives this result:

`CONFIG_BPF_LSM=y`

rksharma95:
@daemon1024 is there anything pending here? can we close this?

daemon1024:
AppArmor needs to have securityfs mounted and so we can expect the presence of the security lsm filepath. Closing the issue. Thanks @rksharma95

Issue #903, opened 2022-09-19
Add k8s kind support


Ankurk99:
## Feature Request

**Short Description**
Support KubeArmor on kind (https://kind.sigs.k8s.io/docs/user/quick-start/)

**Is your feature request related to a problem? Please describe the use case.**
Currently enforcement doesn't work on kind

roopeshsn:
Hi! I would like to work on this issue if it's still open!

PR #1017, opened 2022-12-12
bpf,enforcer: remove conditional compilation and add alternate bpflsm availability check

### LINKED ISSUE #756: Minikube Start is failing with latest version of minikube (1.26) and k8s(1.24)
## Minikube start

**General Information**

- Environment (minikube)
- Kernel version (Linux eswar-LEGION 5.13.0-52-generic #59~20.04.1-Ubuntu SMP Thu Jun 16 21:21:28 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux)

**To Reproduce**

1. Goto contributions minikube.
2. Install minikube and virtualbox using scripts mentioned under contributions/minikube dir.
3. Then run minikube with start_minikube.sh script. 

**Observed behavior**
Minikube throws an error and fails to start.
minikube-err

**Expected behavior**
Minikube is expected to start with no issues/errors. 

**Temporary solution**
with minikube start, use the below flag
`--kubernetes-version=1.23.1`

daemon1024:
**Purpose of PR?**:
We now compile everything and let probes loading fail at runtime, the failed probes don't impact functioning of other probes so the added complexity to conditional compile wasn't needed.

Remove dependency for /sys/kernel/security/lsm to check for BPF LSM. This should help things work for Minikube and KIND. Yet to be tested on KIND.

Hopefully closes #756 #903
Ref #928

Yes, New KubeArmor init container won't work with older KubeArmor images

Test in Minikube with KVM2 driver uses custom buildroot iso with BPFLSM support but no CONFIG_SECURITY_PATH

We should enforce stable init container before proceeding with this

Need to update docs once Minikube and KIND support verified.

**Checklist:**
- [x] Bug fix.


KubeArmor/enforcer/runtimeEnforcer.go
```diff
@@ -125,31 +127,44 @@ func NewRuntimeEnforcer(node tp.Node, logger *fd.Feeder) *RuntimeEnforcer {
...
+
+probeBPFLSM:
+	if !kl.ContainsElement(lsms, "bpf") {
+		err := features.HaveProgramType(ebpf.LSM)
+		if err == nil {
+			lsms = append(lsms, "bpf")
+		} else {
+			re.Logger.Warnf("BPF LSM not supported %s", err.Error())
 		}
 	}
```
Ankurk99: > Remove dependency for /sys/kernel/security/lsm to check for BPF LSM. This should help things work for Minikube and KIND. Yet to be tested on KIND.
 
I didn't understand this. Here we are checking if "bpf" string is found in /sys/kernel/security/lsm and in case of minikube this file doesn't exist. How will this work in minikube?

daemon1024: It's a not (`<!>kl.Contains`) check. We skip the following check if we already found bpf in the said file. We should only check for support using this method iff we didn't find the lsm file or the lsm file didn't have bpf.

Does that make sense?

rksharma95:
i think the tests are failing due to test utility function not able to find pods where it's looking for pods
with `apparmor` specific annotations `container.apparmor.security.beta.kubernetes.io/...`. that will be patched to the endpoints in case if enforcer is `apparmor`.

and with current change where we have removed the dependency for `/sys/kernel/security/lsm` to check for `BPF LSM`, and in the lsm-order `bpf` is having higher priority the enforcer used by the `KubeArmor` in test env will be `bpf`.

and it's obvious for the pods to not have `apparmor` specific annotations. i think we need to refactor tests.

daemon1024:
Ohh makes sense. Does that mean bpflsm is supported on Github Runners? Or the code is not apt 😢 

I will just confirm if Github Runners do support BPF LSM. If yes then will just force tests to use apparmor else will need to debug PR.

daemon1024:
There's some discrepancy in the support check. Will have to defer this for later since needs further investigation.

KubeArmor/utils/bpflsmprobe/probe.go
```diff
@@ -0,0 +1,68 @@
...
+	record, err := rd.Read()
+	if err != nil {
+		return err
+	}
+
+	if err := binary.Read(bytes.NewBuffer(record.RawSample), binary.LittleEndian, &event); err != nil {
+		return err
+	}
+
+	return nil
```
nyrahul:
```suggestion
	return binary.Read(bytes.NewBuffer(record.RawSample), binary.LittleEndian, &event)
```

KubeArmor/utils/bpflsmprobe/probe.go
```diff
@@ -0,0 +1,68 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2023 Authors of KubeArmor
+
+package probe
```
nyrahul, line 5:
```suggestion
// probe to check whether the probed LSM support is available.
package probe
```

nyrahul (changes requested):
Two minor comments. Otherwise LGTM 👍

nyrahul (approved):
LGTM 👍

PR #1027, opened 2022-12-23
feat(apparmor): use apparmor as alert source

### LINKED ISSUE #1009: Use auditd log for alerting on apparmor systems
### Short Description

Currently, kubearmor uses eBPF for identifying policy violations i.e., it applies checks at kretprobes for the syscalls of interest and checks the return value. If the return value is "Permission Denied" then the policy matching is done to figure out which policy caused this event.

The advantage with this approach is that it works regardless of the LSM in use.

The problem with this approach is that if the syscall returns Permission Denied due to some different reason which is not-kubearmor related then the policy-matcher would still tag it in the context of kubearmor since kubearmor's policy-matcher has no way of knowing whether the denial is because of its own policy.

### Proposal:

Decouple alerting from ebpf programs by using auditd logs. In case of apparmor, the kernel audit logs are available. These audit logs contain the apparmor profile information that caused the denial event. The profile information can be used to map the namespace, container, pod details.

### Tasks
- [x] use kernel audit using netlink to fetch audit logs
- [x] Handling audit/block events natively in apparmor
- [x] fallback logic if the audit mechanism is not available or based on explicit configuration option
- [x] handling process/file/network and creating telemetry alert based on the audit log
- [x] Add system tests
- [x] when bpf-lsm is used as the enforcer then default back to ebpf. Create a separate issue to handle alert events using bpf-lsm.

achrefbensaad:
**Purpose of PR?**:

Fixes #1009

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/KubeArmor/monitor/systemMonitor.go
+++ b/KubeArmor/monitor/systemMonitor.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"sync"
 
@@ -19,6 +20,8 @@ import (
 	"github.com/cilium/ebpf/link"
 	"github.com/cilium/ebpf/perf"
 	"github.com/cilium/ebpf/rlimit"
+	"github.com/elastic/go-libaudit/v2"
+	"github.com/elastic/go-libaudit/v2/auparse"
 
 	kl "github.com/kubearmor/KubeArmor/KubeArmor/common"
 	cfg "github.com/kubearmor/KubeArmor/KubeArmor/config"
@@ -125,11 +128,14 @@ type SystemMonitor struct {
 	Status          bool
 	UptimeTimeStamp float64
 	HostByteOrder   binary.ByteOrder
+
+	// Netlink client
+	NetLinkClient *libaudit.AuditClient
 }
 
 // NewSystemMonitor Function
 func NewSystemMonitor(node *tp.Node, logger *fd.Feeder, containers *map[string]tp.Container, containersLock **sync.RWMutex,
-	activeHostPidMap *map[string]tp.PidMap, activePidMapLock **sync.RWMutex) *SystemMonitor {
+	activeHostPidMap *map[string]tp.PidMap, activePidMapLock **sync.RWMutex, enforcer string) *SystemMonitor {
 	mon := new(SystemMonitor)
 
 	mon.Node = node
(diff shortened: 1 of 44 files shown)


KubeArmor/monitor/systemMonitor.go
```diff
@@ -152,9 +158,138 @@ func NewSystemMonitor(node *tp.Node, logger *fd.Feeder, containers *map[string]t
 	mon.UptimeTimeStamp = kl.GetUptimeTimestamp()
 	mon.HostByteOrder = binary.LittleEndian
 
+	mon.NetLinkClient = mon.InitNetfilter()
```
nyrahul: This should not be inited in all the cases .. for eg. if there is BPF_LSM enable.

KubeArmor/monitor/systemMonitor.go
```diff
@@ -152,9 +158,138 @@ func NewSystemMonitor(node *tp.Node, logger *fd.Feeder, containers *map[string]t
...
+			ActivePidMapLock.Lock()
+			a, _ := strconv.Atoi(match[0][3])
+			for cID, pidInfor := range ActiveHostPidMap {
+				if elem, ok := pidInfor[uint32(a)]; ok {
+					go mon.BuildAppArmorLogBase(cID, elem, match[0])
+					break
+				}
+			}
+			ActivePidMapLock.Unlock()
+		}
```
nyrahul: Can we put the lock/unlock outside the if-else-if chain?
```suggestion

		ActivePidMapLock.Lock()
		if match := profilergx.FindAllStringSubmatch(string(rawEvent.Data), -1); len(match) != 0 {
			a, _ := strconv.Atoi(match[0][4])
			for cID, pidInfor := range ActiveHostPidMap {
				if elem, ok := pidInfor[uint32(a)]; ok {
					go mon.BuildAppArmorLogBase(cID, elem, match[0])
					break
				}
			}
		} else if match := netregx.FindAllStringSubmatch(string(rawEvent.Data), -1); len(match) != 0 {
			a, _ := strconv.Atoi(match[0][3])
			for cID, pidInfor := range ActiveHostPidMap {
				if elem, ok := pidInfor[uint32(a)]; ok {
					go mon.BuildAppArmorLogBase(cID, elem, match[0])
					break
				}
			}
		} else if match := capregx.FindAllStringSubmatch(string(rawEvent.Data), -1); len(match) != 0 {
			a, _ := strconv.Atoi(match[0][3])
			for cID, pidInfor := range ActiveHostPidMap {
				if elem, ok := pidInfor[uint32(a)]; ok {
					go mon.BuildAppArmorLogBase(cID, elem, match[0])
					break
				}
			}
		}
		ActivePidMapLock.Unlock()
```

nyrahul (dismissed):

nyrahul (approved):
LGTM 👍

PR #1074, opened 2023-01-26
Update maintainers list

Ankurk99:
I have updated the maintainers' list based on the KubeArmor community call. 
PS: This might not be the complete list.


nyrahul (approved):

Ankurk99:
Sorry, I just noticed it today that Achref's name is down the list, but according to ascending order it should be at the top.
Fixed: https://github.com/kubearmor/KubeArmor/pull/1132

PR #1097, opened 2023-02-07
enforcer(bpf): automatically mount bpffs

### LINKED ISSUE #1005: Amazon Linux 2 Support for BPF-LSM
I have AL2 EKS worker nodes running on Kernel version 5.15.  Do you have any guidance or documentation on enabling BPF-LSM in kubearmor for AL2 nodes?

The active LSM is blank and all the Postures are listed as "audit" mode.
Screen Shot 2022-11-28 at 10 58 30 AM

DelusionalOptimist:
Signed-off-by: Rudraksh Pareek <rudraksh@accuknox.com>**Purpose of PR?**:

May fix #1005 

Nope

* Tested on k3s using kubearmor-test-k3s.yaml
	* bpffs is not mounted on the host (`umount /sys/fs/bpf`) - kubearmor-init now mounts it when deployed
	* bpffs is mounted on the host - no changes

Though `SYS_ADMIN` capability should be enough for mounting filesystems but bpffs needs the container to be privileged.

**Checklist:**
- [ ] Commit has unit tests


KubeArmor/build/compile.sh
```diff
@@ -11,4 +11,13 @@ else
...
+cp *.bpf.o ignore.lst /opt/kubearmor/BPF/
+
+# checking and mounting bpffs if it isn't
+if mountpoint -q /sys/fs/bpf;
+then
+	echo "bpffs already mounted";
+else
+	 echo "bpffs not mounted. Mounting now.";
+	 mount -t bpf bpffs /sys/fs/bpf
+fi
```
daemon1024: Can we also add the check in the main container? We might be eliminating init container in the future.

deployments/get/objects.go
```diff
@@ -550,7 +552,7 @@ func GenerateDaemonSet(env, namespace string) *appsv1.DaemonSet {
 							Name:  "init",
 							Image: "kubearmor/kubearmor-init:latest",
 							SecurityContext: &corev1.SecurityContext{
-								Privileged: &privileged,
+								Privileged: &kubearmorInitPrivileged,
```
daemon1024: Is this necessary? Can't we bypass this with some specific capability add?

daemon1024 (changes requested):

KubeArmor/build/compile.sh
DelusionalOptimist: Gotcha :+1:

deployments/get/objects.go
DelusionalOptimist: Yeah normally `SYS_ADMIN` should suffice for mounting filesystems but bpffs is an exception ig. (based on my tries with both privileged true and false, couldn't find a source stating this yet)

KubeArmor/build/compile.sh
DelusionalOptimist: But the caveat mentioned in the other comment. We'll need to make it privileged.

deployments/get/objects.go
daemon1024: I don't think we will go ahead with this, if it requires privileged flag.

daemon1024: Cilium also doesn't depend on it, so can you please verify things again? Thanks.

DelusionalOptimist: I'll look into Cilium as well. Calico seemed to be doing the same - https://github.com/projectcalico/calico/blob/a8bdcd60f99f1ab3b0495b7b4fe4d1d3fb719c35/manifests/calico.yaml#L4526-L4547 :thinking:

DelusionalOptimist:
Update - Cilium too uses a privileged init [Mount-bpf](https://github.com/cilium/cilium/blob/666100d96f89afde0d9fb05130997519a5407f87/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml#L532-L555) container for bidirectional mounting and the [cilium-agent](https://github.com/cilium/cilium/blob/666100d96f89afde0d9fb05130997519a5407f87/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml#L289-L298) is unprivileged by default but may be privileged if specified through helm values. Mount-bpf does the mounting work when Cilium is running in K8s.

However the logic we've implemented now would create `/sys/fs/bpf` in kubearmor's container if it wasn't mounted on host and mount the bpffs in it. The host would remain undisturbed and we could bypass the need for privileged container.
Also, @daemon1024 pointed out that we don't really need persistence of bpf maps so even if kubearmor pod was killed, the policies will remain intact and be restored on restart (we also tried this out).
In case Kubearmor is running outside K8s, it would mount bpffs itself if not already mounted.

cc @nyrahul @kranurag7

KubeArmor/enforcer/bpflsm/bpffs.go
```diff
@@ -0,0 +1,222 @@
+package bpflsm
```
nyrahul, line 4: License header is not present.

Please add credits for Cilium folks, since this code is heavily inspired by it.

Wondering if we could just import the specific package from cilium without having to cut-paste the code.

nyrahul (changes requested):
Mostly LGTM. Please check for review comment inline.

KubeArmor/enforcer/bpflsm/bpffs.go
DelusionalOptimist: > Please add credits for Cilium folks, since this code is heavily inspired by it.

Yep, was going to do that along with the gosec changes.
> Wondering if we could just import the specific package from cilium without having to cut-paste the code.

They have some logs and paths put in according to Cilium, so I don't think that we can import directly : (

KubeArmor/enforcer/bpflsm/bpffs.go
```diff
@@ -0,0 +1,228 @@
...
+	// we don't want to change the path on which maps are stored to
+	// be changed once we start writing maps
+	if lockedDown {
+		panic("setMapRoot() call after MapRoot was read")
+	}
+	mapRoot = path
+}
+
+// returns the mapRoot
+func GetMapRoot() string {
```
daemon1024, line 225: Let's handle the linter warning here

daemon1024 (commented):
LGTM, just one minor nit

KubeArmor/enforcer/bpflsm/bpffs.go
DelusionalOptimist: :+1:

achrefbensaad (approved):
LGTM!. Needs to be tested on AZL2 before merging.

nyrahul (approved):
LGTM

DelusionalOptimist:
Tested on AL2. Good to merge.

daemon1024 (approved):

PR #1124, opened 2023-02-22
fix(monitor(bpf)): store full path from path_mknod

### LINKED ISSUE #1118: Relative Paths instead of Absolute Paths in Telemetry Events
## Bug Report

We are receiving Relative Paths instead of Absolute Paths in Telemetry Events. This is causing issues with our Policy Matcher resulting in missing Audit Alerts and `DefaultPosture` in Permission Denied Alerts.

**To Reproduce**

1. Check File telemetry events 

**Expected behavior**

TBA

**Screenshots**
TBA

daemon1024:
**Purpose of PR?**:

Fixes #1118
We leverage security hooks to find out full paths for resources. With file creation related operations, if an enforcement is in place our security hook would never be called since the execution is blocked by a security hook upper in the stack. This commit adds an additional hook to store the full path information which should help us handle cases where `security_file_open` is never called. 

No

Check touch based Alerts on denials

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/KubeArmor/BPF/system_monitor.c
+++ b/KubeArmor/BPF/system_monitor.c
@@ -889,6 +889,13 @@ static __always_inline int security_path_task_arg(struct pt_regs *ctx){
 }
 
 #if defined(SECURITY_PATH)
+SEC("kprobe/security_path_mknod")
+int kprobe__security_path_mknod(struct pt_regs *ctx){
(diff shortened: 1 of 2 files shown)


rksharma95 (approved):
LGTM :+1:

PR #1131, opened 2023-03-01
fix(feature): container detection issue in containerd

### LINKED ISSUE #1058: Container detection logic: is running infinitely for the same container
## Bug Report

**General Information**

- Environment description (GKE, VM-Kubeadm, vagrant-dev-env, minikube, microk8s, ...) any
- Kernel version (run `uname -a`) any
- Orchestration system version in use (e.g. `kubectl version`, ...) any
- Link to relevant artifacts (policies, deployments scripts, ...) any
- Target containers/pods

**To Reproduce**

add this line in the `UpdateContainerdContainer` function under `KubeArmor/core/containerdHandler.go`
```
fmt.Printf("Container %s update was called with %s action\n", containerID, action)
```

**Expected behavior**

once a container has being added with a perticular state, it should not be re-added with the same state.

**Screenshots**

image

Bug reported by: @seswarrajan

Ankurk99:
**Purpose of PR?**:

Fixes #1058 

Tested on k3s version v1.23.9+k3s1, containerd github.com/containerd/containerd v1.6.10

Still needs to be tested on CRI-O runtime - @DelusionalOptimist will be taking a look

**Checklist:**
- [x] Bug fix. Fixes #1058 

### DIFF
--- a/KubeArmor/monitor/systemMonitor.go
+++ b/KubeArmor/monitor/systemMonitor.go
@@ -477,12 +477,12 @@ func (mon *SystemMonitor) InitBPF() error {
 			}
 			mon.Probes["kprobe__"+syscallName], err = link.Kprobe("sys_"+syscallName, mon.BpfModule.Programs["kprobe__"+syscallName], nil)
 			if err != nil {
-				return fmt.Errorf("error loading kprobe %s: %v", syscallName, err)
+				mon.Logger.Printf("error loading kprobe %s: %v", syscallName, err)
 			}
 
 			mon.Probes["kretprobe__"+syscallName], err = link.Kretprobe("sys_"+syscallName, mon.BpfModule.Programs["kretprobe__"+syscallName], nil)
 			if err != nil {
-				return fmt.Errorf("error loading kretprobe %s: %v", syscallName, err)
+				mon.Logger.Printf("error loading kretprobe %s: %v", syscallName, err)
 			}
 
 		}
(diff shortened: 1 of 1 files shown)


KubeArmor/core/crioHandler.go
```diff
@@ -312,17 +312,16 @@ func (dm *KubeArmorDaemon) MonitorCrioEvents() {
 
 	dm.Logger.Print("Started to monitor CRI-O events")
 
+	containers, err := Crio.GetCrioContainers()
```
DelusionalOptimist: @Ankurk99 I think if we move this here, we would detect containers only once when kubearmor starts. After that, we won't get any new containers detected and added to the monitor. Also, same would happen with containerd too.

Ankurk99: :thinking: What my understanding is that we get all the present crio containers at the beginning and then whenever there is a new container added we are already calling UpdateContainer with start action.
https://github.com/kubearmor/KubeArmor/blob/main/KubeArmor/core/crioHandler.go#L340
Does that makes sense?

DelusionalOptimist: I think currently [UpdateCrioContainer()](https://github.com/kubearmor/KubeArmor/blob/a28c6fbd124c306a4fcc481913edc171fcc6e860/KubeArmor/core/crioHandler.go#L199) takes the container ID and gets info about that particular container from the CRI-API/Containerd API, then adds it to the container map.
The only function which gets all currently present containers is [GetCrioContainer()](https://github.com/kubearmor/KubeArmor/blob/a28c6fbd124c306a4fcc481913edc171fcc6e860/KubeArmor/core/crioHandler.go#L152). It will return all present containers and then we compare it with the container map in [CrioHandler](https://github.com/kubearmor/KubeArmor/blob/a28c6fbd124c306a4fcc481913edc171fcc6e860/KubeArmor/core/crioHandler.go#L32).
If both match, we didn't have any new containers and we continue, if the size of the returned map is greater, we call UpdateCrioContainer() to add it to this map .

PR #1133, opened 2023-03-06
fix(tests): Use enforcer-agnostic annotation in tests

### LINKED ISSUE #1129: ginkgo tests should work with BPF LSM
**Short Description**

We use ginkgo for running tests locally as well as in CI. However, currently they depend on Apparmor annotations of the test workload for validation.
We should modify the tests accordingly so that they work with BPF LSM as well.

rksharma95:
**Purpose of PR?**:
At this point the ginkgo tests are implemented in such a way that they will work only if current enforcer is AppArmor due to the dependency on AppArmor annotations. Merging this PR will enable the tests to work with bpflsm enforcer too.

_As GH action doesn't support enabling bpflsm for now this fix will enable users do testing locally with KubeArmor running as system process._
Fixes #1129 

No
tested for both AppArmor and BPFLSM enforcer, able to run the tests in both scenarios.

**Checklist:**

### DIFF
--- a/tests/block/block_test.go
+++ b/tests/block/block_test.go
@@ -54,10 +54,7 @@ var _ = Describe("Posture", func() {
 	// var sql string
 
 	BeforeEach(func() {
-		wp = getWpsqlPod("wordpress-",
-			"container.apparmor.security.beta.kubernetes.io/wordpress: localhost/kubearmor-wordpress-mysql-wordpress-wordpress")
-		// sql = getWpsqlPod("mysql-",
-		// 	"container.apparmor.security.beta.kubernetes.io/mysql: localhost/kubearmor-wordpress-mysql-mysql-mysql")
+		wp = getWpsqlPod("wordpress-", "kubearmor-policy: enabled")
 	})
 
 	AfterEach(func() {
--- a/tests/configmap/kubearmor_config_test.go
+++ b/tests/configmap/kubearmor_config_test.go
@@ -64,9 +64,9 @@ var _ = Describe("KubeArmor-Config", func() {
 	var partialyAnnotated string
 	var fullyAnnotated string
 	BeforeEach(func() {
-		unannotated = getUnannotatedPod("unannotated-", "container.apparmor.security.beta.kubernetes.io/ubuntu-1: localhost/kubearmor-unannotated-unannotated-deployment-ubuntu-1")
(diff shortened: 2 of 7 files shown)


DelusionalOptimist (commented):
Thank you for working on this @rksharma95. Looks good to me. :rocket:
Just have to fix the gosec issue.

DecFox:
Hey @rksharma95, is there a specific reason why the `WriteToFile` permissions have been elevated here? (since it is the reason for the failing go-sec)

rksharma95:
> Hey @rksharma95, is there a specific reason why the `WriteToFile` permissions have been elevated here? (since it is the reason for the failing go-sec)

to read current enforcer info from the probe data, as tests are allowed to run as non-root user, in fact we are running the tests as non-root user at present.

tests/util/kartutil.go
```diff
@@ -295,10 +300,41 @@ func K8sExecInPodWithContainer(pod string, ns string, container string, cmd []st
 	return buf.String(), errBuf.String(), nil
 }
 
+func isBpfLsmEnforcer() (bool, error) {
+	if _, err := os.Stat("/tmp/karmorProbeData.cfg"); err != nil {
+		log.Errorf("unable to karmorProbeData.cfg: %s", err)
+		return false, nil
```
daemon1024: We don't use host process mode but deploy kubearmor on k8s in tests. Not sure if this would work. We should add the check to extract as well maybe?

Aryan-sharma11:
Is this PR ready to be merged? @rksharma95

rksharma95:
> Is this PR ready to be merged? @rksharma95

no, there's still some pending task.

rksharma95:
rather than checking enforcer specific annotations, can we check of some enforcer-agnostic annotations i.e. `kubearmor-policy: enabled` to ensure that the pod has been processed by the kubearmor. 

any thoughts here? @daemon1024

daemon1024 (approved):
LGTM

DelusionalOptimist (commented):
@rksharma95 I think we missed the visibility test suite. :eyes:

DelusionalOptimist (approved):
LGTM. Thanks sticking with this @rksharma95 : )

DelusionalOptimist:
Also, thank you @DecFox, @Aryan-sharma11 for probing into this PR. :raised_hands:

PR #1137, opened 2023-03-07
manifest(deploy): do not enable Host Policy Enforcement by default

daemon1024:
**Purpose of PR?**:
Enabling KubeArmor Host policy increases the visibility and load on KubeArmor by a lot and most of the managed providers don't need host policy enforcement, so disabling it by default


DelusionalOptimist (approved):
CITY : )

PR #1145, opened 2023-03-07
manifest: change image pull policy for init container

daemon1024:
This will help us update images when stable tag is updated. Will introduce a flag in kArmor to use local images as well.


Ankurk99 (approved):
LGTM!

PR #1150, opened 2023-03-10
Set default GRPC listening port for kubearmor as environment variable…

### LINKED ISSUE #275: Systemd Mode Karmor logs producing error
***Bug Report***

**General Information**

Folks just working on adding some systemd tests and following the installation in bare metal policy sleep deny but when I try to get the log I get the below error. Seems looking for cluster but this is not suppose to happen as the installation is in systemd mode for VM policies.

**To Reproduce**

$ karmor log --json
Error: unknown command “log” for “karmor”
Did you mean this?
logs

And when I use Karmor logs I get the below

$ karmor logs --json
Error: Get “http://localhost:8080/api/v1/pods?labelSelector=kubearmor- app%3Dkubearmor-relay”: dial tcp 127.0.0.1:8080: connect: connection refused

it’s looking for a cluster on bare metal which would not exist if we are doing systemd mode policy implementation.

**Expected behavior**

$ karmor logs --json

{
"Timestamp": 1639803960,
"UpdatedTime": "2021-12-18T05:06:00.077564Z",
"ClusterName": "Default",
"HostName": "pandora",
"HostPID": 3390423,
"PPID": 168556,
"PID": 3390423,
"UID": 1000,
"PolicyName": "hsp-kubearmor-dev-proc-path-block",
"Severity": "1",
"Type": "MatchedHostPolicy",
"Source": "zsh",
"Operation": "Process",
"Resource": "/usr/bin/sleep",
"Data": "syscall=SYS_EXECVE",
"Action": "Block",
"Result": "Permission denied"
}

Chinwendu20:
**Purpose of PR?**:

Fixes [##275](https://github.com/kubearmor/kubearmor-client/issues/275)

### DIFF
--- a/KubeArmor/packaging/post-install.sh
+++ b/KubeArmor/packaging/post-install.sh
@@ -22,4 +22,4 @@ fi
 
 # Set default GRPC listening port for kubearmor as environment variable
 echo "export KUBEARMOR_SERVICE=:32767" >> ~/.bashrc
-source ~/.bashrc
+. ~/.bashrc
--- a/KubeArmor/packaging/post-install.sh
+++ b/KubeArmor/packaging/post-install.sh
@@ -19,3 +19,7 @@ fi
 # start kubearmor.service
 /bin/systemctl daemon-reload
 /bin/systemctl start kubearmor.service
+
+# Set default GRPC listening port for kubearmor as environment variable
+echo "export KUBEARMOR_SERVICE=:32767" >> ~/.bashrc
+source ~/.bashrc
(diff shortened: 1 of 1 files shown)


KubeArmor/packaging/post-install.sh
```diff
@@ -19,3 +19,7 @@ fi
 # start kubearmor.service
 /bin/systemctl daemon-reload
 /bin/systemctl start kubearmor.service
+
+# Set default GRPC listening port for kubearmor as environment variable
+echo "KUBEARMOR_SERVICE=:32767" >> /etc/environment
+. /etc/environment
```
nyrahul: sourcing of this environment file won't have effect on the env outside of the script. User still has to logout/login for changes to come into effect.

Chinwendu20: Thanks a lot. I have updated the script

PR #1187, opened 2023-03-27
fix: res.Container.ID[:12] lead to panic

### LINKED ISSUE #1179: Index out of bound on containers with short names
## Bug Report

I've spawned kuberarmor on a kubernetes cluster, but noticed one pod constantly dying with the error:
```
panic: runtime error: slice bounds out of range 
```
After some hours of despair while debugging I found out that a container that we spawned on the machine manually called "htop" caused that. 
https://github.com/kubearmor/KubeArmor/blob/main/KubeArmor/core/containerdHandler.go#L135 causes kubearmor to fail on containers with a name/id less then 12 chars. Not sure if this is intended by design, but it should probably not panic because of this.

**General Information**

- k3s, containered
- Kernel 5.15.77
- kubernetes 1.25

**To Reproduce**

1. create containers with fixed Id's under 12 chars. e.g `ctr run --rm -d docker.io/library/debian:latest cont1`
2. run kubearmor on kubernetes
3. see logs for the node running the created container

**Expected behavior**

Not to fail on a `slice out of bounds` or at least not to crash the entire KubeArmor pod.

xiao-jay:
**Purpose of PR?**:

Fixes #1179 

No

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/KubeArmor/core/containerdHandler.go
+++ b/KubeArmor/core/containerdHandler.go
@@ -132,8 +132,11 @@ func (ch *ContainerdHandler) GetContainerInfo(ctx context.Context, containerID s
 	// == container base == //
 
 	container.ContainerID = res.Container.ID
-	container.ContainerName = res.Container.ID[:12]
-
+	if len(res.Container.ID) < 12 {
+		container.ContainerName = res.Container.ID
+	} else {
+		container.ContainerName = res.Container.ID[:12]
+	}
 	container.NamespaceName = "Unknown"
 	container.EndPointName = "Unknown"
 
@@ -318,7 +321,7 @@ func (dm *KubeArmorDaemon) UpdateContainerdContainer(ctx context.Context, contai
 			dm.RuntimeEnforcer.RegisterContainer(containerID, container.PidNS, container.MntNS)
 		}
 
-		dm.Logger.Printf("Detected a container (added/%s/pidns=%d/mntns=%d)", containerID[:12], container.PidNS, container.MntNS)
+		dm.Logger.Printf("Detected a container (added/%s/pidns=%d/mntns=%d)", container.ContainerName, container.PidNS, container.MntNS)
 
 	} else if action == "destroy" {
 		dm.ContainersLock.Lock()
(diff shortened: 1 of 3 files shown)


Ankurk99 (approved):
LGTM!

xiao-jay:
@Ankurk99 I need some time to install flatcar caontainer os to verify I truly solve this problem.

Ankurk99:
@xiao-jay It's okay, we don't actually need to reproduce the issue. 
But @daemon1024 have some questions regarding the PR.

KubeArmor/core/containerdHandler.go
```diff
@@ -132,8 +132,11 @@ func (ch *ContainerdHandler) GetContainerInfo(ctx context.Context, containerID s
 	// == container base == //
 
 	container.ContainerID = res.Container.ID
-	container.ContainerName = res.Container.ID[:12]
-
+	if len(res.Container.ID) < 12 {
+		container.ContainerName = res.Container.ID
+	} else {
+		container.ContainerName = res.Container.ID[:12]
+	}
```
daemon1024: Let's not do this.
Let's keep it
```suggestion
	container.ContainerName = res.Container.ID
```

This is because we are only trimming to so that we have a saner logging mechanism. I will comment in the next review what we can do to avoid that.

KubeArmor/core/containerdHandler.go
```diff
@@ -318,7 +321,7 @@ func (dm *KubeArmorDaemon) UpdateContainerdContainer(ctx context.Context, contai
 			dm.RuntimeEnforcer.RegisterContainer(containerID, container.PidNS, container.MntNS)
 		}
 
-		dm.Logger.Printf("Detected a container (added/%s/pidns=%d/mntns=%d)", containerID[:12], container.PidNS, container.MntNS)
+		dm.Logger.Printf("Detected a container (added/%s/pidns=%d/mntns=%d)", container.ContainerName, container.PidNS, container.MntNS)
```
daemon1024:
```suggestion
		dm.Logger.Printf("Detected a container (added/%.12s/pidns=%d/mntns=%d)", containerID, container.PidNS, container.MntNS)
```

Notice the `%.12s` that will limit the string size to 12 and will not panic otherwise even if the length is shorter.

daemon1024 (changes requested):
Thanks a lot for the PR, I have suggested for a different way to fix this issue. Please find it inline and lemme know WDYT?

KubeArmor/core/containerdHandler.go
xiao-jay: ok, I think is nice, @daemon1024 I have change code to your suggest.

DelusionalOptimist (approved):
Thanks @xiao-jay!
LGTM. :rocket:

daemon1024 (approved):

PR #1198, opened 2023-04-14
fix: continue even if syscall not found

Ankurk99:
**Description**
In the event that a syscall is not found, our previous implementation would result in a return statement, causing KubeArmor to crash. To address this issue, this pull request resolves the problem by handling such cases with an error message and allowing the code execution to continue uninterrupted.

**Problem**
The problem was faced on:
**OS**: Red Hat Enterprise Linux CoreOS 410.84.202211032138-0 (Ootpa)
**Container Runtime**: cri-o://1.23.3-20.rhaos4.10.git89344de.el8
**Kernel Version**: 4.18.0-305.65.1.el8_4.x86_64

**Tested**
This change is locally tested (k3s, containerd, 5.15.81-1-MANJARO) with latest KubeArmor.

**Are there any further changes required?**
Yes, we will need to make this change Backward-compatible. That will be handled in a separate PR.

### DIFF
--- a/KubeArmor/monitor/systemMonitor.go
+++ b/KubeArmor/monitor/systemMonitor.go
@@ -477,12 +477,12 @@ func (mon *SystemMonitor) InitBPF() error {
 			}
 			mon.Probes["kprobe__"+syscallName], err = link.Kprobe("sys_"+syscallName, mon.BpfModule.Programs["kprobe__"+syscallName], nil)
 			if err != nil {
-				return fmt.Errorf("error loading kprobe %s: %v", syscallName, err)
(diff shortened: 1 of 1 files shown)


achrefbensaad (approved):
LGTM!, Lets merge once we get a green pipeline.

We should have an issue to delete the syscall compatibility logic from the init container as it is rendered obselete with this change.

nyrahul (approved):
👍

daemon1024 (approved):
Need to follow up with removal of tests in init container, but that will be a breaking change so let's handle it in a seperate PR. Ref https://github.com/kubearmor/KubeArmor/pull/1017

PR #1199, opened 2023-04-14
(backport) fix: continue even if syscall not found

Ankurk99:
:warning: Backport fixes from: https://github.com/kubearmor/KubeArmor/pull/1198

**Description** 
 In the event that a syscall is not found, our previous implementation would result in a return statement, causing KubeArmor to crash. To address this issue, this pull request resolves the problem by handling such cases with an error message and allowing the code execution to continue uninterrupted.


nyrahul (approved):
LGTM

PR #1135, opened 2023-03-06
adding `DeploymentName` for alerts metadata

Prateeknandle:
**Purpose of PR?**:

Fixes #1065 

**Description:**
Adding owner information of the pod for alerts metadata

**Does this PR introduce a breaking change?**
No

**If the changes in this PR are manually verified, list down the scenarios covered:**:
tested for each controller - deployments, replicaset, statefulset, daemonset

**Output:**
```
{
  "Timestamp": 1679588263,
  "UpdatedTime": "2023-03-23T16:17:43.146109Z",
  "ClusterName": "default",
  "HostName": "prateek-lenovo-ideapad-310-15isk",
  "NamespaceName": "default",
  "Owner": {
    "Ref": "DaemonSet",
    "Name": "my-daemonset",
    "Namespace": "default"
  },
  "PodName": "my-daemonset-xhrr9",
  "Labels": "app=my-app",
  "ContainerID": "0e6a98ec00521ed92fd29664ae238f60463a93b819e6c54a4494cf371e63e86a",
  "ContainerName": "my-container",
  "ContainerImage": "docker.io/library/nginx:latest@sha256:1ed4dff2b99011798f2c228667d7cb4f4e2bd76b2adc78fd881d39f923e78c9d",
  "HostPPID": 671928,
  "HostPID": 672009,
  "PPID": 41,
  "PID": 46,
  "ParentProcessName": "/bin/bash",
  "ProcessName": "/bin/sleep",
  "PolicyName": "new",
  "Severity": "5",
  "Message": "block /bin/sleep",
  "Type": "MatchedPolicy",
  "Source": "/bin/bash",
  "Operation": "Process",
  "Resource": "/bin/sleep",
  "Data": "syscall=SYS_EXECVE",
  "Enforcer": "AppArmor",
  "Action": "Block",
  "Result": "Permission denied"
}
```

PR #1202, opened 2023-04-19
Annotation issue

### LINKED ISSUE #1180: pods deployed outside of deployments (such as statefulsets, replicasets) are not annotated on KubeArmor install
## Bug Report

When KubeArmor is installed, it applies apparmor annotation (wherever applicable) on pods that are deployed through deployments but not for pods that are deployed using statefulsets, replicasets or anything else. Note that if a pod is deployed (through statefulsets, replicaset or directly) after the kubearmor is installed then the annotations are applied normally.

**General Information**

- Environment description (GKE, VM-Kubeadm, vagrant-dev-env, minikube, microk8s, ...), any
- Kernel version (run `uname -a`), any
- Orchestration system version in use (e.g. `kubectl version`, ...), any
- Link to relevant artifacts (policies, deployments scripts, ...), any
- Enforcer: apparmor
- Target containers/pods.

**To Reproduce**

1. Create a statefulset
2. Install kubearmor

**Expected behavior**

Statefulset pods should be restarted and annotated with apparmor annotations

**Workaround?**
Restarting the statefulsets, replicasets or deleting the pods will work. The new pods will have the annotations applied.

Shreyas220:
**Purpose of PR?**:

When kubeArmor is installed it doesnt add annotation to already present replicaset and statefulset

Fixes #1180 

### 1) Deployment Name 
to get the deployment name we were using this `pod.Metadata["deploymentName"]` 
but the field  `deploymentName` doesn't exist for stateful set and replicaset 

#### Solution 
we check the **Kind** of pod and manually update the `pod.Metadata["deploymentName"]`  field for replicaset and statefulset 

### 2) Updating container list 
To update the container list we used this 

```
deploy, err := K8s.K8sClient.AppsV1().Deployments(pod.Metadata["namespaceName"]).Get(context.Background(), deploymentName, metav1.GetOptions{})

```
we cannot get replicaset and stateful set using this
logic is we loop through the containers and see if we annotations for that and update `updateAppArmor` variable as true which is required to go into `PatchResourceWithAppArmorAnnotations()` method 

#### Solution 
We again check for the **kind** of pod and use appropriate method to Get the the replicaset or statefulset 

### 3) In `PatchDeploymentWithAppArmorAnnotations` Method 

In this method we use this to patch the annotations but it only for deployments hence not updating replicaset and stateful set 

```
	_, err := kh.K8sClient.AppsV1().Deployments(namespaceName).Patch(context.Background(), deploymentName, types.StrategicMergePatchType, []byte(spec), metav1.PatchOptions{})
	if err != nil {
		return err
	}

``` 

#### Solution 

We send Kind of deployment and again check use the resource appropriate method to patch the annotations 

 

## Additional information for reviewers 

Using Patch method to update annotation restarts the deployments and statefulset , 
but by default patching doesn't restart the replicaset 

#### 1) Delete the replicaset (X)
we **do not** go with deleting the replicaset as it often permanently deletes the pod


KubeArmor/core/kubeUpdate.go
```diff
@@ -541,12 +541,18 @@ func (dm *KubeArmorDaemon) WatchK8sPods() {
 				if ownerRef != nil {
 					podOwnerName = ownerRef.Name
 					if ownerRef.Kind == "ReplicaSet" {
-						deploymentName := K8s.GetDeploymentNameControllingReplicaSet(pod.Metadata["namespaceName"], podOwnerName)
+						deploymentName := ownerRef.Name //K8s.GetDeploymentNameControllingReplicaSet(pod.Metadata["namespaceName"], podOwnerName)
```
Shreyas220: @DelusionalOptimist need you review here,
we need `ownerRef.Name` here  for `K8s.K8sClient.AppsV1().ReplicaSets(pod.Metadata["namespaceName"]).Get(context.Background(), deploymentName, metav1.GetOptions{})`

didnt see any errors while testibg but wanted to be sure before removing it

KubeArmor/core/k8sHandler.go
```diff
@@ -248,6 +248,46 @@ func (kh *K8sHandler) PatchDeploymentWithAppArmorAnnotations(namespaceName, depl
...
 
+	if kind == "StatefulSet" {
+		fmt.Println("stateful set found ", deploymentName)
+		_, err := kh.K8sClient.AppsV1().StatefulSets(namespaceName).Patch(context.Background(), deploymentName, types.StrategicMergePatchType, []byte(spec), metav1.PatchOptions{})
+		if err != nil {
+			return err
+		} else {
+			return nil
+		}
+
```
nyrahul:
```suggestion
		if err != nil {
			return err
		}
		return nil
```

KubeArmor/core/k8sHandler.go
```diff
@@ -248,6 +248,46 @@ func (kh *K8sHandler) PatchDeploymentWithAppArmorAnnotations(namespaceName, depl
...
+			return err
+		}
+		replicas := *rs.Spec.Replicas
+		fmt.Println("Replica set found ", deploymentName, "replicaas are ", replicas)
+		_, err = kh.K8sClient.AppsV1().ReplicaSets(namespaceName).Patch(context.Background(), deploymentName, types.MergePatchType, []byte(spec), metav1.PatchOptions{})
+		if err != nil {
+			return err
+		}
+		fmt.Println("Patching replicaset")
+		time.Sleep(1 * time.Second)
```
nyrahul: why is the sleep needed here?

KubeArmor/core/k8sHandler.go
```diff
@@ -248,6 +248,46 @@ func (kh *K8sHandler) PatchDeploymentWithAppArmorAnnotations(namespaceName, depl
...
+
+		// To update the annotations we need to restart the replicaset,we scale it down and scale it back up
+		patchData := []byte(fmt.Sprintf(`{"spec": {"replicas": 0}}`))
+		_, err = kh.K8sClient.AppsV1().ReplicaSets(namespaceName).Patch(context.Background(), deploymentName, types.StrategicMergePatchType, patchData, metav1.PatchOptions{})
+		if err != nil {
+			fmt.Printf("Error scaling up ReplicaSet: %v\n", err)
```
nyrahul: Lets use appropriate primitives to report error.

KubeArmor/core/k8sHandler.go
```diff
@@ -248,6 +248,46 @@ func (kh *K8sHandler) PatchDeploymentWithAppArmorAnnotations(namespaceName, depl
...
+			os.Exit(1)
+		}
+		time.Sleep(2 * time.Second)
+		patchData2 := []byte(fmt.Sprintf(`{"spec": {"replicas": %d}}`, replicas))
+		_, err = kh.K8sClient.AppsV1().ReplicaSets(namespaceName).Patch(context.Background(), deploymentName, types.StrategicMergePatchType, patchData2, metav1.PatchOptions{})
+		if err != nil {
+			fmt.Printf("Error scaling up ReplicaSet: %v\n", err)
+			os.Exit(1)
```
nyrahul: We should never return exit. Failures should not impact the full daemonset.

nyrahul (changes requested):

KubeArmor/core/k8sHandler.go
Shreyas220: leftovers from testing, forgot to remove them

KubeArmor/core/k8sHandler.go
Shreyas220: leftovers from testing, forgot to remove them

KubeArmor/core/kubeUpdate.go
DelusionalOptimist: @Shreyas220 if I understand correctly, this pod ownerRef would be a replica set. We actually need the deployment name, which is the owner of the replicaset itself. Thus we have this in place.
The profile naming convention we have current goes like: `kubearmor-namespace-deployment-container` and this would change it.

Shreyas220: so `K8s.GetDeploymentNameControllingReplicaSet()`  made sure we use the naming convention ?

The deployment name returned from `K8s.GetDeploymentNameControllingReplicaSet()` connot be used to perform .GET() and .Patch() operations

DelusionalOptimist: > so K8s.GetDeploymentNameControllingReplicaSet() made sure we use the naming convention ?

Yes, the deployment name returned by `K8s.GetDeploymentNameControllingReplicaSet()` is be used to create apparmor profiles.

> The deployment name returned from K8s.GetDeploymentNameControllingReplicaSet() connot be used to perform .GET() and .Patch() operations

Not sure what do you mean :thinking:.
We have always used it to patch the deployments earlier in the method `PatchDeploymentWithAppArmorAnnotations`. In fact, I think you won't be able to patch the deployment if you use the ReplicaSet name.

KubeArmor/core/k8sHandler.go
```diff
@@ -248,6 +248,40 @@ func (kh *K8sHandler) PatchDeploymentWithAppArmorAnnotations(namespaceName, depl
...
+		_, err = kh.K8sClient.AppsV1().ReplicaSets(namespaceName).Patch(context.Background(), deploymentName, types.StrategicMergePatchType, patchData, metav1.PatchOptions{})
+		if err != nil {
+			return err
+		}
+		time.Sleep(2 * time.Second)
+		patchData2 := []byte(fmt.Sprintf(`{"spec": {"replicas": %d}}`, replicas))
+		_, err = kh.K8sClient.AppsV1().ReplicaSets(namespaceName).Patch(context.Background(), deploymentName, types.StrategicMergePatchType, patchData2, metav1.PatchOptions{})
+		if err != nil {
+			return err
+		}
```
DelusionalOptimist, line 280: @Shreyas220 we seem to be going an infinite loop when a new Deployment is added and we're patching it. Can you check?

Shreyas220: Hey I tried but not able to reproduce this error

DelusionalOptimist: Not sure how to reproduce. Let's sync.
image

KubeArmor/core/kubeUpdate.go
Shreyas220: yeah found out what the issue is 
even for the infinite loop issue you mentioned below it was related to this 

so if the owner is a replica set it doesn't return any value so changed it to `ownerref.Name`

now if the owner of replicaset is a deployment we  using this `ownerref.Name` creates a problem resulting in the infinite loop

but the infinite loop only happens when we do make run not when i use container to deploy kubearmor 

cant seem to figure that one out :joy:

KubeArmor/core/k8sHandler.go
Shreyas220: Check again should be solved now :grin:

KubeArmor/core/kubeUpdate.go
Shreyas220: this was the reason for the error below 
only reproducible when testing using `make run`

DelusionalOptimist (commented):
@Shreyas220 can you checkout and sync your changes with recently merged #1135? I think it contains some relevant changes w.r.t pod metadata.

PR #1212, opened 2023-05-04
feat(visibility): disable visibility globally and enable only for specified namespaces

### LINKED ISSUE #1208: Turn off visibility/telemetry by default
## Feature Request

Currently, KubeArmor enables visibility/telemetry for every namespace by default. Thus all the events happening in context to all the workloads are by default reported to the userspace. This has performance implications. If the aim of the user is to simply apply a policy and get only policy violations then the visibility/telemetry is not needed to be enabled.

The intention of this issue is to turn off the visibility/telemetry by default in context to every namespace and the users can [selectively enable for any namespace](https://github.com/kubearmor/KubeArmor/blob/main/getting-started/kubearmor_visibility.md) that they are interested in.

yasin-cs-ko-ak:
**Purpose of PR?**:
This PR contains a new feature for disabling visibility global and only enable it for specific namespaces.
Fixes #1208 

No

1. Deployed kubearmor manifest to verify the changes

**Checklist:**


nyrahul (changes requested):
Hey @yasin-cs-ko-ak , you cannot make manual changes to these files. These manifests are auto-generated using [deployer](https://github.com/kubearmor/KubeArmor/blob/main/deployments/main.go).

You need to change the deployer and generate these yamls

nyrahul (approved):
Hey @achrefbensaad , please check this change from @yasin-cs-ko-ak ... it is better to change existing visbility flag as per this PR.

nyrahul (changes requested):
Handle the ConfigVisibility and remove dependency on cli args.

yasin-cs-ko-ak:
The following changes has been made:
1. Fixing ConfigMap to take values from depolyment for visibility (`visibility: none`)
2. Adding ConfigMap changes to watchers to check runtime modification in the deployment.
3. CLI dependency for visbility argument is removed. 

Kindly review the changes.

cc: @nyrahul @achrefbensaad @rksharma95

PR #1207, opened 2023-04-27
chore: Bump dependency versions

Vyom-Yadav:
Signed-off-by: Vyom-Yadav [jackhammervyom@gmail.com](mailto:jackhammervyom@gmail.com)

---

**Purpose of PR?**:

Bumps Dependency versions and go version in:

- `pkg/KubeArmorHostPolicy`
- `pkg/KubeArmorPolicy`

Bumps go version in CI tasks.

PR #1218, opened 2023-05-07
chore: Updated dependency versions 

Aryan-sharma11:
**Purpose of PR?**:

Updates dependencies in 
`KubeArmor`
`pkg/KubeArmorAnnotation`
`pkg/KubeArmorController`
`pkg/KubeArmorHostPolicy`
`pkg/KubeArmorPolicy`
`protobuf`

No


KubeArmor/go.mod
```diff
@@ -22,102 +22,106 @@ replace (
...
+	golang.org/x/sys v0.6.0
+	google.golang.org/grpc v1.55.0
+	k8s.io/api v0.26.4
+	k8s.io/apimachinery v0.26.4
+	k8s.io/client-go v0.26.1
 	k8s.io/cri-api v0.25.0
-	k8s.io/utils v0.0.0-20221107191617-1a15be271d1d
-	sigs.k8s.io/controller-runtime v0.13.1-0.20221201045826-d9912251cd81
+	k8s.io/utils v0.0.0-20230220204549-a5ecb0141aa5
+	sigs.k8s.io/controller-runtime v0.14.6
```
DelusionalOptimist: @Aryan-sharma11 let's try to keep all dependencies related to k8s at least at the same minor version (v0.26.x).
If you're getting any build error after updating one of these, it would be expected because we'll need to update the code accordingly first.

KubeArmor/go.mod
```diff
@@ -22,104 +22,105 @@ replace (
...
+	github.com/kubearmor/KubeArmor/protobuf v0.0.0-20230510133055-4e30a28b6352
+	github.com/opencontainers/runtime-spec v1.1.0-rc.2
+	github.com/spf13/viper v1.15.0
+	go.uber.org/zap v1.24.0
+	golang.org/x/sys v0.8.0
+	google.golang.org/grpc v1.55.0
+	k8s.io/api v0.26.4
+	k8s.io/apimachinery v0.26.4
+	k8s.io/client-go v0.26.4
+	k8s.io/cri-api v0.27.1
```
Aryan-sharma11, line 46: Note :- @DelusionalOptimist here `cri-api` is at `v0.27.1`  because `containerd v1.7.1` requires this version, so downgrading `cri-api` to `v0.26.4` will downgrade containerd to `v1.7.0` too

rksharma95:
are we not upgrading the dependencies for `KubeArmorPolicy` and `KubeArmorHostPolicy`. 
also we have a duplicate PR here #1207

DelusionalOptimist:
@rksharma95 As I mentioned in https://github.com/kubearmor/KubeArmor/pull/1207#issuecomment-1541545094, we're not upgrading the policy controllers in this PR because of the discussion I saw there.
However I see that you've now updated KubeArmor deps there as well. Since the changes in #1207 are less, let's merge that first and update this accordingly.

DelusionalOptimist (approved):
LGTM! :rocket:

KubeArmor/go.mod
```diff
@@ -19,42 +19,40 @@ replace (
 	github.com/kubearmor/KubeArmor/deployments => ../deployments
 	github.com/kubearmor/KubeArmor/pkg/KubeArmorController => ../pkg/KubeArmorController
 	github.com/kubearmor/KubeArmor/protobuf => ../protobuf
-	k8s.io/api => k8s.io/api v0.26.4
-	k8s.io/apimachinery => k8s.io/apimachinery v0.26.4
-	k8s.io/client-go => k8s.io/client-go v0.26.4
```
rksharma95: one question here, is it safe to remove this? are we sure here that executing go mod tidy will not upgrade these dependencies even if some other dependency upgrade these packages. i don't have complete understanding of how go mod tidy works.

DelusionalOptimist: I think only a `go mod tidy` won't update it unless someone runs `go get -u`.
But yes, you're right. We should have a replace directive otherwise once we integrate renovate, it will also try to update these to v0.27. :heavy_check_mark:

rksharma95 (approved):
LGTM :+1:

PR #1246, opened 2023-05-25
feat(operator): kubearmor-operator

### LINKED ISSUE #914: KubeArmor operator
# KubeArmor operator

## Aim

- Simplify our volumes mounts
- Unlock multi-enforcers cluster support
- Unlock multi-CRI cluster support
- Improve Karmor update procedure

## Node snitch
The node snitch is responsible for detecting the node runtime socket location(docker, crio, containerd, ...) and the node security enforcer (SeLinux, AppArmor, Bpf). After detection, the snitch will label the node with needed information.

## KubeArmor Operator
Is reponsible for installing KubeArmor and KubeArmor CRD's on the cluster, it makes use of the snitch to detect CRI and node enforcer.

## Diagram

kubearmor-operator drawio

## Work

 - [x] Node snitch
	 - [x] CRI detection
	 - [x] node enforcer detection
 - [x] KubeArmor operator
	 - [x] Install node snitch
	 - [x] Watch for node events 
	 - [x] Install CRD
	 - [x] Install KubeArmor daemonsets
	 - [x] Watch for changes in installed resources
- [x] Integrate with karmor
- [x] Test on multi-enforcer multi-container runtime cluster (Decided we will not do this unless users specifically ask for it. Till date there has been no clusters which face this problem).

### LINKED ISSUE #1200: RedHat OpenShift Certified Operator
## Feature Request

**Short Description**

Get KubeArmor Operator certified with RedHat OpenShift Product Certification,it'll provide the users with confidence that the KubeArmor Operator has been tested and verified to work on OpenShift in production env. This issue tracks the steps needed to complete the certification process.

**Requirements**

To certify the operator, these [requirements](https://access.redhat.com/documentation/en-us/red_hat_software_certification/8.59/html/red_hat_openshift_software_certification_policy_guide/assembly-products-managed-by-an-operator_openshift-sw-cert-policy-container-images) must be met.

```[tasklist]
### Tasks
- [ ] https://github.com/kubearmor/KubeArmor/issues/1201
- [ ] https://github.com/kubearmor/KubeArmor/issues/1409
- [x] Certify Each Container Image Managed by the KubeArmor Operator.
- [x] Address the requirements to be met by the Operator
- [x] Test the operator against the targeted OpenShift version.
- [x] Submit the operator for certification to the Red Hat Ecosystem Catalog.
```

rksharma95:
**Purpose of PR?**:
This PR is continuation of existing PR #925 , it adds KubeArmor Operator, it's a basic installation operator that can deploy KubeArmor on any k8s cluster considering the dynamic node configurations in terms of container runtime and LSM ( Linux Security Module). 

**Changes Includes**

- Created a CRD `KubeArmorConfig` to define configurations options for the KubeArmor.
- Implemented reconciler/watcher to deploy the KubeArmor applications with desire state.
- Added CI workflow to test and publish KubeArmor Operator to the docker registry `kubearmor/kubearmor-operator`.

**How to Test**
- Build a local image of the kubearmor-operator using `make docker build`
- deploy the operator using either helm chart or deployment file `deployments/operator.yaml` and use the locally build image there.
- after operator got deployed and running, apply sample (kubearmorconfig) crd `config/samples/sample-config.yml`
- observe the crd status and once it shows status `Running` it means all the required KubeArmor Components i.e., daemonset, kubearmor-relay, kubearmor-controller, etc. has been deployed on the cluster.
- try updating the `kubearmorconfig` and verify that KubeArmor Configured accordingly.

```
$ kubectl get pods -A
NAMESPACE     NAME                                  READY   STATUS    RESTARTS   AGE
kube-system   aws-node-b9vsm                        1/1     Running   0          73d
kube-system   aws-node-hpj66                        1/1     Running   0          73d
kube-system   coredns-7975d6fb9b-5v948              1/1     Running   0          73d
kube-system   coredns-7975d6fb9b-dzbxj              1/1     Running   0          73d
kube-system   kube-proxy-ftgzd                      1/1     Running   0          73d
kube-system   kube-proxy-n26q6                      1/1     Running   0          73d
kube-system   kubearmor-operator-6fd4b5d465-qrzlk   1/1     Running   0          7s

$ kubectl get kubearmorconfig -A -w
NAMESPACE     NAME                     STATUS
kube-system   kubearmorconfig-sample   Pending
kube-system   kubearmorconfig-sample   Running

$ kubectl get pods -A
kube-system   aws-node-b9vsm                          1/1     Running   0          73d
kube-system   aws-node-hpj66                          1/1     Running   0          73d
kube-system   coredns-7975d6fb9b-5v948                1/1     Running   0          73d
kube-system   coredns-7975d6fb9b-dzbxj                1/1     Running   0          73d
kube-system   kube-proxy-ftgzd                        1/1     Running   0          73d
kube-system   kube-proxy-n26q6                        1/1     Running   0          73d
kube-system   kubearmor-bpf-containerd-98c2c-7tdk8    1/1     Running   0          2m5s
kube-system   kubearmor-bpf-containerd-98c2c-9r7b2    1/1     Running   0          2m5s
kube-system   kubearmor-controller-7c48c74d58-n7s99   2/2     Running   0          2m3s
kube-system   kubearmor-operator-6fd4b5d465-qrzlk     1/1     Running   0          4m17s
kube-system   kubearmor-relay-6fddb8865b-86hfx        1/1     Running   0          2m42s
```

**HLD Diagram**

The linked issue #914 includes an attached High-Level Design (HLD) diagram illustrating the design of the KubeArmor Operator.

Attaching another diagram here that illustrates kubearmor deployment workflow using the operator:
kubearmor-operator

No

**Checklist:**


.github/workflows/ci-operator-release.yaml
```diff
@@ -0,0 +1,43 @@
+name: ci-release-operator
+
+on:
```
DelusionalOptimist, line 3: If KubeArmorOperator will be the main source for installing KubeArmor, I think we'll also need to publish stable images on tag releases? :thinking:

.github/workflows/ci-operator-release.yaml
```diff
@@ -0,0 +1,43 @@
...
+    name: Build & Push KubeArmor Operator
+    defaults:
+      run:
+        working-directory: ./pkg/KubeArmorOperator
+    runs-on: ubuntu-20.04
+    timeout-minutes: 60
+    steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: v1.20
```
DelusionalOptimist:
```suggestion
          go-version: "v1.20"
```

pkg/KubeArmorOperator/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+COPY main.go main.go
+COPY cmd cmd
+COPY common common
+COPY controllers controllers
+COPY enforcer enforcer
+COPY k8s k8s
+COPY runtime runtime
+# Build
+RUN CGO_ENABLED=0 GOOS=${GOOS} GOARCH=${GOARCH} GO111MODULE=on go build -a -o kubearmor-operator main.go
+RUN ln -s kubearmor-operator snitch
```
DelusionalOptimist: @rksharma95 this symlink would exist only in this build stage. In the final image we'll still have two copies of this binary which will be unnecessary addition to the image size.

pkg/KubeArmorOperator/PROJECT
```diff
@@ -0,0 +1,19 @@
+# Code generated by tool. DO NOT EDIT.
+# This file is used to track the info used to scaffold your project
+# and allow the plugins properly work.
+# More info: https://book.kubebuilder.io/reference/project-config.html
+domain: kubearmor.com
+layout:
+- go.kubebuilder.io/v3
```
DelusionalOptimist: Should we move to go/v4. :thinking: 
Ref - https://book.kubebuilder.io/plugins/go-v3-plugin.html

pkg/KubeArmorOperator/config/samples/kubearmor-default.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: operator.kubearmor.com/v1
+kind: Config
```
DelusionalOptimist:
```suggestion
kind: KubeArmorConfig
```

pkg/KubeArmorOperator/config/samples/operator_v1_config.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: operator.kubearmor.com/v1
+kind: Config
```
DelusionalOptimist:
```suggestion
kind: KubeArmorConfig
```

pkg/KubeArmorOperator/manifest/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
DelusionalOptimist: Can we add a single generated manifest file to the `deployments` directory like we have for other controllers? Helps in easier deployments :sweat_smile:

pkg/KubeArmorOperator/common/defaults.go
```diff
@@ -0,0 +1,263 @@
...
+	RuntimeStorageLabel             string = "kubearmor.io/runtime-storage"
+	SocketLabel                     string = "kubearmor.io/socket"
+	RandLabel                       string = "kubearmor.io/rand"
+	OsLabel                         string = "kubernetes.io/os"
+	ArchLabel                       string = "kubernetes.io/arch"
+	DeletAction                     string = "DELETE"
+	AddAction                       string = "ADD"
+	Namespace                       string = "kube-system"
+	Privileged                      bool   = true
+	OperatorImage                   string = "kubearmor/kubearmor-operator:24h"
```
DelusionalOptimist:
```suggestion
	OperatorImage                   string = "kubearmor/kubearmor-operator:latest"
```
Maybe?

DelusionalOptimist (changes requested):
@rksharma95 thanks for this PR! Great work going on. Please find the comments that I've added.

Also, I tried it out by building an image first locally and then generating manifests using `make deploy` and `make manifests`. Was able to get the operator installed. Hopefully I did it all correctly.
Once I applied the operator config given in samples directory, the operator seemed to have gone in a loop trying to update CRD status and the kubearmor-snitch pod creation failed due to image not found. Relay and controller deployments were created but kubearmor wasn't.

rksharma95:
> @rksharma95 thanks for this PR! Great work going on. Please find the comments that I've added.
> 
> Also, I tried it out by building an image first locally and then generating manifests using `make deploy` and `make manifests`. Was able to get the operator installed. Hopefully I did it all correctly. Once I applied the operator config given in samples directory, the operator seemed to have gone in a loop trying to update CRD status and the kubearmor-snitch pod creation failed due to image not found. Relay and controller deployments were created but kubearmor wasn't.

kubearmor-snitch must be failed because the image built locally should also be mentioned in the application itself (`common/defaults.go`) because snitch will deploy the same image too and at this point it's failed to find image at `kubearmor/kubearmor-operator` repo.

pkg/KubeArmorOperator/Dockerfile
rksharma95: so to create a symlink in the final image should i use alpine image instead?

pkg/KubeArmorOperator/PROJECT
rksharma95: done :+1:

pkg/KubeArmorOperator/Dockerfile
DelusionalOptimist: Why do we need to depend on binary name? Can we just use a flag or sub-command to differentiate between operator and snitch mode? :thinking:

pkg/KubeArmorOperator/cmd/main.go
```diff
@@ -0,0 +1,25 @@
...
+	"path/filepath"
+
+	operator "github.com/kubearmor/KubeArmor/pkg/KubeArmorOperator/cmd/operator"
+	snitch "github.com/kubearmor/KubeArmor/pkg/KubeArmorOperator/cmd/snitch-cmd"
+	"github.com/spf13/cobra"
+)
+
+func main() {
+	var command *cobra.Command
+	binaryName := filepath.Base(os.Args[0])
```
DelusionalOptimist: Let's use a flag or sub-command to distinguish between the two here, if possible?

pkg/KubeArmorOperator/cmd/operator/root.go
```diff
@@ -0,0 +1,78 @@
...
+	SilenceErrors: true,
+}
+
+func init() {
+	if home := homedir.HomeDir(); home != "" {
+		Cmd.PersistentFlags().StringVar(&KubeConfig, "kubeconfig", filepath.Join(home, ".kube", "config"), "Path to the kubeconfig file to use")
+	} else {
+		Cmd.PersistentFlags().StringVar(&KubeConfig, "kubeconfig", "", "Path to the kubeconfig file to use")
+	}
+	Cmd.PersistentFlags().StringVar(&LsmOrder, "lsmorder", "bpf,apparmor,selinux", "lsm preference order to use")
```
DelusionalOptimist:
```suggestion
	Cmd.PersistentFlags().StringVar(&LsmOrder, "lsm", "bpf,apparmor,selinux", "lsm preference order to use")
```
Let's keep consistency with Kubearmor's flags.

pkg/KubeArmorOperator/cmd/snitch-cmd/root.go
```diff
@@ -0,0 +1,146 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2021 Authors of KubeArmor
+
+// Package cmd is the collection of all the subcommands available in kArmor while providing relevant options for the same
+package nitch
```
DelusionalOptimist:
```suggestion
package snitch
```

pkg/KubeArmorOperator/internal/controller/cluster.go
```diff
@@ -0,0 +1,391 @@
...
+	RuntimeSocket  string
+	RuntimeStorage string
+	Arch           string
+}
+
+func NewClusterWatcher(client *kubernetes.Clientset, log *zap.SugaredLogger, extClient *apiextensionsclientset.Clientset, opv1Client *opv1client.Clientset, pathPrefix, deploy_name string) *ClusterWatcher {
+	if informer == nil {
+		informer = informers.NewSharedInformerFactory(client, 0)
+	}
+	if deployment_uuid == "" {
```
DelusionalOptimist, line 57: I think `deployment_uuid` would always be empty before this.

pkg/KubeArmorOperator/config/crd/bases/operator.kubearmor.com_kubearmorconfigs.yaml
```diff
@@ -0,0 +1,82 @@
...
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.4.1
+  creationTimestamp: null
+  name: kubearmorconfigs.operator.kubearmor.com
+spec:
+  group: operator.kubearmor.com
+  names:
+    kind: KubeArmorConfig
```
DelusionalOptimist, line 13: With this new CR, we'll be deprecating the current `kubearmor-config` configmap, right? Because it is the same thing and we'll have two conflicting sources of configuration otherwise.

daemon1024 (commented):
Folks, quick query w.r.t. #1265 

We have volume mounts w.r.t kernel headers 
which includes
/usr/src
/lib/modules

We don't need these mounts if the kernel version is above 5.2. Can we add that check in Operator and avoid those mounts?

Also we don't need /etc/apparmor if no AppArmor is present. Can we add that check as well. The code seems huge, can you comment where the logic is for generating the deployment file?

pkg/KubeArmorOperator/internal/controller/cluster.go
```diff
@@ -0,0 +1,391 @@
...
+		daemonset := generateDaemonset(daemonsetName, enforcer, runtime, socket, runtimeStorage)
+		_, err := clusterWatcher.Client.AppsV1().DaemonSets(common.Namespace).Create(context.Background(), daemonset, v1.CreateOptions{})
+		if err != nil {
+			clusterWatcher.Log.Warnf("Cannot Create daemonset %s, error=%s", daemonsetName, err.Error())
+		}
+	}
+
+}
+
+func (clusterWatcher *ClusterWatcher) WatchConfigCrd() {
```
DelusionalOptimist, line 222: We're watching the CRs created for `KubeArmorConfig` and not the CRD itself here, right? Can we fix this in logs and errors?

pkg/KubeArmorOperator/internal/controller/cluster.go
```diff
@@ -0,0 +1,391 @@
...
+	clusterWatcher.DaemonsetsLock.Unlock()
+
+	if deleteDaemonSet {
+		err := clusterWatcher.Client.AppsV1().DaemonSets(common.Namespace).Delete(context.Background(), daemonsetName, v1.DeleteOptions{})
+		if err != nil {
+			clusterWatcher.Log.Warnf("Cannot delete daemonset %s, error=%s", daemonsetName, err.Error())
+		}
+	}
+	if newDaemonSet {
+		daemonset := generateDaemonset(daemonsetName, enforcer, runtime, socket, runtimeStorage)
```
DelusionalOptimist: Question: Currently we're only creating a single daemonset for all kinds of nodes, right? Are we handling if different nodes have different enforcers or container runtimes?

pkg/KubeArmorOperator/internal/controller/resources.go
```diff
@@ -0,0 +1,585 @@
...
+				Name:      runtime + "-socket",
+				MountPath: socket,
+				ReadOnly:  true,
+			})
+			break
+		}
+	}
+	// lookup runtime storage location
+
+	for _, storageLocation := range common.RuntimeStorageVolumes[runtime] {
```
DelusionalOptimist, line 130: I'm using k3s with containerd (embedded) which has container storage at `/run/k3s/containerd` however it wasn't mounted on kubearmor. What might have happened? :thinking:

DelusionalOptimist (changes requested):
A few points to address for now.

pkg/KubeArmorOperator/internal/controller/cluster.go
DelusionalOptimist: https://github.com/kubearmor/KubeArmor/pull/1246#pullrequestreview-1513832099

@daemon1024 this might be what you were looking for w.r.t generating daemonset.

rksharma95:
> can you comment where the logic is for generating the deployment file?

@daemon1024 the complete workflow is like, 
when the snitch gets deployed on a node it adds the enforcer and runtime info as node labels, enfocer and runtime detection logic is implemented in `KubeArmorController/enforcer` and `KubeArmorController/runtime` respectively. 
 these info will be used by the node watchers defined in `KubeArmorOperator/internal/controller/cluster.go` to generate/update the daemonset.

pkg/KubeArmorOperator/internal/controller/resources.go
```diff
@@ -0,0 +1,585 @@
...
+		return nil, nil
+	}
+	for _, e := range strings.Split(enforcer, ".") {
+		vol = append(vol, common.EnforcerVolumes[e]...)
+		volMnt = append(volMnt, common.EnforcerVolumesMounts[e]...)
+	}
+	return
+}
+
+func genRuntimeVolumes(runtime, runtimeSocket, runtimeStorage string) (vol []corev1.Volume, volMnt []corev1.VolumeMount) {
```
daemon1024, line 107: How do we deal with AWS BottleRocket where dockershim is being used?

Ref https://github.com/kubearmor/KubeArmor/blob/0dd8f8bb9507aba1a5361092d85372f0fffccd34/deployments/BottleRocket/kubearmor.yaml#L2652

pkg/KubeArmorOperator/config/crd/bases/operator.kubearmor.com_kubearmorconfigs.yaml
rksharma95: no i don't think so, we support multiple deployment options i.e. helm and operator we are managing these configurations using ConfigMap in general and watching for the changes in KubeArmor. this CR will be the only source of the configuration when operator is being used to deploy the kubearmor. 

but yes i agree, currently it's possible to  have different configurations with this CR and the Configmap.

pkg/KubeArmorOperator/internal/controller/resources.go
rksharma95: i think we missed this case, need to handle this. thanks for pointing this out.

pkg/KubeArmorOperator/config/crd/bases/operator.kubearmor.com_kubearmorconfigs.yaml
DelusionalOptimist: @rksharma95 Just to make sure we are on the same page, will `karmor install` install the operator or kubearmor directly?

DelusionalOptimist: IMO we shouldn't be asking users to install kubearmor directly using Helm charts because there is no way to handle environment specific volumes etc automatically in Helm. Helm does support pre-install hooks however it is triggered only after kubernetes templates are generated from charts.
So, the users will have to specify it by overriding the `--environment.name` value so that proper container runtime etc maybe be used. It is very confusing and error prone.

rksharma95: imo `karmor install` should install KubeArmor directly, as operator requires user to deply operator CRs to deploy the KubeArmor. wdyt ?

DelusionalOptimist: I think karmor could still create the KubeArmorConfig CRs, if that's the need.
Let's get opinion on this from others as well?

pkg/KubeArmorOperator/internal/controller/cluster.go
rksharma95: > Are we handling if different nodes have different enforcers or container runtimes?

yes we're handling this case, we are probing each node for container runtime and enforcer and deploying the daemonset accordingly.

rksharma95:
> Also we don't need /etc/apparmor if no AppArmor is present. Can we add that check as well. The code seems huge, can you comment where the logic is for generating the deployment file?

yes we are detecting the enforcer and mounting the volumes accordingly.

pkg/KubeArmorOperator/internal/controller/cluster.go
rksharma95: @DelusionalOptimist do you see any issue here, yes we are checking if it's nil then only we are updating it.

DelusionalOptimist: What I was unsure about is if this check is needed in the first place. But yeah, let's leave it be this way, no harm done. :+1:

deployments/get/objects.go
```diff
@@ -368,6 +368,7 @@ func GenerateDaemonSet(env, namespace string) *appsv1.DaemonSet {
 										"IPC_LOCK",
 										"CAP_DAC_OVERRIDE",
 										"CAP_DAC_READ_SEARCH",
+										"CAP_AUDIT_READ",
```
achrefbensaad: Why we need this ?

achrefbensaad (approved):

deployments/get/objects.go
rksharma95: it's not required anymore, removed, thanks for pointing this out.

pkg/KubeArmorOperator/enforcer/enforcer.go
```diff
@@ -0,0 +1,79 @@
...
+
+probeBPFLSM:
+	if !slice.ContainsString(supportedLsms, "bpf", nil) {
+		err := probe.CheckBPFLSMSupport()
+		if err == nil {
+			supportedLsms = append(supportedLsms, "bpf")
+		} else {
+			log.Warnf("BPF LSM not supported %s", err.Error())
+		}
+	}
```
daemon1024:
```suggestion
probeLSM:
	if !slice.ContainsString(supportedLsms, "bpf", nil) {
		err := probe.CheckBPFLSMSupport()
		if err == nil {
			supportedLsms = append(supportedLsms, "bpf")
		} else {
			log.Warnf("BPF LSM not supported %s", err.Error())
		}
	}
	
	if !slice.ContainsString(supportedLsms, "apparmor", nil) {
		// Read file `/sys/module/apparmor/parameters/enabled`
	}
```

This is based on https://kubernetes.io/docs/tutorials/security/apparmor/#before-you-begin, You might want to include the link in code too.

This will help us probe for AppArmor even if the security LSM file is not there.

achrefbensaad (changes requested):
Please squash all commits

achrefbensaad (approved):
LGTM

pkg/KubeArmorOperator/internal/controller/resources.go
rksharma95: it's handled

pkg/KubeArmorOperator/internal/controller/resources.go
rksharma95: it has been fixed

.github/workflows/ci-operator-release.yaml
rksharma95: we'll handle it later as we need to refactor our existing CI release workflow.

DelusionalOptimist (approved):
LGTM. :shipit:

PR #1252, opened 2023-05-31
fix(enforcer): correct rule handling for matchDirectories

daemon1024:
We use hints to optimise directory map lookups in kernel space, but there are chances a particular path could both be actual directory rule and hint at the same time. We had incorrect handling where when hints are added they overrid the actual directory rule config. This commit fixes that handling


nyrahul (approved):
LGTM 👍

achrefbensaad (approved):
LGTM

PR #1293, opened 2023-06-29
feat: Annotating pre existing replicasets, statefulsets and daemonsets

### LINKED ISSUE #1180: pods deployed outside of deployments (such as statefulsets, replicasets) are not annotated on KubeArmor install

Shreyas220:
**Purpose of PR?**:

When kubeArmor is installed it doesnt add annotation to already present replicaset and statefulset

Fixes #1180 

### 1) Deployment Name 
to get the deployment name we were using this `pod.Metadata["deploymentName"]`  and to update the annotation we need ownerref.name as the deployment name 

### 2) Updating container list 
To update the container list we used this 

```
deploy, err := K8s.K8sClient.AppsV1().Deployments(pod.Metadata["namespaceName"]).Get(context.Background(), deploymentName, metav1.GetOptions{})

```
we cannot get replicaset and stateful set using this
logic is we loop through the containers and see if we annotations for that and update `updateAppArmor` variable as true which is required to go into `PatchResourceWithAppArmorAnnotations()` method 

#### Solution 
We again check for the **kind** of pod and use appropriate method to Get the the replicaset or statefulset 

### 3) In `PatchDeploymentWithAppArmorAnnotations` Method 

In this method we use this to patch the annotations but it only for deployments hence not updating replicaset and stateful set 

```
	_, err := kh.K8sClient.AppsV1().Deployments(namespaceName).Patch(context.Background(), deploymentName, types.StrategicMergePatchType, []byte(spec), metav1.PatchOptions{})
	if err != nil {
		return err
	}

``` 

#### Solution 

We send Kind of deployment and again check use the resource appropriate method to patch the annotations 

 

## Additional information for reviewers 

Using Patch method to update annotation restarts the deployments and statefulset , 
but by default patching doesn't restart the replicaset 

#### 1) Delete the replicaset (X)
we **do not** go with deleting the replicaset as it often permanently deletes the pod


KubeArmor/core/kubeUpdate.go
```diff
@@ -551,13 +551,14 @@ func (dm *KubeArmorDaemon) WatchK8sPods() {
 					if ownerRef.Kind == "ReplicaSet" {
 						deploymentName, deploymentNamespace := K8s.GetDeploymentNameControllingReplicaSet(pod.Metadata["namespaceName"], podOwnerName)
 						if deploymentName != "" {
-							pod.Metadata["deploymentName"] = deploymentName
-							pod.Metadata["owner.controllerName"] = deploymentName
+							pod.Metadata["deploymentName"] = ownerRef.Name
+							pod.Metadata["owner.controllerName"] = ownerRef.Name
```
Prateeknandle:
```
pod.Metadata["deploymentName"] = deploymentName
pod.Metadata["owner.controllerName"] = deploymentName
```
changing the value of both these field to `ownerRef.Name` will give us the name of the rs rather we want deploymentName and if deployment not exists for rs then we can have the rs name, which is covered already.
So ig no need to change the values here.

KubeArmor/core/kubeUpdate.go
```diff
@@ -702,12 +708,49 @@ func (dm *KubeArmorDaemon) WatchK8sPods() {
 					updateAppArmor := false
 
 					if deploymentName, ok := pod.Metadata["deploymentName"]; ok {
-						deploy, err := K8s.K8sClient.AppsV1().Deployments(pod.Metadata["namespaceName"]).Get(context.Background(), deploymentName, metav1.GetOptions{})
-						if err == nil {
-							for _, c := range deploy.Spec.Template.Spec.Containers {
-								containers = append(containers, c.Name)
+						if ownerRef != nil {
+							if ownerRef.Kind == "StatefulSet" {
```
Prateeknandle: What if we deploy a Pod, I don't see here this case covered? This will be handled when `ownerRef == nil `

Prateeknandle:
Also please add an example of updated pod with annotations added, replica set would be a good example for this.

Shreyas220:
> Also please add an example of updated pod with annotations added, replica set would be a good example for this.

both of them were present before installing kubearmor 

after installing kubearmor ->

### normal deployment with replicas 
```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    container.apparmor.security.beta.kubernetes.io/nginx: localhost/kubearmor-default-nginx-deployment-nginx
    kubearmor-policy: enabled
    kubearmor-visibility: process,file,network,capabilities
  creationTimestamp: "2023-07-04T04:24:12Z"
  generateName: nginx-deployment-85b57f68f8-
  labels:
    app: nginx
    pod-template-hash: 85b57f68f8
  name: nginx-deployment-85b57f68f8-qdfjs
```
yaml that i used

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 2
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80

```


### kind: replicaset 
```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    container.apparmor.security.beta.kubernetes.io/testnginx: localhost/kubearmor-default-web-testnginx
    kubearmor-policy: enabled
    kubearmor-visibility: process,file,network,capabilities
  creationTimestamp: "2023-07-04T04:24:58Z"
  generateName: web-
  labels:
    role: web
  name: web-6nxvb
  namespace: default

```
yaml i used 

```
apiVersion: apps/v1
kind: ReplicaSet
metadata:
  name: web
  labels:
    env: dev
    role: web
spec:
  replicas: 2
  selector:
    matchLabels:
      role: web
  template:
    metadata:
      labels:
        role: web
    spec:
      containers:
      - name: testnginx
        image: nginx

```

achrefbensaad (changes requested):
These wont work, Lets have a discussion,

KubeArmor/core/k8sHandler.go
```diff
@@ -248,6 +248,47 @@ func (kh *K8sHandler) PatchDeploymentWithAppArmorAnnotations(namespaceName, depl
 
 	spec = spec + `}}}}}`
 
+	if kind == "StatefulSet" {
```
Prateeknandle, line 251: Add an case for Pod as well here

KubeArmor/core/kubeUpdate.go
```diff
@@ -509,6 +509,66 @@ func (dm *KubeArmorDaemon) UpdateEndPointWithPod(action string, pod tp.K8sPod) {
 	}
 }
 
+func getTopLevelOwner(obj metav1.ObjectMeta, namespace string, objkind string) (string, string, string, error) {
```
Shreyas220: @achrefbensaad made changes i think this should cover all cases. let me know what case is left.
I think i should move this to k8s handler will do it tommrow

KubeArmor/core/kubeUpdate.go
```diff
@@ -509,6 +509,66 @@ func (dm *KubeArmorDaemon) UpdateEndPointWithPod(action string, pod tp.K8sPod) {
...
+
+	switch ownerRef.Kind {
+	case "Pod":
+		pod, err := K8s.K8sClient.CoreV1().Pods(namespace).Get(context.Background(), ownerRef.Name, metav1.GetOptions{})
+		if err != nil {
+			return "", "", "", err
+		}
+		if len(pod.OwnerReferences) > 0 {
+			return getTopLevelOwner(pod.ObjectMeta, namespace, "Pod")
+		}
```
Prateeknandle: This case will not occur for our use case.
Ref - https://accuknox.slack.com/archives/C0229TD83RA/p1688386506276139

For a static Pod `ownerRef` would be nil. 
For ex check how it is implemented here https://github.com/kubearmor/KubeArmor/blob/main/KubeArmor/core/kubeUpdate.go#L584

KubeArmor/core/kubeUpdate.go
```diff
@@ -509,6 +509,66 @@ func (dm *KubeArmorDaemon) UpdateEndPointWithPod(action string, pod tp.K8sPod) {
...
+		return deployment.Name, "Deployment", deployment.Namespace, nil
+	case "ReplicaSet":
+		replicaset, err := K8s.K8sClient.AppsV1().ReplicaSets(namespace).Get(context.Background(), ownerRef.Name, metav1.GetOptions{})
+		if err != nil {
+			return "", "", "", err
+		}
+		if len(replicaset.OwnerReferences) > 0 {
+			return getTopLevelOwner(replicaset.ObjectMeta, namespace, "ReplicaSet")
+		}
+		return replicaset.Name, "ReplicaSet", replicaset.Namespace, nil
```
Prateeknandle: why deployment case is not handled here?

Prateeknandle (commented):
Looks clean, adding some comments to handle.

KubeArmor/core/kubeUpdate.go
Shreyas220: that is handled , the function is recusively being called if there is a delpoyment owner it will be handled by case "deployment"

KubeArmor/core/kubeUpdate.go
Shreyas220: this is not for static pods

KubeArmor/core/kubeUpdate.go
Prateeknandle: ohh yes, my bad, got confused with `objkind`.

KubeArmor/core/kubeUpdate.go
Prateeknandle: can you give an example when Pod could be an owner of an kubernetes object related to our use case here? 
I'm not sure whether this could be useful here.

Shreyas220: :thinking: umm i cant find particular example for this case 
but i thought of this case as kiem tool would run as a job and will be created by another pod so the owner would be that pod.

Shreyas220: now that i think about it, this analogy doesnt make sense :sweat_smile: .
will this create any error ?
if not should we let it be there just in case (edge case) ?
from what i know its technically possible but doesnt really happen

Prateeknandle: Ohk, this won't create any error, I was just ensuring that we should not have anything unnecessary, anyways we can keep  it.

Prateeknandle (approved):

KubeArmor/enforcer/bpflsm/enforcer_bpfeb.o
achrefbensaad, line 1: Please revert the changes to `enforcer/bpflsm`

achrefbensaad (commented):
LGTM!, this can be merged after the enforcer changes are reverted. also we should  we should think about a new name for the `deploymentName` field as now we support more than deploymenyts. IMO this can be in a seperate PR as it will require changes in multiple places.

nyrahul:
We need automated tests for this issue.

.github/workflows/ci-test-ginkgo.yml
```diff
@@ -54,6 +54,12 @@ jobs:
         run: |
           make docker-build
 
+      - name: deploy pre existing pod 
+        run: |
+          kubectl apply -f ./tests/syscalls/manifests/pre-run-pod.yaml
+          sleep 60
+          kubectl get pods -A
```
DelusionalOptimist: @Shreyas220 I see that we've included these only in the CI. What if someone runs ginkgo locally?

tests/syscalls/syscalls_test.go
```diff
@@ -54,6 +57,73 @@ var _ = Describe("Syscalls", func() {
...
+			rspodName := rspods.Items[0].Name
+
+			rsannotated := false
+			for key, value := range rspods.Items[0].Annotations {
+				log.Printf("K8sGetPods pod=%s ns=%s Key=%v value=%s", rspods.Items[0].Name, "nginx", key, value)
+				if key == "kubearmor-policy" {
+					rsannotated = true
+				}
+
+			}
```
DelusionalOptimist: Can we use the existing helper functions for getting and matching pods here? :thinking:

tests/syscalls/syscalls_test.go
```diff
@@ -54,6 +57,73 @@ var _ = Describe("Syscalls", func() {
 		KspDeleteAll()
 	})
 
+	Describe("Annotation", func() {
```
DelusionalOptimist: Syscall test suite is meant for testing syscalls related policies and events. Can we put this in a different test suite. Maybe ksp?
Also, can we add a simple check to see if policy enforcement works for pre-existing pods? It would be nice to have this for edge cases, we currently don't have any such test in ksp.

DelusionalOptimist (changes requested):
Some points to address

.github/workflows/ci-test-ginkgo.yml
DelusionalOptimist: It isn't possible to do so in local as the flow for this is:
1. These pods are deployed.
2. Kubearmor is deployed.

We can guarantee this in the CI but not in dev's local env. So, we just resort to skipping it there.

DelusionalOptimist (commented):
@Shreyas220 Looking good. Please squash your commits.

DelusionalOptimist (approved):
LGTM. :rocket:

DelusionalOptimist:
@achrefbensaad Merging is blocked. Please approve.

achrefbensaad (approved):
LGTM!

PR #1313, opened 2023-07-10
fix(CI): KubeArmor-Controller version release

### LINKED ISSUE #1291: Versioned Release for kubearmor-relay and kubearmor-controller
## Feature Request

At this point we don't have versioned release (v*) workflow for [kubearmor-relay](https://github.com/kubearmor/kubearmor-relay-server) and [KubeArmorController](https://github.com/kubearmor/KubeArmor/tree/main/pkg/KubeArmorController).

```[tasklist]
### Tasks
- [ ] https://github.com/kubearmor/kubearmor-relay-server/issues/32
- [ ] https://github.com/kubearmor/KubeArmor/pull/1322
```

rksharma95:
**Purpose of PR?**:

After merging this PR, a versioned release of the `kubearmor-controller` will be created (if there's any change) along with the KubeArmor release under the same version tag as KubeArmor, as well as a latest release will be created for each change/PR merge.

Fixes #1291 

**Checklist:**
- [x] Bug fix. Fixes #1291 


.github/workflows/ci-stable-release.yml
```diff
@@ -65,3 +68,43 @@ jobs:
           base: main
           signoff: true
           delete-branch: true
+
+  build-and-push-controller-versioned-release:
```
daemon1024: I don't think we don't need this workflow

Ideally it should add simple as 
https://github.com/kubearmor/KubeArmor/blob/ea80cd58d2621914dfdcd9e17cd176172d8bb69f/.github/workflows/ci-stable-release.yml#L40

Adding another line below this 
```
regctl image copy kubearmor/kubearmor-controller:$STABLE_VERSION kubearmor/kubearmor-controller:stable --digest-tags 
```

This would work because `kubearmor/kubearmor-controller:vX.X` already exists and is pushed as part of `ci-latest-release`

daemon1024 (changes requested):

.github/workflows/ci-stable-release.yml
rksharma95: so with these changes we'll be releasing `kubearmor/kubearmor-controller:vX.X` along with each release of the KubeArmor not only the stable release right!

.github/workflows/ci-latest-release.yml
```diff
@@ -143,3 +165,50 @@ jobs:
         run: |
           STABLE_VERSION=`cat STABLE-RELEASE`
           regctl image copy kubearmor/kubearmor:$STABLE_VERSION kubearmor/kubearmor:stable --digest-tags
+
+      - name: Build & Push KubeArmorController
+        if: steps.match.outputs.tag == 'true'
+        working-directory: ./pkg/KubeArmorController
+        run: PLATFORM=$PLATFORM make docker-buildx TAG=$STABLE_VERSION
```
kranurag7: I think we are not populating this environment variable inside this step, which might break things. Is it env variable populated by GitHub itself? If no, let's use `env` block inside this step.

.github/workflows/ci-latest-release.yml
```diff
@@ -9,16 +9,38 @@ on:
...
+    name: Check what pkg were updated
+    if: github.repository == 'kubearmor/kubearmor'
+    runs-on: ubuntu-20.04
+    timeout-minutes: 5
+    outputs:
+      kubearmor: ${{ steps.filter.outputs.kubearmor}}
+      controller: ${{ steps.filter.outputs.controller }}
+    steps:
+    - uses: actions/checkout@v3
+    - uses: dorny/paths-filter@v2
```
kranurag7, line 30: Why are we needing this one? I think `working-directory` should suffice here. I am thinking it to be something like, if the conditions are satisfied to create a release then enter the `pkg/KubeArmorController` and then use `make docker-buildx`

.github/workflows/ci-latest-release.yml
rksharma95: yes you are right! i missed it to remove it from here we are handling it in makefile itself, thanks for pointing this out.

.github/workflows/ci-latest-release.yml
rksharma95: we are first checking if a pkg (kubearmor, controller) has been updated before making a release:
so with this workflow release will be like:
| pkg changed              | release         |
|--------------------------|-----------------|
| kubearmor                | kubearmor only  |
| controller               | controller only |
| kubearmor+ <br>  controller | release both    |

.github/workflows/ci-latest-release.yml
```diff
@@ -101,8 +123,8 @@ jobs:
 
   push-stable-version:
     name: Create KubeArmor stable release
-    needs: build
-    if: github.ref != 'refs/heads/main'
+    needs: [build, check]
+    if: github.ref != 'refs/heads/main' && needs.check.outputs.kubearmor == 'true'
```
daemon1024, line 127: push stable should not be limited to kubearmor, we need stable images for controller as well.


Noting need to add regctl copy for controller versioned release too

daemon1024 (changes requested):

daemon1024 (approved):

PR #1344, opened 2023-08-03
Feat/Docker-Compose-Deployment

### LINKED ISSUE #1341: Docker Compose Deployment for securing unorchestrated container
## Feature Request

**Short Description**

We currently have a systemd deployment which helps manage unorchestrated containers and host policies
Ref https://github.com/kubearmor/KubeArmor/blob/main/getting-started/kubearmor_vm.md

Folks might want to just start another container and not deal with the package management hassle to start systemd service.

Docker Compose File and Documentation to run KubeArmor directly with docker.

Here's how you can do it

```sh
docker run -v /opt/kubearmor/BPF:/opt/kubearmor/BPF --privileged kubearmor/kubearmor-init:stable

# Followed by

docker run -v /opt/kubearmor/BPF:/opt/kubearmor/BPF -v /sys/fs/bpf:/sys/fs/bpf -v /sys/fs/bpf:/sys/fs/bpf -v /sys/kernel/security:/sys/kernel/security -v /sys/kernel/debug:/sys/kernel/debug  -v /var/run/containerd/containerd.sock:/var/run/containerd/containerd.sock -v /run/containerd:/run/containerd -v /var/lib/docker:/var/lib/docker --privileged  --pid=host --ipc=host --net=host kubearmor/kubearmor:latest -k8s=false
```

This is privielged but we won't need the privileges and can mention exact capabilities as well in the docker compose file.

yashvardhanmishra:
**Purpose of PR?**:
 introducing a Docker Compose deployment option, enabling users to run KubeArmor directly with Docker

Fixes #1341 

- Added a new docker-compose.yml file in the root directory of the KubeArmor repository.
- Defined two services in the docker-compose.yml file: `kubearmor-init` and `kubearmor`.
- Configured volume mappings for both services . 

**Checklist:**
- [ ] Commit has unit tests


daemon1024:
Hey @yashvardhanmishra the list of capabilities can be found here https://github.com/kubearmor/KubeArmor/blob/52eb6fe9053a8c18d64792ab1d694e6dfc70a60b/deployments/get/objects.go#L332

We don't need any capabilities in init container, so let's keep it to default privileges neither true/false.

But there's a need to execute things in order. Init container needs to be executed before the main container. Main container has a dependency and needs to wait for the init container to be finished.

@sarthaksarthak9 I would love your inputs here.

@yashvardhanmishra Can you also squash your commits and sign off. Thank You.

sarthaksarthak9:
> Hey @yashvardhanmishra the list of capabilities can be found here
> 
> https://github.com/kubearmor/KubeArmor/blob/52eb6fe9053a8c18d64792ab1d694e6dfc70a60b/deployments/get/objects.go#L332
> 
> We don't need any capabilities in init container, so let's keep it to default privileges neither true/false.
> 
> But there's a need to execute things in order. Init container needs to be executed before the main container. Main container has a dependency and needs to wait for the init container to be finished.
> 
> @sarthaksarthak9 I would love your inputs here.
> 
> @yashvardhanmishra Can you also squash your commits and sign off. Thank You.

Please note that using depends_on only ensures the order of container startup, not the completion of specific actions inside the containers. You should ensure that the initialization logic in the kubearmor-init container finishes successfully before the kubearmor container starts its operations.

docker-compose.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - SYS_RESOURCE
+      - IPC_LOCK
+      - CAP_DAC_OVERRIDE
+      - CAP_DAC_READ_SEARCH
+    pid: host
+    ipc: host
+    network_mode: host
+    depends_on:
+      kubearmor-init:
+        condition: service_healthy
```
yashvardhanmishra: @daemon1024  By specifying `condition: service_healthy`, the `kubearmor` container will wait for the `kubearmor-init` container to be marked as healthy before it starts its operations

docker-compose.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - SYS_RESOURCE
+      - IPC_LOCK
+      - CAP_DAC_OVERRIDE
+      - CAP_DAC_READ_SEARCH
+    pid: host
+    ipc: host
+    network_mode: host
+    depends_on:
+      kubearmor-init:
+        condition: service_healthy # container will wait for the kubearmor-init container
```
daemon1024, line 36:
```suggestion
        condition: service_completed_successfully # container will wait for the kubearmor-init container 
```
> service_completed_successfully. This specifies that a dependency is expected to run to successful completion before starting a dependent service.

We should use this condition instead
https://docs.docker.com/compose/startup-order/#control-startup

daemon1024 (changes requested):
Thank You for handling it. I have a couple of concerns please find them below and inline.

Where do we pass the arguments to the kubearmor container?
We need to pass `-k8s=false` to KubeArmor container.

Have you tested the container based on https://github.com/kubearmor/KubeArmor/blob/main/getting-started/kubearmor_vm.md

We can skip the installation but we need to verify policy enforcement and alerts with this mode.

Can you also attach the logs of the kubearmor container once it starts. 

Please squash the commits as well.

docker-compose.yaml
yashvardhanmishra: hey @daemon1024 sorry for the delay , i was too busy ( due to exams ) , here is the update  - 

- i don't know what is the problem with `- /sys/kernel/security:/sys/kernel/security` volume it give me 
` Error response from daemon: mkdir /sys/kernel/security: operation not permitted` while  running the `kubearmor` container 
- else when excluding the above volume its running fine 
- and i have verified policy enforcement and alerts after passing `-k8s=false` and your above suggested change as well .

please guide me in this and there wont be any delay from now

docker-compose.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - SETUID
+      - SETGID
+      - SETPCAP
+      - SYS_ADMIN
+      - SYS_PTRACE
+      - MAC_ADMIN
+      - SYS_RESOURCE
+      - IPC_LOCK
+      - CAP_DAC_OVERRIDE
+      - CAP_DAC_READ_SEARCH
```
Ankurk99, line 30: Hey, I think the aim of explicitly specifying the capabilities is to remove the escalated `privileged: true` permission, right? 
If that's the case, then were you able to verify the kubearmor container running without the `priviledged` access?

ShubhamTatvamasi: @yashvardhanmishra please update.

daemon1024:
Closing due to inactivity, please reopen and handle review comments if active. Thank You for the efforts made till now.

PR #1345, opened 2023-08-04
chore(deployment):  Update RBAC rules for KubeArmor ClusterRole

Shreyas220:
**Purpose of PR?**:

Fixes: To update annotations for replicaset, daemon and replicaset we require patch rule which does not exist yet.  
Required for #1293 

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/pkg/KubeArmorOperator/deployments/helm/templates/clusterrole-rbac.yaml
+++ b/pkg/KubeArmorOperator/deployments/helm/templates/clusterrole-rbac.yaml
@@ -114,22 +114,15 @@ rules:
   - apps
   resources:
   - deployments
+  - replicasets
+  - daemonsets
+  - statefulsets
   verbs:
   - get
   - patch
   - list
   - watch
   - update
-- apiGroups:
-  - apps
-  resources:
-  - replicasets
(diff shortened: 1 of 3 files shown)


pkg/KubeArmorOperator/deployments/helm/templates/clusterrole-rbac.yaml
```diff
@@ -114,22 +114,15 @@ rules:
   - apps
   resources:
```
DelusionalOptimist, line 115: @rksharma95 Please tell if my understanding is correct. The operator needs these perms so that it can pass them on to KubeArmor, right?

rksharma95: yes you're right

DelusionalOptimist (approved):
We'll be sticking to the commit way right now as `deployments` package is outside `KubeArmorOperator` Dockerfile context. So images won't build.
Ref: https://github.com/kubearmor/KubeArmor/blob/main/pkg/KubeArmorOperator/Dockerfile

cc @daemon1024

PR #1377, opened 2023-08-19
fix(operator): include kubearmor installation via helm

### LINKED ISSUE #1374: Apply default kubearmor config as part of the Helm installation
## Feature Request

**Short Description**
Currently, when installing KubeArmor using Helm, we first install the operator and then apply the sample [KubeArmorConfig](https://raw.githubusercontent.com/kubearmor/KubeArmor/main/pkg/KubeArmorOperator/config/samples/sample-config.yml). Ideally, it should be packaged as part of the Helm installation process.

Deploy KubeArmor with default configuration as a part of helm installation. 

Edit: we can have KubeArmorConfig CR as part of operator helm, it should be configurable with a helm value by default set to false.

Ref PR: https://github.com/kubearmor/KubeArmor/pull/1377

Ankurk99:
With this pull request, KubeArmor will be installed with default configurations using the operator through a single Helm install command.

**Purpose of PR?**:

Fixes #1374 

No

Installation verified locally on k3s.

**Checklist:**
- [x] Bug fix. Fixes #1374 
- [x] New feature - NA
- [x] Breaking change - NA


deployments/helm/KubeArmorOperator/values.yaml
```diff
@@ -4,3 +4,26 @@ kubearmorOperator:
...
+
+    defaultVisibility: process,file,network
+
+    kubearmorImage:
+        image: kubearmor/kubearmor:stable
+        imagePullPolicy: Always
+
+    kubearmorInitImage:
+        image: kubearmor/kubearmor-init:stable
+        imagePullPolicy: Always
```
nyrahul: Shall we remove use of stable tag altogether and start using v0.11.0 tag?
We can remove dependency on `STABLE_RELEASE` altogther? WDYT? CC: @daemon1024

nyrahul (changes requested):

deployments/helm/KubeArmorOperator/values.yaml
Ankurk99: Then won't we have to update the charts every time we create a release?

nyrahul (approved):
LGTM

deployments/helm/KubeArmorOperator/README.md
```diff
@@ -1,28 +1,33 @@
...
 ```
 
 ## Values
+
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | kubearmorOperator.name | string | kubearmor-operator | name of the operator's deployment |
 | kubearmorOperator.image.repository | string | kubearmor/kubearmor-operator | image repository to pull KubeArmorOperator from |
 | kubearmorOperator.image.tag | string | latest | KubeArmorOperator image tag |
 | kubearmorOperator.imagePullPolicy | string | IfNotPresent | pull policy for operator image |
```
DelusionalOptimist, line 25: Can we update this table with the new values?

deployments/helm/KubeArmorOperator/README.md
```diff
@@ -1,28 +1,33 @@
...
 
-Once installed, the operator waits for the user to create a `KubeArmorConfig` object.
+Once installed, the operator uses [sample configuration](../../../pkg/KubeArmorOperator/config/samples/sample-config.yml) to create `KubeArmorConfig` object.
```
DelusionalOptimist: It's a bit misleading to say that the configuration on that path is used by the operator. I think we can stop referencing it all together.
How about: 
```suggestion
The operator needs a `KubeArmorConfig` object in order to create resources related to KubeArmor. A default config is present in Helm `values.yaml` which can be overridden during Helm install.

It is possible to specify configuration even after KubeArmor resources have been installed by directly editing the created `KubeArmorConfig` CR.
```

deployments/helm/KubeArmorOperator/templates/deployment.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ .Values.kubearmorOperator.name }}
-  namespace: {{ .Release.Namespace }}
+  namespace: {{ .Values.namespace | default "kubearmor" }}
```
DelusionalOptimist: IMO `namespace` in values is easy to overlook because most people are used to setting namespace with the `-n` flag. This will make the Helm `-n` flag useless which feels unconventional.
How about to fallback to kubearmor NS we do something like:
```suggestion
  namespace: {{- if eq .Release.Namespace "default" }} {{ printf "kubearmor" | trim }} {{- else }} {{ .Release.Namespace }} {{- end }}
```

deployments/helm/KubeArmorOperator/templates/serviceaccount.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ .Values.kubearmorOperator.name }}
-  namespace: {{ .Release.Namespace }}
\ No newline at end of file
+  namespace: {{ .Values.namespace | default "kubearmor" }}
```
DelusionalOptimist: Please also update namespace in RBAC rules' `subject`.

DelusionalOptimist (changes requested):
PTAL at the suggested comments.

deployments/helm/KubeArmorOperator/README.md
Ankurk99: The values seems to be matching the current state defined at Helm `values.yaml`. Can you please define what's need to be updated here?

deployments/helm/KubeArmorOperator/templates/deployment.yaml
Ankurk99: > IMO namespace in values is easy to overlook because most people are used to setting namespace with the -n flag. This will make the Helm -n flag useless which feels unconventional.


I didn't understand. What my understanding is that the default namespace to be used will be `kubearmor` and if a user defines any other ns using `-n` flag then that namespace will be used. Am I missing something?

deployments/helm/KubeArmorOperator/README.md
DelusionalOptimist: We added `configSpec` to values.yaml, right? Shouldn't we put in this table what it is for?

deployments/helm/KubeArmorOperator/templates/deployment.yaml
DelusionalOptimist: @Ankurk99  Variable references prefixed with `.Values` are used to refer to variables that can be specified by user. `values.yaml` is intended to be the place where charts define these values so that the users can reference it (through docs or the file itself) and also so that charts can set defaults for the values. This is the accepted best practice.

However, the `-n` flag translates internally to `.Release.Namespace` built-in variable. It is used to specify namespace that a chart should be installed into.
Ref - https://helm.sh/docs/chart_template_guide/builtin_objects/#helm

What I'm saying is instead of adding a new variable for namespace with `.Values.namespace`, let's use the already present built-in namespace object (`.Release.Namespace`).

deployments/helm/KubeArmorOperator/README.md
Ankurk99: The configSpec defined in `values.yaml` is the same as [KubeArmorConfig specification](https://github.com/kubearmor/KubeArmor/tree/main/deployments/helm/KubeArmorOperator#kubearmorconfig-specification)

DelusionalOptimist: Agreed. But how are users supposed to know if there is a `configSpec` variable which can be used to specify `KubeArmorConfig` in the first place!

deployments/helm/KubeArmorOperator/templates/deployment.yaml
Ankurk99: Got the point. I had an interesting conversation with Achref, and he made a good point. What if user actually wants to apply the kubearmor-operator in the `default` ns? 
I think we should change the ns value to be fetched from Release.Namespace and mention the default ns with the `-n` flag in the installation itself.

achrefbensaad (changes requested):
Can you please add a test for kubearmor via the chart ?

deployments/helm/KubeArmorOperator/README.md
```diff
@@ -105,8 +111,10 @@ NAME                               COMPLETIONS   DURATION   AGE
 job.batch/kubearmor-snitch-lglbd   1/1           3s         11m
 ```
 
-## Uninstall The Operator
+## Uninstall the Operator
+
 Uninstalling the Operator will also uninstall KubeArmor from all your nodes. To uninstall, just run:
+
 ```bash
 helm uninstall kubearmor -n kube-system
```
DelusionalOptimist:
```suggestion
helm uninstall kubearmor -n kubearmor
```

deployments/helm/KubeArmorOperator/kubearmor-operator.yaml
```diff
@@ -0,0 +1,333 @@
+---
```
DelusionalOptimist: Have we added this manifest generated by `helm template` intentionally? :thinking: 
We already have the one generated by kubebuilder for testing purposes kept here - https://github.com/kubearmor/KubeArmor/tree/main/deployments/operator

deployments/helm/KubeArmorOperator/templates/deployment.yaml
DelusionalOptimist: Sounds good :+1:

deployments/helm/KubeArmorOperator/kubearmor-operator.yaml
Ankurk99: My bad, I mistakenly left that. Thanks for pointing.

daemon1024:
Please include deployments/helm in the workflow paths.

Also update namespace in 
https://github.com/kubearmor/KubeArmor/blob/1e57ef20b00cd5419d5a6096af253246d6dd6b12/.github/workflows/ci-test-ginkgo.yml#L76C93-L76C104

deployments/helm/KubeArmorOperator/values.yaml
```diff
@@ -4,3 +4,26 @@ kubearmorOperator:
...
+    defaultNetworkPosture: audit
+
+    defaultVisibility: process,file,network
+
+    kubearmorImage:
+        image: kubearmor/kubearmor:stable
+        imagePullPolicy: Always
+
+    kubearmorInitImage:
+        image: kubearmor/kubearmor-init:stable
```
ShubhamTatvamasi: shouldn't we be moving to v0.11.0 instead on stable tag?

Ankurk99:
> Can you please add a test for kubearmor via the chart ?

Addressed :+1:

ShubhamTatvamasi:
Will it work on ARM based CPU?

DelusionalOptimist (approved):
LGTM! :rocket:

Ankurk99:
>Will it work on ARM based CPU?

The PR only automates the process of deploying KubeArmor from the Operator. If the Operator installation works fine on ARM then this will also work. 
We will need to verify the helm installation on ARM.

ShubhamTatvamasi (approved):

achrefbensaad (approved):
LGTM

deployments/helm/KubeArmorOperator/values.yaml
Ankurk99: Currently, we are following the flow of using `stable` and `latest` releases. If we decide to change that in future, then I think we should handle that in a separate PR.

Ankurk99:
We have reconsidered our approach and have opted to stick with the current method of installing the operator first, followed by the kubearmorconfig CR. The reasons are:
- it will be easier to use custom configurations by just editing the `kubearmorconfig` CR. This will make it easier to test the custom builds for CI testing.
- It is generally considered good practice to install an operator that subsequently awaits a Custom Resource to proceed.

PR #1396, opened 2023-09-05
feat: Listing the policies applied on containers in Un-orchestrated environment

### LINKED ISSUE #1311: list the applied policies for unorchestrated workloads
We need to list the applied policies on unorchestrated workloads using `karmor probe`, the same way we list the applied policies for orchestrted workloads.

we can add grpc endpoint and use it to list the policies for workloads.

Aryan-sharma11:
**Purpose of PR?**:

Fixes #1311 
image
image
This PR exposes the Policy and containers List which will  be utilised by Kubearmor client in systemd mode. 

**Does this PR introduce a breaking change?** NA

**Checklist:**
- [x] Commit has unit tests


protobuf/karmor.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+import "google/protobuf/empty.proto";
+package karmor;
+
+option go_package="github.com/kubearmor/KubeArmor/protobuf";
+
+message ContainerData {
```
daemon1024: Why is this a new proto? Shouldn't it be part of the Policy Proto? WDYT?

daemon1024 (changes requested):
Did a quick review, one review in line. Can you handle lint errors as well?

ShubhamTatvamasi:
@Aryan-sharma11 can you please rebase and run the CI tests again?

daemon1024 (approved):

KubeArmor/core/karmorprobedata.go
```diff
@@ -62,3 +72,77 @@ func (dm *KubeArmorDaemon) SetKarmorData() {
...
+	// Mapping Hostpolicies to their host hostName : HostPolicy
+	hostMap := make(map[string]*pb.HostSecurityPolicies)
+
+	dm.HostSecurityPoliciesLock.Lock()
+	for _, hp := range dm.HostSecurityPolicies {
+
+		hostName := ""
+		for _, v := range hp.Spec.NodeSelector.MatchLabels {
+			hostName = v
+		}
```
DelusionalOptimist: Are we looking for a specific label to match here? Why do we need this loop?
Also, KubeArmor only protects a single node anyways. The name is stored in `dm.Node.NodeName`. Shouldn't we directly use that?

KubeArmor/core/karmorprobedata.go
```diff
@@ -62,3 +72,77 @@ func (dm *KubeArmorDaemon) SetKarmorData() {
...
+	dm.HostSecurityPoliciesLock.Unlock()
+
+	return containerlist, containerMap, hostMap
+
+}
+
+// GetProbeData() sends policy data through grpc client
+func (p *Probe) GetProbeData(c context.Context, in *empty.Empty) (*pb.ProbeResponse, error) {
+
+	var containerMap map[string]*pb.ContainerData
```
DelusionalOptimist: This declaration is not needed.
```suggestion
```

protobuf/go.mod
```diff
@@ -11,12 +11,12 @@ replace (
 )
 
 require (
+	github.com/golang/protobuf v1.5.3
```
DelusionalOptimist: Can we stick to `google.golang.org/protobuf`? Mixing protobuf dependecies has led to terrible runtime panics in the past : (

.github/workflows/ci-test-ginkgo.yml
DelusionalOptimist, line 1: Also need to update other workflows where ginkgo runs. For instance `ci-latest-release.yaml`, `ci-test-controllers`.

tests/util/kartutil.go
```diff
@@ -598,3 +602,44 @@ func K8sRuntimeEnforcer() string {
 	runtimeEnforcer := nodes.Items[0].Labels["kubearmor.io/enforcer"]
 	return runtimeEnforcer
 }
+
+// CommandParse executes docker commmands
+func CommandParse(cmdstr string) (string, error) {
```
DelusionalOptimist:
```suggestion
func RunDockerCommand(cmdstr string) (string, error) {
```

DelusionalOptimist (changes requested):
@Aryan-sharma11 Mostly looking good. Please find some comments inline.

Though I couldn't tryout probe with kubearmor-client, I did manage to get the probe output with grpcurl.


image

In the above, I created a policy for ubuntu-vm but then deleted it. However it still showed up in the response.
Also, ubuntu-3 container didn't exist but it still showed up in the "containerMap" as it had a policy. This is fine but IMO it shouldn't have policyEnabled = 1 as this container doesn't exist. Though I think this information is taken directly from stated stored in KubeArmor but we might need to handle these edge cases either here or in client.

Aryan-sharma11:
> @Aryan-sharma11 Mostly looking good. Please find some comments inline.
> 
> Though I couldn't tryout probe with kubearmor-client, I did manage to get the probe output with grpcurl.
> 
> image
> 
> In the above, I created a policy for ubuntu-vm but then deleted it. However it still showed up in the response. Also, ubuntu-3 container didn't exist but it still showed up in the "containerMap" as it had a policy. This is fine but IMO it shouldn't have policyEnabled = 1 as this container doesn't exist. Though I think this information is taken directly from stated stored in KubeArmor but we might need to handle these edge cases either here or in client.



> @Aryan-sharma11 Mostly looking good. Please find some comments inline.
> 
> Though I couldn't tryout probe with kubearmor-client, I did manage to get the probe output with grpcurl.
> 
> image
> 
> In the above, I created a policy for ubuntu-vm but then deleted it. However it still showed up in the response. Also, ubuntu-3 container didn't exist but it still showed up in the "containerMap" as it had a policy. This is fine but IMO it shouldn't have policyEnabled = 1 as this container doesn't exist. Though I think this information is taken directly from stated stored in KubeArmor but we might need to handle these edge cases either here or in client.

Yes In the client we handle this case of policy being there and container not being present as we match the container Map with the Container List so only the active containers are shown.  for [ref](https://github.com/kubearmor/kubearmor-client/pull/372/files#diff-0b3ca41e370fff5ec9331a998f39c4c430b7bac2ceb0318fcb2a98f31e178b03R631).

DelusionalOptimist (approved):
LGTM. :rocket:

PR #1408, opened 2023-09-11
feat(ksp):  Adding protected pods to KSP

### LINKED ISSUE #1393: Include protected pods to the ksp status
## Feature Request

**Short Description**

Currently we rely on `karmor probe` output to figure out what are the pods that are protected by a certain ksp (KubeArmor Security Policy). We would like to include that information in the ksp status.

**Problem with `karmor probe`**

In `karmor probe` we are performing the policy matching on the client side. This turned out to cause some flakiness in output.

Kubearmor already maintains a list of pods and theirs respective ksp's. We can use this information to update the ksp with the protected targets list.

By doing this, we will be able to query the list of protected pods by using `kubectl` and stabilize `karmor probe`

- [ ] Propose a KSP update
- [ ] Update KSP code
- [ ] Update kubearmor to patch the ksp with the protected pods/resources list

PrimalPimmy:
**Purpose of PR?**:

Kubearmor always has maintained a list of protected pods and their respective policies, this PR adds a patch to show protected pods for the ksp.
Fixes #1393 

No

Below are the list of action items that needs to be done before merging:
- [x] Change ProtectedPods from Spec to the Status field of CRD
- [x] Find a way to use the function in `WatchK8sPod`
- [ ] Add test cases for this 

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/KubeArmor/config/config.go
+++ b/KubeArmor/config/config.go
@@ -33,6 +33,8 @@ type KubearmorConfig struct {
 	KVMAgent   bool // Enable/Disable KVM Agent
 	K8sEnv     bool // Is k8s env ?
 
+	ProtectedPods bool //To enable/disable showing protected pods in KSP
+
 	DefaultFilePosture         string // Default Enforcement Action in Global File Context
 	DefaultNetworkPosture      string // Default Enforcement Action in Global Network Context
 	DefaultCapabilitiesPosture string // Default Enforcement Action in Global Capabilities Context
@@ -79,6 +81,9 @@ const ConfigVisibility string = "visibility"
 
 // ConfigHostVisibility Host visibility key
 const ConfigHostVisibility string = "hostVisibility"
+
+// ConfigProtectedPods Protected Pods key
+const ConfigProtectedPods string = "enableProtectedPods"
 
 // ConfigKubearmorPolicy Kubearmor policy key
 const ConfigKubearmorPolicy string = "enableKubeArmorPolicy"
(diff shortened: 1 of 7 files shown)


PrimalPimmy:
image

PrimalPimmy:
TODO:
- [x]  Adding containers as part of pods
- [x] Adding flag for this feature

PrimalPimmy:
For multicontainer:
image


If only one container is protected:
image

rksharma95:
```yaml
  Selector:
    Match Labels:
      App:   busybox
  Severity:  3
Status:
  Protected Pods:
    nginx-77b4fdf86c-kfbsq [nginx]
```
changing target pods in the policy is not changing the protected pod status.

rksharma95:
Hey @PrimalPimmy why we choose managing another state with the maps for policy -> pod,container mapping instead of fetching the target pod information from the kubeapi server itself. we're already watching for both KSPs and Pods and we can update the information of the protected pods by a policy based on ksp and pod resource events. wdyt?

PrimalPimmy:
@rksharma95 I didn't understand. Could you elaborate how we would achieve this using pod resource events if we do that? Any examples where we do a similar thing?

rksharma95:
> @rksharma95 I didn't understand. Could you elaborate how we would achieve this using pod resource events if we do that? Any examples where we do a similar thing?

in brief I'm thinking something like:

| Resource  | Event  | Action |
|---------------|----------|----------|
| Ksp | Added/Updated | Get list of targeted pods matches with selector labels and check if it's a container specific policy then add the containers and pods accordingly to protected pods except it's annotated with `kubearmor-policy: audited`|
| Pod | Deleted | list the KSPs and check if a policy targeted the deleted pod then remove it from protected pods |

wdyt?

PR #1418, opened 2023-09-14
fix(operator): seperate snitch from the operator

### LINKED ISSUE #1409: Error running non-root container on containerd
## Bug Report

**General Information**

- Environment description (k3s)
- Kernel version (run `uname -a`) 5.14.0-284.25.1.el9_2.x86_64
- Orchestration system version in use (e.g. `kubectl version`, ...) v1.27.5+k3s1
- Link to relevant artifacts (policies, deployments scripts, ...)
- Target containers/pods `kubearmor-operator-*`

**To Reproduce**

1. Set up k3s with containerd
2. Deploy KubeArmor Operator using helm or manifest
3. Check if kubearmor operator pod is running, error `exec /operator: operation not permitted` is expected here

We're aware of this issue that we also faced with running (ubi-based, non-root) kubearmor container image on k8s clusters with containerd container runtime.  In these kind of scenarios the workaround is to add required capabilities to the binaries itself using the `setcap` during building the container image and assigning the required capabilities using the pod security context without `allowPrivilegeEscalation: false`. We tried all these workarounds but unfortunately none worked.
```
runAsUser: 1000
runAsGroup: 1000
allowPrivilegeEscalation: true
```
```
runAsUser: 0
runAsGroup: 0
allowPrivilegeEscalation: true
```

Related: https://github.com/kubernetes/enhancements/blob/master/keps/sig-security/2763-ambient-capabilities/README.md, 
https://github.com/containerd/containerd/issues/2516

**Expected behavior**

The operator should run successfully without any operation not permitted error.

**Screenshots**

rksharma95:
**Purpose of PR?**:

We're facing some issues with running containers with non-root user assigned with some extra privileges on containerd runtime. This PR handles this issue for KubeArmor Operator container image.

With this PR the snitch utility will be separated from the operator and  will be distributed with a new container image, by separating the snitch from the operator, now operator itself doesn't require any additional privileges i.e. SYS_ADMIN, SYS_RESOURCE, IPC_LOCK, etc. so now operator image can be run successfully within a non-root user container. however the snitch container image `kubearmor-snitch` will be running as root user with only necessary capabilities assigned.
 
Fixes (partially) #1409  

No
The changes has been successfully tested manually on a RedHat OpenShift Cluster.

**Checklist:**
- [x] Bug fix. (Partially) Fixes #1409
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) **(Not Applicable)**
- [x] Commit has unit tests


pkg/KubeArmorOperator/Dockerfile
```diff
@@ -58,8 +59,24 @@ RUN groupadd --gid 1000 default \
...
+
+LABEL name="kubearmor-snitch" \
+      vendor="Accuknox" \
+      version=${VERSION} \
+      release=${VERSION} \
+      summary="kubearmor-snitch container image based on redhat ubi" \
+      description="KubeArmor-Snitch, A CLI Utility to Detect node related informations for KubeArmor"
+
+ARG OPERATOR_DIR=pkg/KubeArmorOperator
+COPY --from=builder --chown=1000:1000 /KubeArmor/$OPERATOR_DIR/snitch /snitch
```
daemon1024: DO we need 1000 here? If it's a root binary?

pkg/KubeArmorOperator/internal/controller/resources.go
```diff
@@ -234,6 +230,8 @@ func deploySnitch(nodename string, runtime string) *batchv1.Job {
 							},
 						},
 						SecurityContext: &corev1.SecurityContext{
+							RunAsUser:  &rootUser,
+							RunAsGroup: &rootUser,
```
daemon1024, line 234: Do we need to explicitly specify it?

daemon1024 (commented):
Some queries inline. LGTM in general.

pkg/KubeArmorOperator/Dockerfile
rksharma95: it's a leftover i need to remove it, thanks for pointing this out.

pkg/KubeArmorOperator/internal/controller/resources.go
rksharma95: yes we need to specify it explicitly. on redhat if it's not specified explicitly it will be running as a non-root user it might be true for some other platform too.

daemon1024 (approved):

nyrahul (approved):

PR #1427, opened 2023-09-20
fix(core): handle containers in CRI agnostic way

daemon1024:
It is possible that runtime itself uses an arbitrary runtime names. Example robin platform uses robin as Container Runtime Version, in which case we depend on the CRISocket passed to KubeArmor for determining the CRI Version

there are CRI providers out there which use different CRI name such as robin instead of docker,containerd or crio

**Purpose of PR?**:

Fixes Deployment on Robin

tested on robin platform

**Checklist:**
- [x] Bug fix.

### DIFF
--- a/KubeArmor/core/kubeArmor.go
+++ b/KubeArmor/core/kubeArmor.go
@@ -512,15 +512,15 @@ func KubeArmor() {
 			}
 
 			// monitor containers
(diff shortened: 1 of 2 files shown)


DelusionalOptimist (approved):
LGTM!

PR #1430, opened 2023-09-21
docs: updated FAQ for kind cluster support on apparmor

nyrahul:
Kind clusters with distributions supporting only AppArmor needs special instructions specified as part of this FAQ update.

**Root cause**: Kind [recommends disabling AppArmor](https://kind.sigs.k8s.io/docs/user/known-issues/#apparmor) by default because of [known issues](https://github.com/moby/moby/issues/7512#issuecomment-51845976). However, KubeArmor requires AppArmor for policy enforcement in cases where BPF-LSM is not available. For slightly-old Ubuntu/Debian based distros, BPF-LSM is not available and thus the engine has to rely on AppArmor. KubeArmor injects a default apparmor profile for every workload and that overrides the default apparmor behavior and thus in general gets everything else working. However, KubeArmor does not apply default apparmor policies for the `kube-system` namespace and hence kubearmor-relay ends up using the default `cri.d.apparmor.d` apparmor policy that results in `socket: permission denied` for its grpc server bind socket call.

Ref: moby/moby#7512

**Checklist:**
- [x] Bug fix.
- [x] Commit has unit tests

### DIFF
--- a/getting-started/FAQ.md
+++ b/getting-started/FAQ.md
@@ -257,7 +257,7 @@ This will enable the `KubeArmorHostPolicy` and host based visibility for the k8s
 
 </details>
 
-<details><summary><h4>Unable to get KubeArmor policy enforcement with Kind clusters</h4></summary>
+<details><summary><h4>Using KubeArmor with Kind clusters</h4></summary>
 
 KubeArmor works out of the box with Kind clusters supporting BPF-LSM. However, with AppArmor only mode, Kind cluster needs additional provisional steps. You can check if BPF-LSM is supported/enabled on your host (on which the kind cluster is to be deployed) by using following:
 ```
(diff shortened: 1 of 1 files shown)


getting-started/FAQ.md
```diff
@@ -286,5 +286,10 @@ apt update && apt install apparmor-utils -y && systemctl restart containerd
 
 After this, exit out of the node shell and follow the [getting-started guide](https://github.com/kubearmor/KubeArmor/blob/main/getting-started/deployment_guide.md).
 
+If the `kubearmor-relay` pod goes into CrashLoopBackOff, apply the following patch:
+```sh
+kubectl patch deployment kubearmor-relay -n kube-system --type=json -p='[{"op": "add", "path": "/spec/template/metadata/annotations/container.apparmor.security.beta.kubernetes.io~1kubearmor-relay-server", "value": "unconfined"}]'
```
daemon1024: Will kube-system reference here be an issue since, we are moving out of it?

nyrahul: Fixed. It should work regardless of the namespace in which kubearmor-relay is deployed.

daemon1024 (approved):

PR #1441, opened 2023-10-04
feat(config): add untracked ns as a configurable option

Ankurk99:
**Purpose of PR?**:
Previously, the namespaces to be ignored from system monitor were hard coded, this PR makes it configurable.

No

Tested locally on k3s. Cases tested:
- [x] Default namespaces are being ignored
- [x] ns supplied as an env variable is working fine

**Checklist:**

### DIFF
--- a/KubeArmor/config/config.go
+++ b/KubeArmor/config/config.go
@@ -43,81 +43,41 @@ type KubearmorConfig struct {
 
 	CoverageTest bool // Enable/Disable Coverage Test
 
-	LsmOrder  []string // LSM order
-	BPFFsPath string   // path to the BPF filesystem
+	ConfigUntrackedNs []string // untracked namespaces
+	LsmOrder          []string // LSM order
+	BPFFsPath         string   // path to the BPF filesystem
 }
-
-// PolicyDir policy dir path for host policies backup
-const PolicyDir string = "/opt/kubearmor/policies/"
-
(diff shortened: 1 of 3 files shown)


KubeArmor/config/config.go
```diff
@@ -113,6 +114,9 @@ const ConfigCoverageTest string = "coverageTest"
 // ConfigK8sEnv VM key
 const ConfigK8sEnv string = "k8s"
 
+// Untracked Namespaces
+const UntrackedNs string = "untrackedNs"
```
daemon1024:
```suggestion
// Config Constants
const (
      ...
      ConfigUntrackedNs string = "untrackedNs"
      ...
 )
 
```

We have redundant comments here for exported constants, let's group them into a single constant declaration, this will prevent future lint warnings as well.
Let's use Config prefix for these constants, we are usually following it except LsmOrder and bpffs path.
For ref check
https://github.com/kubearmor/KubeArmor/blob/c81427eda39a1879206791a1504495ab37024b3a/KubeArmor/enforcer/bpflsm/rulesHandling.go#L16

daemon1024 (changes requested):
LGTM, Suggested a cleanup inline to eliminate linter warning. Thanks

daemon1024 (approved):

Aryan-sharma11 (approved):

PR #1496, opened 2023-11-06
fix(BPF-LSM):replace return statement with a check condition when file path is NULL

Prateeknandle:
**Purpose of PR?**:
There are some events for which `struct file` in `parent_task` becomes `NULL`. The existed implementation was that if  `struct file` is `NULL` we return empty for that event, due to which the further execution for matching and blocking didn't processed. Due to this block actions from outside the containers using `kubectl exec` sometimes failed, since there was a return statement when `struct file` was `NULL` while handling source path.

To handle this we replace return statements with if conditions, which check whether source's `struct file` is NULL or not. If anything is found NULL we do not go through the execution of fromSource checks.

NO

YES, 
checked the execution of blocked action using `kubectl exec` and it's working as expected.

**Checklist:**
- [ ] Commit has unit tests


Prateeknandle:
Hey Folks, PR is ready for review, can you please review it.
Thanks.

daemon1024 (approved):

Aryan-sharma11:
Looking good, just one thing is there any scope for adding tests regarding this?

Aryan-sharma11 (approved):

PR #1335, opened 2023-07-28
add multienforcer controller

rksharma95:
**Purpose of PR?**:
this PR adds multi-enforcer controller that handles the pod annotation based on the enforcer that scheduled node is configured with.
Fixes #

**Does this PR introduce a breaking change?**
No
**If the changes in this PR are manually verified, list down the scenarios covered:**:

image

**Additional information for reviewer?** :
_Mention if this PR is part of any design or a continuation of previous PRs_

**Checklist:**
- [ ] Bug fix. Fixes #<issue number>
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] PR Title follows the convention of  `<type>(<scope>): <subject>`
- [ ] Commit has unit tests
- [ ] Commit has integration tests

PR #1497, opened 2023-11-06
modifies the detect enforcer function in the controller

### LINKED ISSUE #1389: Remove kernel security path dependecy for kubearmor-controller
## Feature Request

**Short Description**

KubeArmor Controller make use of kernel security path `/sys/kernel/security/lsm` to detect the enforcer and this dependency can be removed safely now with KubeArmor Operator, as it deploys snitch on the cluster which detects the enforcer and attach this information to node using the labels. KubeArmor Controller can make use of this therefore it's no longer will required kernel security path to be mounted.

It will remove the dependency of a hostpath to be mounted and will improve the securityContext of the deployment.

remove dependency of kernel security path for KubeArmor Controller, get information of the enforcer using:
- node labels, or
- Add label to the KubeArmor Controller itself with the operator and use that.

swastik959:
**Purpose of PR?**:
kubearmor makes use of /sys/kernel/security/lsm to detect enforcer this can be removed as the operator now deploys the snitch which detects the enforcer and adds it to the node label
Fixes #1389 

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/.github/workflows/ci-latest-release.yml
+++ b/.github/workflows/ci-latest-release.yml
@@ -35,7 +35,7 @@ jobs:
             - "KubeArmor/**"
             - "protobuf/**"
           controller:
-            - 'pkg/KubeArmorController/**'
+            - 'pkg//**'
 
   build:
     name: Create KubeArmor latest release
--- a/pkg/KubeArmorController/main.go
+++ b/pkg/KubeArmorController/main.go
@@ -4,14 +4,15 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"os"
-	"path/filepath"
 	"strings"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
 
+	"k8s.io/apimachinery/pkg/types"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
 	"k8s.io/apimachinery/pkg/runtime"
@@ -26,6 +27,9 @@ import (
 	securityv1 "github.com/kubearmor/KubeArmor/pkg/KubeArmorController/api/security.kubearmor.com/v1"
 	"github.com/kubearmor/KubeArmor/pkg/KubeArmorController/controllers"
 	"github.com/kubearmor/KubeArmor/pkg/KubeArmorController/handlers"
+	"github.com/kubearmor/KubeArmor/pkg/KubeArmorOperator/common"
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	//+kubebuilder:scaffold:imports
 )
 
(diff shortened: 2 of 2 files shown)


rksharma95:
@swastik959 thanks for your contribution, can you also handle the deployment related changes with this PR itself.

nyrahul (commented):
Thank you for the PR. For (others) reference, the PR removes the need to depend on `/sys/kernel/security/lsm` file to detect enforcer settings. KubeArmor implemented an operator pattern sometime back that affixes a label on the node that specifies the policy enforcer in use.

However, my understanding is that this label is created only when the operator pattern is used and if someone deploys the kubearmor using the karmor cli then these labels are present.

@daemon1024 @Ankurk99 @achrefbensaad , isn't this correct?

I guess, we eventually intend to use operator mode of deployment even with karmor cli, but right now it is not present.

It is best to keep this PR active till next release and merge it along with the karmor changes.

Ankurk99:
>if someone deploys the kubearmor using the karmor cli then these labels are present.

 using the karmor cli then these labels are **not** present.
 Correct :+1:

swastik959:
@nyrahul can you create a seperate issue for that I would like to work on it .

.github/workflows/ci-latest-release.yml
```diff
@@ -35,7 +35,7 @@ jobs:
             - "KubeArmor/**"
             - "protobuf/**"
           controller:
-            - 'pkg/KubeArmorController/**'
+            - 'pkg//**'
```
kranurag7, line 38: I don't think this is required as of now but if we intend to trigger on operator changes as well then please revise this to `'pkg/**'`

pkg/KubeArmorController/main.go
```diff
@@ -26,6 +27,9 @@ import (
 	securityv1 "github.com/kubearmor/KubeArmor/pkg/KubeArmorController/api/security.kubearmor.com/v1"
 	"github.com/kubearmor/KubeArmor/pkg/KubeArmorController/controllers"
 	"github.com/kubearmor/KubeArmor/pkg/KubeArmorController/handlers"
+	"github.com/kubearmor/KubeArmor/pkg/KubeArmorOperator/common"
```
kranurag7, line 30: Please do `go mod tidy` here and update the deps.

pkg/KubeArmorController/main.go
```diff
@@ -139,29 +143,29 @@ func main() {
 }
 
 // detect the enforcer on the node
-func detectEnforcer(logger logr.Logger) string {
+func detectEnforcer(logger logr.Logger, client client.Client, nodeName string) string {
```
kranurag7, line 146: We can use `context.Context` here instead of using `context.TODO()` later, but I think it's okay given we use `context.TODO()` at other places in the code as well. 
// cc @daemon1024 

w.r.t this change, please update the call where we are registering webhook with updated arguments.
We should update this `Enforcer: detectEnforcer(setupLog),` as well.

pkg/KubeArmorController/main.go
```diff
@@ -139,29 +143,29 @@ func main() {
...
-		}
+	node := &v1.Node{}
+	err := client.Get(context.TODO(), types.NamespacedName{Name: nodeName}, node)
+	if err != nil {
+		logger.Error(err, "Failed to get node information")
+		return ""
 	}
 
-	enforcer := string(lsm)
+	enforcerLabel, exists := node.metadata.Labels[common.EnforcerLabel]
```
kranurag7, line 156:
```suggestion
	enforcerLabel, exists := node.ObjectMeta.Labels[common.EnforcerLabel]
```

pkg/KubeArmorController/main.go
```diff
@@ -139,29 +143,29 @@ func main() {
...
-	if _, err := os.Stat(filepath.Clean(lsmPath)); err == nil {
-		lsm, err = os.ReadFile(lsmPath)
-		if err != nil {
-			logger.Info("Failed to read /sys/kernel/security/lsm " + err.Error())
-			return ""
-		}
+	node := &v1.Node{}
+	err := client.Get(context.TODO(), types.NamespacedName{Name: nodeName}, node)
+	if err != nil {
+		logger.Error(err, "Failed to get node information")
```
kranurag7, line 152: Let's handle the cases where we miss getting information from some specific node in a big cluster. 
```suggestion
		logger.Error(err, "Failed to get enforcer information from node", "nodeName", node.Name)
```

kranurag7 (changes requested):

pkg/KubeArmorController/main.go
swastik959: @kranurag7 running go mod tidy is giving me a lot of errors of this type . `pr/secret.go/go/pkg/mod/sigs.k8s.io/kustomize@v2.0.3+incompatible/pkg/validators: import path "pr/secret.go/go/pkg/mod/sigs.k8s.io/kustomize@v2.0.3+incompatible/pkg/validators" should not have @version`

pkg/KubeArmorController/main.go
swastik959: @kranurag7 should I make changes in podannotator struct itself or just add client anf node name in the same file's  main function

DelusionalOptimist:
Folks, FYI kubearmor-client now installs KubeArmor using the operator, since it was a blocker to this PR earlier.
cc @swastik959 @rksharma95 @kranurag7

Aryan-sharma11:
Resolved with #1335. Thanks for the PR @swastik959

PR #1505, opened 2023-11-10
fix(monitor): refactor ExecPath handling to improve performance

daemon1024:
We read from procfs incase we missed information from ebpf, this ran everytime because we received the full information later down the line from bprm check. os.Readlink calls are expensive and running it on every log incurred lot of performance penalty.

This commit modifies the behaviour to only probe procfs on kretprobe only if the full path is yet not populated.

Even still if we miss the full path, we send out the relative path. The earlier behaviour was to set the Source to be empty, which would lead to dropping the alert.

Check in feeder to drop logs if no source has been removed, even without log.Source, the log event has useful information. It should be responsibility of the client to do the sanity check not KubeArmor.

Nope, it's backward compatible.

In progress, checking performance on dev cluster

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/KubeArmor/feeder/feeder.go
+++ b/KubeArmor/feeder/feeder.go
@@ -590,7 +590,7 @@ func (fd *Feeder) PushLog(log tp.Log) {
 	}
 
 	if log.Source == "" {
-		return
+		fd.Debug("Pushing Telemetry without source")
 	}
 
 	// set hostname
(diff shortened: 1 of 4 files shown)


KubeArmor/feeder/feeder.go
```diff
@@ -590,7 +590,7 @@ func (fd *Feeder) PushLog(log tp.Log) {
 	}
 
 	if log.Source == "" {
-		return
+		fd.Debug("Sending Log without no source")
```
seswarrajan:
```suggestion
		fd.Debug("Sending Log without source")
```

nyrahul (approved):
LGTM

KubeArmor/feeder/feeder.go
Ankurk99:
```suggestion
		fd.Debug("Sending Log without the source")
```

seswarrajan (approved):

Issue #1356, opened 2023-08-08
Events dropped due to busy event channel


daemon1024:
## Bug Report

KubeArmor Pod logs a LOT of 
`2023-08-08 06:40:11.xxxx	WARN	Event droped due to busy event channel`

**To Reproduce**

Trigger a lot of events in KubeArmor repeatedly

**Expected behavior**

It shouldn't be logged, it suggests there's an issue

**Screenshots**
![image](https://github.com/kubearmor/KubeArmor/assets/47106543/90425dd4-bb25-424a-9fa8-392dcaee47ee)

daemon1024:
There's some flaw with the replay logic we have

https://github.com/kubearmor/KubeArmor/blob/46a1e116bcd7dcb43f87e8a3c685e434649f19b9/KubeArmor/monitor/systemMonitor.go#L622

This was introduced back in https://github.com/kubearmor/KubeArmor/pull/1068 to fill up context on events where we might have missed due to delay in information from CRI.

h4shk4t:
I would like to pick up this issue.

clemenkok:
Would like to work on this.

nyrahul:
Folks, we might need a fix on this immediately. It would be better if the KubeArmor maintainer team picks it up.

daemon1024:
These are more apparent on hosts running both generic applications and kubernetes env ( example local k8s setups )
Or on dockerized node environment like kind, minikube.

KubeArmor assosciates each event with either the container or the host based on pid ns and mnt ns.

But there are events out there which don't categorize into either of them but still have a different linux namespace.
Example
- systemd user processes
- flatpaks, snaps, appimages
- dockerised nodes, they are host for the containers but in the kernel space they are actually a container and kubearmor cannot understand that.

KubeArmor cannot assosciate the said event and hence the drops log.

Some performance improvement have prevented the tight loop on replay channel, as well the ebpf prog will not send file data for these events which drastically reduces the number of these logs.

But these will be specially handled based on how we plan to implement #1503

PR #1491, opened 2023-11-01
feat(bpf,core,feeder,monitor): improve performance and owner only handling

daemon1024:
We leveraged CRI Storage mounts to stat owner information from container fs, now we extract that information from eBPF eliminating the need of reading/mounting containerfs

**Purpose of PR?**:

Fixes #1178

**Does this PR introduce a breaking change?**
No

**If the changes in this PR are manually verified, list down the scenarios covered:**:
CI Pass will validate this

**Additional information for reviewer?** :
_Mention if this PR is part of any design or a continuation of previous PRs_

**Checklist:**
- [ ] Bug fix. Fixes #<issue number>
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] PR Title follows the convention of  `<type>(<scope>): <subject>`
- [ ] Commit has unit tests
- [ ] Commit has integration tests

PR #1508, opened 2023-11-15
fix(*): KubeArmor goes brrr... ⚡⚡

daemon1024:
This PR handles of lots of edge cases which incur a lot of Performance load on KubeArmor.
Please go through individual commit messages to understand what has changed.
This in conjuction with #1491 and #1505 have brought down kubearmor CPU usage from >1000Mi to <100Mi ⚡


KubeArmor/BPF/system_monitor.c
```diff
@@ -408,81 +430,55 @@ static __always_inline void get_outer_key(struct outer_key *pokey,
...
-#else // MONITOR_CONTAINER or MONITOR_CONTAINER_AND_HOST
+// == Pid NS Management == //
 
+static __always_inline u32 add_pid_ns()
+{
+    struct task_struct *task = (struct task_struct *)bpf_get_current_task();
+    u32 one = 1;
     u32 pid_ns = get_task_pid_ns_id(task);
-    if (pid_ns == PROC_PID_INIT_INO)
+    if (get_kubearmor_config(_MONITOR_HOST) && pid_ns == PROC_PID_INIT_INO)
```
nyrahul: I think we should reverse the checks for performance reasons.
```suggestion
    if (pid_ns == PROC_PID_INIT_INO && get_kubearmor_config(_MONITOR_HOST))
```

KubeArmor/BPF/system_monitor.c
```diff
@@ -408,81 +430,55 @@ static __always_inline void get_outer_key(struct outer_key *pokey,
...
     { // host
         u32 pid = bpf_get_current_pid_tgid() >> 32;
-        if (bpf_map_lookup_elem(&pid_ns_map, &pid) != 0)
+        if (!bpf_map_lookup_elem(&pid_ns_map, &pid))
```
nyrahul, line 454: Isn't this change completely changing the semantics? reversing the check.

nyrahul: Oh I see the code below and it makes sense

KubeArmor/BPF/system_monitor.c
```diff
@@ -324,6 +324,28 @@ struct visibility
...
 
+#define DEFAULT_VISIBILITY_KEY 0xc0ffee
+
+// == Config == //
+
+enum
+{
+    _MONITOR_HOST = 0,
+    _MONITOR_CONTAINER = 1,
+    _ENFORCER_BPFLSM = 2,
```
nyrahul, line 335: Having MONITOR and ENFORCER in the same enum might be a problem in the future. Won't we need enforcer and Monitor enabled at the same time?

nyrahul (changes requested):

KubeArmor/BPF/system_monitor.c
daemon1024: These enums are not values but indexes of the map.

All three of the values are configured independently and serve different usecases. It's not an if else case.

daemon1024:
Closes https://github.com/kubearmor/KubeArmor/issues/1356

nyrahul (approved):
LGTM

PR #1521, opened 2023-11-28
fix(core): handle visibility if namespace info is missing

rksharma95:
**Purpose of PR?**:

this PR fixes Visibility Issue that might occur due to Pod-Container event latency. in case if container runtime detect the containers and emit those events before the pod events then the namespace value will not be present and the visibility will be defined incorrectly irrespective of namespace level visibility configuration. 

Fixes #

**Checklist:**
- [ ] Commit has unit tests


daemon1024 (changes requested):
Please don't commit changes in enforcer/*

daemon1024 (approved):

PR #1526, opened 2023-11-29
feat(gRPC): Implement m-TLS Support using self-signed certificate

### LINKED ISSUE #1464: Secure KubeArmor gRPC Endpoint
We use gRPC Insecure

https://github.com/kubearmor/kubearmor-relay-server/blob/34b9f3bd270e3edf6fbe3aae91504cc5f7f83ee3/relay-server/server/relayServer.go#L311

rksharma95:
**Purpose of PR?**:
This PR introduces m-TLS support using self-signed certificate to secure KubeArmor gRPC endpoint. with the tls encryption we ensure secure communication between the KubeArmor server and the clients i.e. kubearmor-relay.

Fixes #1464 

**Checklist:**
- [x] Bug fix. Fixes #1464 
- [ ] Commit has unit tests


KubeArmor/cert/cert.go
```diff
@@ -0,0 +1,209 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 Authors of KubeArmor
+
+package cert
```
achrefbensaad, line 6: please fix lint issue

achrefbensaad (dismissed):
Code Looks good to me, Can you please add the steps to test this PR.

KubeArmor/cert/cert.go
rksharma95: > Can you please add the steps to test this PR.

the testing will require the code changes in-place with the client. i will add the instructions for testing with subsequent PRs is it okay?

achrefbensaad: Sure

KubeArmor/feeder/feeder.go
```diff
@@ -32,6 +34,16 @@ import (
...
+var CACertPath = cert.CertPath{
+	Base:     cfg.GlobalCfg.TLSCertPath,
+	CertFile: "ca.crt",
+	KeyFile:  "ca.key",
+}
+var ServerCertPath = cert.CertPath{
+	Base:     cfg.GlobalCfg.TLSCertPath,
+	CertFile: "server.crt",
+	KeyFile:  "server.key",
+}
```
daemon1024: Why are these global variabled? and cannot be used from GlobalCfg directly?

daemon1024 (commented):
LGTM in general.
Small question/comment.

Can we include high level flow diagram of how the certificates are going to be generated, used and verified across KubeArmor and KubeArmor clients/relay

KubeArmor/feeder/feeder.go
rksharma95: the path to cert (directory) is configurable and can be specified using `--tlsCertPath` flag, and for simplicity we assume that ca/client/server certs are exist with particular file names. i.e. `ca.crt`, `ca.key`, `server.crt` etc.

daemon1024: But if we make it a global variable. It's never going to respect the value right? It's only going to use the build time value i.e. whatever is the default value is?

rksharma95: It set the path to --tlsCertPath="cert path" runtime flag value, currently the dynamic reload of any tls config is not being supported it require to restart kubearmor process.

daemon1024: I tried the following code
I set a global variable to another global variable
It only retained the build time value
Here's the code
```go
package main

import (
	"flag"
	"fmt"
)

var svar string = "original value"
var Globalvariable string = svar

func main() {
	flag.StringVar(&svar, "svar", "bar", "a string var")

	flag.Parse()

	fmt.Println("svar:", svar)
	fmt.Println("globalvar:", Globalvariable)
}
```

Here's the behaviour
image

Am I missing something here?

rksharma95: Got your point, yes you're right it will set the default value only. my bad i will update it. thanks for pointing this out :microscope:

rksharma95: is it okay if we hard code the file names for certs? do you think we need to have the configs for that too?
i.e. instead of `--tlsCertPath`, configs for cert and key path `[--caCertPath, --caKeyPath, --serverCertPath, --serverKeyPath]`

KubeArmor/feeder/feeder.go
```diff
@@ -711,3 +738,18 @@ func (fd *Feeder) PushLog(log tp.Log) {
...
+func loadTLSCredentials(ip string) (credentials.TransportCredentials, error) {
+	// create certificate configurations
+	serverCertConfig := cert.DefaultKubeArmorServerConfig
+	serverCertConfig.IPs = []string{ip}
+	serverCertConfig.NotAfter = time.Now().Add(365 * 24 * time.Hour) //valid for 1 year
+	// as of now daemonset creates certificates dynamically
+	tlsConfig := cert.TlsConfig{
+		CertCfg:      serverCertConfig,
+		CertProvider: cfg.GlobalCfg.TLSCertProvider,
+		CACertPath:   CACertPath,
```
daemon1024:
```suggestion
		CACertPath:   cert.CertPath{
	                                     Base:     cfg.GlobalCfg.TLSCertPath,
	                                     CertFile: "ca.crt",
	                                     KeyFile:  "ca.key",
                                           },
```

Are these variables being used somewhere else? Why can't we do this?

rksharma95: they are not being used anywhere else, we can do this.

daemon1024: Or just create a function in cert module saying
```go
func GetCACertPath(){
return cert.CertPath{
	                                     Base:     cfg.GlobalCfg.TLSCertPath,
	                                     CertFile: "ca.crt",
	                                     KeyFile:  "ca.key",
                                           }
}
```

This will always be using latest value of `cfg.GlobalCfg.TLSCertPath` whenever called

rksharma95: or 
```golang
func GetCACertPath(basePath string){
return cert.CertPath{
	                                     Base:    basePath,
	                                     CertFile: "ca.crt",
	                                     KeyFile:  "ca.key",
                                           }
}
```

then we can use it across all applications i.e. relay, karmor

daemon1024: Yeah makes sense

daemon1024 (approved):

achrefbensaad (approved):
LGTM

PR #1534, opened 2023-12-05
fix(core,monitor,feeder): fix telemetry missing data and configurable debug stdout logs

DelusionalOptimist:
**Purpose of PR?**:
Fixes
- Telemetry logs without data and only having cluster and hostname.
It was happening because of the recent removal of condition where we skip a log when `source` is empty. Which led to two things:
	- Visibility based dropping of host events, which happens in userspace, didn't work as intended. We intentionally leave the log empty if visibility of host is not enabled and this condition [here](https://github.com/kubearmor/KubeArmor/blob/719ae8563da75f877966f7d2cd675ed9187f1100/KubeArmor/feeder/feeder.go#L719) doesn't take into consideration empty `log.Type`. Thus, it ends up pushing a log which only has cluster and hostname taken from Node data. Also, the existing check for early return on empty source if `log.Type == HostLog` didn't work because HostLog type was only set if visibility was enabled.
	- Events, regardless of container or host, were skipped from sending over gRPC if they didn't have source. However, logs might be dropped based on visibility configuration, thus they'll have no log fields. So we put in a check for log.Type and if a log doesn't have a type, up until the point where we check it in `PushLog`, we skip sending it. This is safe to do as `log.Type` is not written over in any check after this point.

- Removes `DEBUG` logs from stdout. Need to use the environment var `DEBUG=true` to set this.
	NOTE: the flag `-debug` won't affect this as logger is initialized before config.

- Fixes panics at some places.

Fixes #

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/KubeArmor/Makefile
+++ b/KubeArmor/Makefile
@@ -50,7 +50,7 @@ run: build
 	cd $(CURDIR); sudo rm -f /tmp/kubearmor.log
 	cd $(CURDIR)/BPF; make clean
 	cd $(CURDIR)/BPF; make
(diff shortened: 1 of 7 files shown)


nyrahul (approved):
LGTM

daemon1024 (approved):

PR #1563, opened 2023-12-24
chore (build): updated workflows to decouple app version and chart version.

### LINKED ISSUE #1556: Have different chart and app version for helm charts
## Feature Request
- We as of now use same `version` and `appVersion` for kubearmor helm charts and this issue tracks having a separation between them. 

> I'm thinking of changing the versioning of our charts, differentiating `appVersion` and `chartVersion` to prevent overwrites on previous chart version, which often creates a lot of confusion.

_Originally posted by @DelusionalOptimist in https://github.com/kubearmor/KubeArmor/pull/1488#discussion_r1384820453_

suryyyansh:
**Purpose of PR?**:

Fixes #1556

No

locally ran the actions and checked the charts, all seems well.

_none_

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/.github/workflows/ci-latest-helm-chart-release.yaml
+++ b/.github/workflows/ci-latest-helm-chart-release.yaml
@@ -22,10 +22,16 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Update Chart.yaml
-        run: |
+        run: |  
+
           STABLE_VERSION=`cat STABLE-RELEASE`
-          sed -i -e "s/appVersion:.*/appVersion: $STABLE_VERSION/g" deployments/helm/*/Chart.yaml
-          sed -i -e "s/version:.*/version: $STABLE_VERSION/g" deployments/helm/*/Chart.yaml
+          for dep in deployments/helm/*/Chart.yaml; do
+            OLD_CHART_VERSION=`cat $dep | grep version | cut -d " " -f 2`
+            NEW_CHART_VERSION=`echo $OLD_CHART_VERSION | awk -F. '{$NF=$NF+1; print}' | tr ' ' '.'`
+            
+            sed -i -e "s/appVersion:.*/appVersion: $STABLE_VERSION/g" $dep
+            sed -i -e "s/version:.*/version: $NEW_CHART_VERSION/g" $dep
+          done
 
       - name: Publish Helm chart
         uses: stefanprodan/helm-gh-pages@master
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-latest-helm-chart-release.yaml
```diff
@@ -22,10 +22,16 @@ jobs:
...
 
       - name: Update Chart.yaml
-        run: |
+        run: |  
+
           STABLE_VERSION=`cat STABLE-RELEASE`
-          sed -i -e "s/appVersion:.*/appVersion: $STABLE_VERSION/g" deployments/helm/*/Chart.yaml
-          sed -i -e "s/version:.*/version: $STABLE_VERSION/g" deployments/helm/*/Chart.yaml
+          for dep in deployments/helm/*/Chart.yaml; do
+            OLD_CHART_VERSION=`cat $dep | grep version | cut -d " " -f 2`
```
kranurag7, line 29: Please move these two variable outside of this for loop.

.github/workflows/ci-latest-helm-chart-release.yaml
```diff
@@ -22,10 +22,16 @@ jobs:
...
+            NEW_CHART_VERSION=`echo $OLD_CHART_VERSION | awk -F. '{$NF=$NF+1; print}' | tr ' ' '.'`
```
kranurag7, line 30: This only handles patch releases, we also need to think about the case where we will need a minor/major release. 
Maybe we can use some other approach here.

.github/workflows/ci-latest-helm-chart-release.yaml
suryyyansh: Wouldn't doing that be counter intuitive as it would force both charts to have the same version?

kranurag7:
```bash
$ STABLE_VERSION=`cat STABLE-RELEASE`
$ for dep in deployments/helm/*/Chart.yaml; do
            OLD_CHART_VERSION=`cat $dep | grep version | cut -d " " -f 2`
            NEW_CHART_VERSION=`echo $OLD_CHART_VERSION | awk -F. '{$NF=$NF+1; print}' | tr ' ' '.'`
            sed -i -e "s/appVersion:.*/appVersion: $STABLE_VERSION/g" $dep
            sed -i -e "s/version:.*/version: $NEW_CHART_VERSION/g" $dep
done
$ cat deployments/helm/KubeArmor/Chart.yaml | grep version
version: v1.1.2
$ cat deployments/helm/KubeArmorOperator/Chart.yaml | grep version
# Application charts are a collection of templates that can be packaged into versioned archives
# This is the chart version. This version number should be incremented each time you make changes
# to the chart and its templates, including the app version.
version: Application.This.to.v1.1.2
```
Please notice the version of KubeArmorOperator chart.

PR #1632, opened 2024-02-15
fix(CI): Change latest release CI to containerd runtime

### LINKED ISSUE #1627: Rise of flakey tests
## Bug Report

**General Information**

- Some Ginkgo tests are failing in many PRs. Doesn't happen all the time but has got more frequent.
- Usually fails in the first attempts and have to restart the failed tests a few times for them to succeed.
- This may be due to Kubearmor slowing down in enforcing policies before we test policy violation.

**To Reproduce**

- Open up a PR and watch for the tests.

**Expected behavior**

We should not be having flaky tests at all.

PrimalPimmy:
**Purpose of PR?**:

Fixes #1627 , hopefully 

We are changing the runtime of our CI release workflow from docker to containerd.

No

**Checklist:**
- [x] Commit has unit tests


.github/workflows/ci-latest-release.yml
```diff
@@ -71,20 +71,35 @@ jobs:
...
+          sleep 60
+          kubectl get pods -A
+
+      - name: Run KubeArmor
+        run: |
+          docker save kubearmor/kubearmor-init:latest | sudo k3s ctr images import -
+          docker save kubearmor/kubearmor:latest | sudo k3s ctr images import -
+          docker save kubearmor/kubearmor-operator:latest | sudo k3s ctr images import -
+          docker save kubearmor/kubearmor-snitch:latest | sudo k3s ctr images import -
+
```
daemon1024: Image tags are not necessarily latest, for tag releases it uses the git tag as tag. SO please use that instead of latest

daemon1024 (changes requested):

.github/workflows/ci-latest-release.yml
daemon1024: You can see how .steps.var.output.tag is being used. Let's make sure it at both build and install time

.github/workflows/ci-latest-release.yml
```diff
@@ -71,20 +71,35 @@ jobs:
...
-      - name: Deploy KubeArmor into Kubernetes
+      - name: Build Kubearmor-Operator
+        working-directory: pkg/KubeArmorOperator
         run: |
-          helm upgrade --install kubearmor ./deployments/helm/KubeArmor \
-          --values ./KubeArmor/build/kubearmor-helm-test-values.yaml \
-          --set kubearmor.image.tag=${{ steps.vars.outputs.tag }} \
-          --set kubearmorInit.image.tag=${{ steps.vars.outputs.tag }} \
-          -n kubearmor --create-namespace;
+          make docker-build
```
daemon1024: This might need a ENV variable set too ig otherwise builds are latest
https://github.com/kubearmor/KubeArmor/blob/291c2b61178a9b8ba0744924596315081250cd5f/.github/workflows/ci-operator-release.yaml#L63C33-L63C85

daemon1024 (commented):
Please verify on your own repo if this works or not. The final push will not happen but atleast the tests can be verified

.github/workflows/ci-latest-release.yml
PrimalPimmy: although I have updated, I'm thinking about what to do with the operator and snitch versions, can make them follow the same tag version here or build as latest, since they get pushed in another CI I guess.

DelusionalOptimist: Folks, the `KubeArmorConfig` manifest that we have used only references `latest` images.
Ref - https://github.com/kubearmor/KubeArmor/blob/main/pkg/KubeArmorOperator/config/samples/kubearmor-test.yaml#L17-L22

So I think KubeArmor deployment will fail if images with the `latest` tag are not found. Let's export both tagged and `latest` images to k3s registry?

daemon1024: We should be changing the manifest with required version right?

DelusionalOptimist: Sure. The manifest is also referenced in ci-test-ginkgo.
Let's replace with sed in this workflow only?

daemon1024: Sure sed works.

daemon1024 (dismissed):

tests/k8s_env/ksp/ksp_test.go
```diff
@@ -109,12 +109,13 @@ var _ = Describe("Ksp", func() {
...
-			Expect(err).To(BeNil())
-			fmt.Printf("OUTPUT: %s\n", sout)
-			Expect(sout).To(MatchRegexp("PING.*127.0.0.1"))
+			Eventually(func() string {
+				sout, _, err := K8sExecInPod(ub1, "multiubuntu",
+					[]string{"bash", "-c", "ping -c 1 127.0.0.1"})
+				Expect(err).To(BeNil())
+				fmt.Printf("OUTPUT: %s\n", sout)
+				return sout
+			}, 10*time.Second, 2*time.Second).Should(MatchRegexp("PING.*127.0.0.1"))
```
daemon1024: We don't need eventually for Audit Policies, Let's refrain from adding that here.

tests/k8s_env/ksp/ksp_test.go
```diff
@@ -196,7 +203,7 @@ var _ = Describe("Ksp", func() {
 			err = KarmorLogStart("policy", "multiubuntu", "Network", ub1)
 			Expect(err).To(BeNil())
 
-			sout, _, err = K8sExecInPod(ub1, "multiubuntu",
+			sout, _, err := K8sExecInPod(ub1, "multiubuntu",
```
daemon1024: On the otherhand, we need to add the Eventually check here

daemon1024 (changes requested):
Let's carefully add the Eventually statement. 
We only need the Eventually statement when AppArmor related policy is being applied, not in audit/visibility related tests.

Can we think of a way to refactor this such that it's a function argument true/false.

PrimalPimmy:
Also I have added eventually to blockposture to test something, it won't be there in the final PR

daemon1024:
> Also I have added eventually to blockposture to test something, it won't be there in the final PR

I think we can add if we want to all test suites, that can be a different PR too.

tests/k8s_env/ksp/ksp_test.go
```diff
@@ -196,11 +192,17 @@ var _ = Describe("Ksp", func() {
...
+				MatchRegexp("Could not resolve host: google.com"), true,
+			)
+
+			Eventually(func() string {
+				sout, _, err := K8sExecInPod(ub1, "multiubuntu",
+					[]string{"bash", "-c", "curl google.com"})
+				Expect(err).To(BeNil())
+				fmt.Printf("OUTPUT: %s\n", sout)
+				return sout
+			}, 10*time.Second, 2*time.Second).Should(MatchRegexp("Could not resolve host: google.com"))
```
daemon1024: Why have both?

tests/k8s_env/ksp/ksp_test.go
```diff
@@ -732,13 +700,13 @@ var _ = Describe("Ksp", func() {
 			err = KarmorLogStart("policy", "multiubuntu", "File", ub4)
 			Expect(err).To(BeNil())
 
-			time.Sleep(5 * time.Second) // for policy
+			if strings.Contains(K8sRuntimeEnforcer(), "apparmor") {
+				time.Sleep(5 * time.Second) // for policy
+			}
```
daemon1024: we shouldn't need this

tests/k8s_env/ksp/ksp_test.go
```diff
@@ -1110,11 +1050,13 @@ var _ = Describe("Ksp", func() {
...
 			Expect(err).To(BeNil())
 
-			sout, _, err := K8sExecInPod(ub3, "multiubuntu",
-				[]string{"bash", "-c", "cat /home/user1/dir1/key1.txt"})
-			Expect(err).To(BeNil())
-			fmt.Printf("OUTPUT: %s\n", sout)
-			Expect(sout).To(MatchRegexp("key1\\.txt.*Permission denied"))
+			if strings.Contains(K8sRuntimeEnforcer(), "apparmor") {
+				time.Sleep(5 * time.Second) // for policy
+			}
```
daemon1024: not needed

tests/k8s_env/ksp/ksp_test.go
```diff
@@ -1847,7 +1747,7 @@ var _ = Describe("Ksp", func() {
 			err = KarmorLogStart("system", "multiubuntu", "File", ub4)
 			Expect(err).To(BeNil())
 
-			sout, _, err = K8sExecInPod(ub4, "multiubuntu",
+			sout, _, err := K8sExecInPod(ub4, "multiubuntu",
 				[]string{"bash", "-c", "cat /credentials/password"})
 			Expect(err).To(BeNil())
```
daemon1024: This can be moved to AssetCommand with eventually false as well

tests/util/kartutil.go
```diff
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	types2 "github.com/onsi/gomega/types"
```
daemon1024:
```suggestion
	gomegaTypes "github.com/onsi/gomega/types"
```
I don't think naming it types2 would be a good practice

tests/k8s_env/ksp/ksp_test.go
```diff
@@ -528,7 +511,7 @@ var _ = Describe("Ksp", func() {
...
 
 			// /bin/bash -c pwd shoud passed as only /bin/ls is blocked by policy
-			sout, _, err = K8sExecInPod(ub1, "multiubuntu",
+			sout, _, err := K8sExecInPod(ub1, "multiubuntu",
 				[]string{"bash", "-c", "/bin/dash -c pwd"})
 			Expect(err).To(BeNil())
 			fmt.Printf("OUTPUT: %s\n", sout)
```
daemon1024: We can move this to AssertCommand with eventually false, maybe keep the matcher to contain valid linux path or be * something like that

daemon1024 (changes requested):
Overall LGTM. Some nitpicks and reundundant code comments

- Some rendundant time.Sleep needs to be removed ( I have not commented on all of them )
- Let's move all K8sExecInPod to AssertCommand
- Some specific comments are inline, do check them out

tests/k8s_env/ksp/ksp_test.go
PrimalPimmy: missed to clear it, fixed.

tests/k8s_env/ksp/ksp_test.go
```diff
@@ -1657,15 +1537,13 @@ var _ = Describe("Ksp", func() {
...
-			fmt.Printf("OUTPUT: %s\n", sout)
-			Expect(sout).To(ContainSubstring("secret file user1"))
+			AssertCommand(ub3, "multiubuntu", []string{"bash", "-c", "su - user1 -c 'cat /home/user1/secret_data1.txt'"},
+				ContainSubstring("secret file user1"), false,
+			)
 
 			// Test 3: write operation on the file by the owner should also be allowed
-
-			sout, _, err = K8sExecInPod(ub3, "multiubuntu",
+			// No need for AssertCommand here since there is nothing to match
```
daemon1024, line 1545: We should move to AssertCommands anyway to have a streamlined way of Executing, if we need to make changes to the function in future atleast it's going to be at a single place.

Let's create a new Good First Issue to migrate all K8sExec to AssertCommand. We do no need to do this in current PR

daemon1024 (approved):
LGTM 👌🏽 Nice Work

May the force be with us and we do not need to fix tests again 🙏🏽 👼🏽

DelusionalOptimist (approved):
LGTM! :rocket:

tests/k8s_env/ksp/ksp_test.go
daemon1024: Marking this resolved for now, but leaving the issue creation to @PrimalPimmy

PR #1655, opened 2024-02-28
fix(core): fix a couple race conditions

dqsully:
**Purpose of PR?**:

While using KubeArmor, I noticed a couple bugs related to race conditions, and these commits should fix them.

The first race condition is just a concurrent map access that didn't lock the map first, and if two goroutines accessed the map simultaneously the whole KubeArmor container would crash and restart.

The second race condition is at node/KubeArmor startup. We were getting rare alerts (default Audit posture) on container behavior that was explicitly allowed through certain policies at node startup, and I determined that this could happen if a container is being enforced before KubeArmor has loaded all of the security policies, and realized that there's no protections in KubeArmor against that possible race condition.

(There is also one other possible source of a partial-enforcement race condition [here](https://github.com/kubearmor/KubeArmor/blob/33d7814ce8ea6e0878e0a18f45b61b8fc8527c54/KubeArmor/enforcer/bpflsm/rulesHandling.go#L384-L460) which can happen any time pods start or policies change. Currently container rules are added one-by-one via `be.ContainerMap[id].Map.Put()`, but by using `be.ContainerMap[id].Map.BatchUpdate()` and `be.ContainerMap[id].Map.BatchDelete()` we could make sure eBPF policies for a container are updated atomically and there's no partial enforcement happening. Getting this logic right though would take a lot more work and it should definitely happen in a separate PR.)

I have not tested to see if these changes fix the race conditions I encountered, because I can't easily replicate them. However, the concurrent map access issue did give me a stack trace pointing to exactly where the issue was.

While `factory.WaitForCacheSync(wait.NeverStop)` will wait for all informers from the factory to finish syncing their caches, apparently it does not necessarily wait for all event handlers to finish syncing, which is why I replaced it with the `registration.HasSynced` functions. There's more details in the commit message for that change.

**Checklist:**
- [X] Bug fix. Fixes 
- [ ] Commit has unit tests

### DIFF
--- a/KubeArmor/core/kubeArmor.go
+++ b/KubeArmor/core/kubeArmor.go
@@ -149,6 +149,8 @@ func NewKubeArmorDaemon() *KubeArmorDaemon {
 
 // DestroyKubeArmorDaemon Function
 func (dm *KubeArmorDaemon) DestroyKubeArmorDaemon() {
+	close(StopChan)
+
 	if dm.RuntimeEnforcer != nil {
 		// close runtime enforcer
 		if dm.CloseRuntimeEnforcer() {
@@ -702,12 +704,12 @@ func KubeArmor() {
 			timeout = 60 * time.Second
 		}
 
-		ctx, cancel := context.WithTimeout(context.Background(), timeout)
-		defer cancel()
-
 		// watch security policies
 		securityPoliciesSynced := dm.WatchSecurityPolicies()
 		if securityPoliciesSynced == nil {
+			// destroy the daemon
+			dm.DestroyKubeArmorDaemon()
+
 			return
 		}
 		dm.Logger.Print("Started to monitor security policies")
@@ -715,6 +717,9 @@ func KubeArmor() {
 		// watch default posture
 		defaultPostureSynced := dm.WatchDefaultPosture()
 		if defaultPostureSynced == nil {
+			// destroy the daemon
+			dm.DestroyKubeArmorDaemon()
+
 			return
 		}
 		dm.Logger.Print("Started to monitor per-namespace default posture")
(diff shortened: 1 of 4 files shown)


KubeArmor/core/kubeArmor.go
```diff
@@ -694,22 +696,38 @@ func KubeArmor() {
...
 
 		// watch kubearmor configmap
-		go dm.WatchConfigMap()
+		configMapSynced := dm.WatchConfigMap()
+		if configMapSynced == nil {
+			return
+		}
 		dm.Logger.Print("Watching for posture changes")
 
+		synced := cache.WaitForCacheSync(wait.NeverStop, securityPoliciesSynced, defaultPostureSynced, configMapSynced)
```
daemon1024: I don't think it's a good idea to user `wait.NeverStop` here
It was okay to use `wait.NeverStop` earlier because they ran inside a go routine. But this is the main thread. Blocking code in a main thread could possibly result in the program to be stuck here not responding to anything. 
Maybe a context with certain timeout would help here?

daemon1024 (changes requested):
One comment regarding `wait.NeverStop` inline. Let me know what do you think about that.

KubeArmor/core/kubeArmor.go
dqsully: Is there a certain timeout value that I should use here? Or should I add a CLI arg/config entry for it and maybe go with a default 60s, and end-users can override it if necessary?

daemon1024: I think it should be okay to keep a 60 second timeout. Sure let's have a flag we will reuse that timeout at other places when needed.

daemon1024 (dismissed):

KubeArmor/core/kubeArmor.go
```diff
@@ -694,22 +696,46 @@ func KubeArmor() {
...
+		if securityPoliciesSynced == nil {
+			return
+		}
 		dm.Logger.Print("Started to monitor security policies")
 
 		// watch default posture
-		go dm.WatchDefaultPosture()
+		defaultPostureSynced := dm.WatchDefaultPosture()
+		if defaultPostureSynced == nil {
+			return
```
DelusionalOptimist, line 723: Let's have a proper cleanup here and at all new returns by calling the `DestroyKubeArmorDaemon()` function.
Example - https://github.com/kubearmor/KubeArmor/pull/1655/files#diff-fa91eb0e35e389b4f70187ae5fd27e9608b1dc719b240edb96cc6dea3fc41deeR683

DelusionalOptimist (dismissed):
Looks great! Thanks @dqsully for noticing and fixing these. :raised_hands: 
Just one small suggestion.

DelusionalOptimist:
Hey @dqsully this is an important fix for the v1.3.0 release. The release will happen in a couple of days, so I made the small cleanup related change and rebased to get this merged quickly. Hope you don't mind. Thanks. : )

daemon1024 (approved):

rksharma95 (approved):

Issue #1186, opened 2023-03-26
securing KubeArmor pods and policies (Policing the Police!)


nyrahul:
KubeArmor is a security engine and thus it is imperative that it follows all the security best practices. The aim is to ensure security of the KubeArmor itself. Much of the work towards following best security practices (such as not using privileged flag) is already followed. It is now time to notch up the security practices to next level. The proposition is as follows:

- [x] Dogfooding
  - [x] Protect AppArmor policies mount point to be accessible only by KubeArmor binary. Enforce a KubeArmor policy to ensure this.
  - [x] Only allow access of service account token for kubearmor binary in the KubeArmor pod.
  - [x] Use hardening policies for KubeArmor pod.
- [x] Remove use of cluster-admin role
- [ ] Satisfy SLSA 3 requirements #1164 
- [x] Ability to use signed policy YAMLs only (orgs can leverage Adm controllers such as Kyverno)
- [x] seccomp profile for kubearmor
- [x] selinux/apparmor ootb profile
- [x] Fix all critical vulnerabilities across all kubearmor-images.
- [x] TLS for all intra-kubearmor communication #1526 
- [x] Reducing the kubearmor base image (can we try distroless kubearmor image?) [moved to UBI]
- [x] Remove unnecessary hostPaths from the manifests
- [x] Remove unnecessary capabilities from the manifests
- [x] Optimize RBAC permissions (implementation detail - config will be handled through Helm values of kubearmor-operator)
	- [x] Check if update perms can be removed for deployments/statefulsets etc - for AppArmor, patch should be suffice
	- [x] BPF LSM doesn't need the perms to annotate the above, set RBAC accordingly. (check kubearmor-policy annotation once though)
- [ ] Use `baseline` level in `enforce` mode Pod Security Admission for KubeArmor namespace
- [ ] attacker can disable kprobes, thus we won't get any future events 
        Predefined hardening policy for checking any changes to `/sys/kernel/debug/kprobes/enabled`

Ankurk99:
[`karmor recommend` for KubeArmor Daemonset](https://gist.github.com/Ankurk99/8cc2bd7b3aebfdfd856035b3c195e3bb)

daemon1024:
- How would KubeArmor daemonset attach a profile to itself?
  - Failed admission if annotation added and no profile found

Potential Solution
- Snitch has the intelligence to load the default apparmor profile.

DelusionalOptimist:
[...]

PR #1661, opened 2024-03-01
feat(core): Seccomp Profile for Kubearmor

PrimalPimmy:
**Purpose of PR?**:

Part of securing kubearmor in issue #1186 

This PR introduces seccomp profile for both kubearmor and kubearmor-init container, by only allowing necessary syscalls needed by Kubearmor to run and function properly.

Although there may not be any breaking change, we may need to be on the lookout incase we missed to whitelist any necessary syscall.

Couple of things that still needs to be done:
 ~Define custom Kubelet directory incase it is different than default  (Default is `/var/lib/kubelet/seccomp`)~ Also will be done later.  Seccomp will simply not load but kubearmor will run normally if path is different, and that's okay for now
~~Runtime default seccomp for all other Kubearmor pods.~~ Will be done in another PR/release

**Checklist:**
- [ ] Commit has unit tests


daemon1024:
Add an option to not add seccomp to security context of daemonset based on config crd

daemon1024 (dismissed):

PrimalPimmy:
Won't squash these 3 commits,  might need them for reference

pkg/KubeArmorOperator/common/defaults.go
```diff
@@ -108,6 +112,7 @@ var ConfigMapData = map[string]string{
 	ConfigDefaultNetworkPosture:      "audit",
 	ConfigVisibility:                 "process,network,capabilities",
 	ConfigDefaultPostureLogs:         "true",
+	ConfigDefaultSeccompEnabled:      "true",
```
rksharma95: in case if only seccomp configuration is changed in the CR, it will not be updated to the configmap, as it's not being checked in `UpdateConfigMapData()` either add it there or keept it seperate from configmap data if it's not being dynamically supported by daemonset itself.

rksharma95 (approved):
LGTM :+1:

daemon1024 (approved):

PR #1663, opened 2024-03-06
updated go version to 1.21.8

VedRatan:
**Purpose of PR?**:
~~This PR updates go version to latest one 1.22 including the CI.~~
Attempted to update to go 1.22 but snyk doesn't support it apparently. Will stick with go1.21.8 for now which has all the needed vuln. fixes backported.

Fixes #

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/.github/workflows/ci-systemd-release.yml
+++ b/.github/workflows/ci-systemd-release.yml
@@ -17,9 +17,9 @@ jobs:
         with:
           submodules: true
 
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v5
         with:
-          go-version: "v1.21"
+          go-version-file: 'KubeArmor/go.mod'
 
       - name: Install the latest LLVM toolchain
         run: ./.github/workflows/install-llvm.sh
--- a/pkg/KubeArmorController/go.mod
+++ b/pkg/KubeArmorController/go.mod
@@ -2,10 +2,12 @@ module github.com/kubearmor/KubeArmor/pkg/KubeArmorController
 
 go 1.21
 
+toolchain go1.21.8
+
 require (
 	github.com/go-logr/logr v1.4.1
 	github.com/onsi/ginkgo/v2 v2.13.0
-	github.com/onsi/gomega v1.29.0
+	github.com/onsi/gomega v1.30.0
 	k8s.io/api v0.29.0
 	k8s.io/apiextensions-apiserver v0.29.0
 	k8s.io/apimachinery v0.29.0
@@ -57,7 +59,7 @@ require (
 	golang.org/x/term v0.16.0 // indirect
 	golang.org/x/text v0.14.0 // indirect
 	golang.org/x/time v0.5.0 // indirect
-	golang.org/x/tools v0.16.0 // indirect
+	golang.org/x/tools v0.16.1 // indirect
 	gomodules.xyz/jsonpatch/v2 v2.4.0 // indirect
 	google.golang.org/appengine v1.6.8 // indirect
 	google.golang.org/protobuf v1.32.0 // indirect
--- a/renovate.json
+++ b/renovate.json
@@ -4,7 +4,7 @@
     "config:base"
   ],
   "constraints": {
-    "go": "1.21"
+    "go": "1.22"
   },
   "postUpdateOptions": ["gomodTidy"],
   "ignorePresets": [":prHourlyLimit2"],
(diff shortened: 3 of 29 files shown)


PrimalPimmy:
Are the errors in CI due to old go.sum? cc @DelusionalOptimist

anurag-rajawat:
`controller-gen` is crashing with go 1.22. So our CI is also failing due the same reason.
https://github.com/golang/go/issues/65637

pkg/KubeArmorController/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kubearmor/KubeArmor/pkg/KubeArmorController
 
-go 1.21
+go 1.22.1
```
kranurag7: I think here I would suggest using only `1.22` otherwise `go` will always try to pull the updated patch version. 
So, if you merge a PR for the patch `1.22.6` then on every system go will try to update it to 1.22.6 even if you run `go version` command in the directory. 

If you use `1.22` then it will support all the 1.22 patch series. 
Again, if you want that everyone should use what we have in patch releases then it's good but I think it's way frequent go download.

renovate.json
```diff
@@ -4,7 +4,7 @@
     "config:base"
   ],
   "constraints": {
-    "go": "1.21"
+    "go": "1.22.1"
```
kranurag7: let's go for 1.22 here as well.

pkg/KubeArmorController/go.mod
PrimalPimmy: yes, but it seems 1.22.1 has the vulnerability fixes:
https://pkg.go.dev/vuln/GO-2024-2600
https://pkg.go.dev/vuln/GO-2024-2599
https://pkg.go.dev/vuln/GO-2024-2598

on 1.22 the go-vuln check will always fail

kranurag7: I was aware of the vulns but not the fact that go-vulncheck requires exact pinning in go.mod.
If that's the case then we can go ahead. 
I've not tested it though (via phone now).

VedRatan: But certain vulnerabilities  are being handled in go 1.22.1.

PrimalPimmy:
@VedRatan use https://github.com/kubernetes-sigs/controller-tools/issues/880 , basically update the controller-gen module to 0.14 to fix it

VedRatan:
> @VedRatan use [kubernetes-sigs/controller-tools#880](https://github.com/kubernetes-sigs/controller-tools/issues/880) , basically update the controller-gen module to 0.14 to fix it

Yes I am upgrading the controller-gen, will be pushing the changes now.

pkg/KubeArmorController/go.mod
kranurag7: Yes, that is true. This is just so that developers have ease with development. 

If we patch go.mod with exact version very frequently then one have to download a new go version on every patch. 

This has nothing to do with production, we are already using patched version as a image and that's important and easy part because we build it using an image which is already patched. 

This point is about developing locally. 
I'm using go 1.22.0 locally and when I checkout your PR and run any go command, I get this.
```bash
$ go version
go: downloading go1.22.1 (linux/amd64)
```
When I update the same to go `1.22` in go.mod, it works fine. 

```diff
$ git diff -U0
diff --git a/KubeArmor/go.mod b/KubeArmor/go.mod
index 36dfd9f8..6d06d1da 100644
--- a/KubeArmor/go.mod
+++ b/KubeArmor/go.mod
@@ -3 +3 @@ module github.com/kubearmor/KubeArmor/KubeArmor
-go 1.22.1
+go 1.22
```
```bash
$ go version
go version go1.22.0 linux/amd64
```
So, do you want that every developer working on the project updates go on his system whenever we patch then please proceed with `1.22.1`
I think for someone this might be too early but given it's for a good and secure reason, I'm okay with this. I want to use the latest go version. :) 
We can proceed with the same.

.github/workflows/ci-systemd-release.yml
```diff
@@ -19,7 +19,7 @@ jobs:
 
       - uses: actions/setup-go@v3
         with:
-          go-version: "v1.21"
```
daemon1024, line 22: Let's use go version from go.mod file

https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file
This will avoid us needing to update workflows all the time

DelusionalOptimist:
Sorry @VedRatan... had to take this over as it was blocking CI for all other PRs for the upcoming release. : (

VedRatan:
No issues @DelusionalOptimist

daemon1024 (approved):

DelusionalOptimist (approved):

PR #1686, opened 2024-03-14
fix(enforcer/bpflsm): panic during bpflsm cleanup

daemon1024:
**Purpose of PR?**:

- panic during bpflsm cleanup
- fix missing proc name in alerts

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/KubeArmor/enforcer/bpflsm/enforcer.go
+++ b/KubeArmor/enforcer/bpflsm/enforcer.go
@@ -405,12 +405,11 @@ func (be *BPFEnforcer) DestroyBPFEnforcer() error {
 	if be == nil {
 		return nil
 	}
-
-	errBPFCleanUp := false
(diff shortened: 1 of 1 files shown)


Aryan-sharma11:
The changes makes sense, although I noticed one thing, if in case BPFLSM fails to load we move on to AppArmor, so if we install Kubearmor using operator ,it would still be showing podname `kubermor-bpf-crio` even if we are running apparmor under the hood. 
Can we handle that wdyt?

daemon1024:
@Aryan-sharma11 I don't think we can handle that right now. But we need to think about how to handle that in the operator refactor. Let's handle that seprately.

Aryan-sharma11 (approved):

rksharma95 (approved):

PR #1709, opened 2024-04-01
fix(grpc): extend grpc maxrecvmsg size for cri handlers

rksharma95:
**Purpose of PR?**:

when there are a lot of containers running, it might exceed the default grpc limit of 4MB msg size for cri handler services and that will lead to the issues.  
```
rpc error: code = ResourceExhausted desc = grpc: received message larger than max
```

No

**Checklist:**
- [x] Bug fix
- [ ] Commit has unit tests

### DIFF
--- a/KubeArmor/common/common.go
+++ b/KubeArmor/common/common.go
@@ -32,6 +32,13 @@ import (
 // == Common == //
 // ============ //
 
+const (
+	// grpc default is 4MB
+	// CRI i.e. containerd service can send msg extended upto 16MB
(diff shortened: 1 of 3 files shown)


daemon1024 (commented):
Do we need to handle this for docker socket as well?

rksharma95:
> Do we need to handle this for docker socket as well?

in case of docker we're using http client https://pkg.go.dev/github.com/docker/docker@v24.0.7+incompatible/client , it's not needed in this case.

daemon1024 (approved):

DelusionalOptimist (approved):
LGTM

PR #1714, opened 2024-04-08
feat: Update KubeArmor to use OCI hooks instead of depending on container runtime socket

### LINKED ISSUE #1390: feat: Leverage OCI Hooks for Container Events
## Feature Request

**Short Description**

 Use OCI hooks and get events in context to container start/stop: Currently KubeArmor mounts docker/containerd/crio UNIX domain socket file in KubeArmor to watch for container events. The aim is to use OCI hooks for getting such container events.

Eliminate exposing docker/containerd/crio UNIX domain sockets inside a container. 

OCI hooks can be used to get containers create/stop events which does not require any access to runtime sockets.

Posix hooks documentation - https://github.com/opencontainers/runtime-spec/blob/v1.1.0/config.md#posix-platform-hooks
Inspektor-gadget runc hook example - https://github.com/inspektor-gadget/inspektor-gadget/tree/main/examples/runc-hook

Nothing.

**Note**

Created out of https://github.com/kubearmor/KubeArmor/issues/1130. Kept earlier issue for OCI registry feature.

AbdelrahmanElawady:
**Purpose of PR?**:
This PR utilizes OCI hooks to get container details inside KubeArmor instead of using container runtime socket. It depends on some Kubernetes annotations to get container name, pod name, namespace name and AppArmor profile (it looks like AppArmor is the only one required out of those but further testing is required).
This PR also updates snitch to configure hooks on the host (currently only CRI-O is supported).

Fixes #1390

No, the goal is to have the same functionality as mounting container runtime socket without the security concerns of doing that.

Tested with Getting Started example for now but more testing will be done.

- Containers created before KubeArmor was deployed is still not handled.

**Checklist:**
- [ ] Commit has unit tests


AbdelrahmanElawady:
Container created before KubeArmor now are handled with a simple detached process that waits on KubeArmor to start then sends all previous containers.

AbdelrahmanElawady:
Also, in order to try out this PR. You only need to build this code and run it on a CRI-O Kubernetes cluster and operator with snitch will take care of setting up the cluster.

KubeArmor/core/hook_handler.go
```diff
@@ -0,0 +1,200 @@
...
+
+// ListenToHook starts listening on a UNIX socket and waits for container hooks
+// to pass new containers
+func (dm *KubeArmorDaemon) ListenToHook() {
+	if err := os.MkdirAll(kubearmorDir, 0750); err != nil {
+		log.Fatal(err)
+	}
+
+	listenPath := filepath.Join(kubearmorDir, "ka.sock")
+	_ = os.Remove(listenPath) // in case kubearmor crashed and the socket wasn't removed
```
achrefbensaad:
```suggestion
	_ = os.Remove(listenPath) // in case kubearmor crashed and the socket wasn't removed
	listenPath := filepath.Join(kubearmorDir, "ka.sock")
	
```

achrefbensaad: you need to handle the error case

achrefbensaad: just reordering for better clarity

KubeArmor/core/hook_handler.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	defer socket.Close()
+	defer os.Remove(listenPath)
+	ready := &atomic.Bool{}
+
+	for {
+		conn, err := socket.Accept()
+		if err != nil {
+			log.Fatal(err)
```
achrefbensaad, line 48: Are we really sure we want to stop the entier execution if a single connection is not handled ?

KubeArmor/core/hook_handler.go
```diff
@@ -0,0 +1,200 @@
...
+	defer conn.Close()
+	buf := make([]byte, 4096)
+
+	for {
+		n, err := conn.Read(buf)
+		if err == io.EOF {
+			return
+		}
+		if err != nil {
+			log.Fatal(err)
```
achrefbensaad, line 71: same comment as above

achrefbensaad: Please handle in other parts in the code as well

KubeArmor/core/hook_handler.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+
+		data := struct {
+			Operation string          `json:"operation"`
+			Detached  bool            `json:"detached"`
+			Container types.Container `json:"container"`
+		}{}
```
achrefbensaad: Define the struct outside function at the top of the file

KubeArmor/core/hook_handler.go
```diff
@@ -0,0 +1,200 @@
...
+
+		_, err = conn.Write([]byte("ok"))
+		if err == io.EOF {
+			return
+		} else if err != nil {
+			log.Println(err)
+			return
+		}
+
+		if data.Operation == "create" {
```
achrefbensaad: Better to use enums

KubeArmor/core/hook_handler.go
```diff
@@ -0,0 +1,200 @@
...
+}
+func (dm *KubeArmorDaemon) handleContainerCreate(container types.Container) {
+	endpoint := types.EndPoint{}
+
+	dm.Logger.Printf("added %s", container.ContainerID)
+
+	dm.ContainersLock.Lock()
+	if _, ok := dm.Containers[container.ContainerID]; !ok {
+		dm.Containers[container.ContainerID] = container
+		dm.ContainersLock.Unlock()
```
achrefbensaad: Better to unlock once outside the if-else block rather than unlocking at each branch

pkg/KubeArmorOperator/hook/crio.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	client := runtime.NewRuntimeServiceClient(conn)
+	return &crioHandler{client: client, conn: conn}, nil
+}
+
+func (h *crioHandler) close() {
+	_ = h.conn.Close()
```
achrefbensaad: handle error

pkg/KubeArmorOperator/hook/crio.go
```diff
@@ -0,0 +1,106 @@
...
+	containers := make([]types.Container, 0)
+	containersList, err := h.client.ListContainers(ctx, &runtime.ListContainersRequest{})
+	if err != nil {
+		return nil, err
+	}
+	for _, container := range containersList.GetContainers() {
+		c, err := h.getContainer(ctx, container.GetId())
+		if err != nil {
+			return nil, err
+		}
```
achrefbensaad: are we sure we want to quit at the first error ?

pkg/KubeArmorOperator/hook/crio.go
```diff
@@ -0,0 +1,106 @@
...
+
+	return container, nil
+}
+
+type containerInfo struct {
+	SandboxID   string     `json:"sandboxID"`
+	Pid         int        `json:"pid"`
+	RuntimeSpec specs.Spec `json:"runtimeSpec"`
+	Privileged  bool       `json:"privileged"`
+}
```
achrefbensaad: move this to the top of the file

achrefbensaad (commented):
1/2

KubeArmor/core/hook_handler.go
AbdelrahmanElawady: The reason for some of the fatal error handling is because if we just logged that there is an error and continued, some containers might go unnoticed for the duration of KubeArmor lifetime. Do you think it's better if we leave KubeArmor running and not monitoring some containers or restart KubeArmor and get all containers from the start (using some form of exiting)?

KubeArmor/core/hook_handler.go
AbdelrahmanElawady: hmm, `listenPath` is defined in the first line, so that wouldn't work?

Manik2708:
Hi! I have been trying to get context regarding OCI hooks. I was reading this PR and what I understood was: Hooks are being created and applied but how all the information related to container is sent to Kubearmor? What kind of scripts are saved in `usr/share/kubearmor/hooks`? Also how details related to containers are being fetched from hooks? Will appreciate your help!

pkg/KubeArmorOperator/hook/crio.go
rootxrishabh: It only fails if crio socket is unavailable, but handler won't start without socket. WDYT?

KubeArmor/core/hook_handler.go
rootxrishabh: @achrefbensaad, if this reordering correct?

KubeArmor/core/hook_handler.go
rootxrishabh: I think we should keep kubearmor running  even in case we skip a few containers at least for now. @daemon1024 WDYT?

pkg/KubeArmorOperator/hook/main.go
```diff
@@ -0,0 +1,279 @@
...
+func runDetached() error {
+	// we need to make sure the process exits at some point
+	time.AfterFunc(1*time.Minute, func() {
+		log.Println("failed to get containers, process timed out")
+		os.Exit(1)
+	})
+	conn := waitOnKubeArmor()
+	defer conn.Close()
+
+	handler, err := newCRIOHandler(runtimeSocket)
```
rootxrishabh, line 79: @AbdelrahmanElawady, apologies for asking this quite late, but why get pre-created containers only for CRI-O and not containerd?

AbdelrahmanElawady: Support for containerd with NRI is done in this PR: https://github.com/kubearmor/KubeArmor/pull/1763
This PR is just for CRI-O.

rootxrishabh (commented):
Initial review - Changes are working as expected!

Environment
```yaml
K8s - kubeadm
runtime - CRIO
O.S - Ubuntu
Enforcer - AppArmor
```

daemon1024:
Succeeded by #1763

PR #1723, opened 2024-04-15
feat : Alert Throttling

Prateeknandle:
**Purpose of PR?**:
### Requirement of Alert Throttling:
When excessive alerts are generated and a particular user's use case does not require all the alerts but needs an overview of the alerts generated, then throttling is required. Since every user would have their own requirements, the conditions for the throttling are configurable. 

**NOTE : Throttling is done per container, it's not system wide.**

Users can configure the throttling condition through the config map or KubeArmorConfig in case of operator deployment.

### Configuration:

In [KubeArmorConfig](https://github.com/kubearmor/KubeArmor/tree/main/deployments/helm/KubeArmorOperator#kubearmorconfig-specification) we can add:
```
spec:
  alertThrottling: true  // enables throttling functioning
  maxAlertPerSec: 10      // Number of maximum alerts that could be generated per sec
  throttleSec: 30        // Number of seconds for which subsequent alerts would be dropped
```

`maxAlertPerSec` and `throttleSec` are configurable, their default values are `10` and `30`, respectively.

If we set `maxAlertPerSec = 10` and `throttleSec = 30`, then maximum 10 alerts would be generated in a second per container and if this condition is crossed for any particular container then for that container a throttled condition alert would be generated and then for next 30 seconds there won't be any alert generated for that container.

**Throttling Alert:**
```
ClusterName: default
HostName: prateek
NamespaceName: wordpress-mysql
PodName: wordpress-586468bf4f-8bf6x
Labels: app=wordpress
ContainerName: wordpress
ContainerID: a4e3d52aeda8a0256d3c8ad819ec1bc6b61e4fa6f6d68858196d5195b5765d9f
ContainerImage: docker.io/library/wordpress:4.8-apache@sha256:6216f64ab88fc51d311e38c7f69ca3f9aaba621492b4f1fa93ddf63093768845
Type: SystemEvent
Operation: AlertThreshold
Enforcer: BPFLSM
Result: Passed
DroppingAlertsInterval: 30
MaxAlertsPerSec: 2
Owner: map[Name:wordpress Namespace:wordpress-mysql Ref:Deployment]
PPID: 0
UID: 0
```
This alert provides the information regarding throttling condition.

Fixes #

Yes,
1. tested with operator for BPFLSM enforcer
2. tested for apparmor enforcer
3. tested for audit alerts

1. **BPFLSM Enforcement:** We created a new eBPF map named `kubearmor_alert_throttle` which is required to hold the state of alerts throttling state per container. We add the containers entry in kernel-space itself. We remove the containers data in user-space for which the enforcement policies are removed.
_PS : We tried to handle deletion of used containers data in kernel-space itself but currently it's not possible._
2. **Apparmor Enforcement:** We used the same eBPF map `kubearmor_alert_throttle` to store the containers state wrt throttling. We add and delete this state data in kernel-space itself. 
_PS : here we delete the conatiners state when the container is being deleted, which is different for BPFLSM, where we delete the state data when there is no enforcement policy further exist for container._
3. **Audit based policies:** For audit alerts we handle the throttling in Feeder (completely in user-space). We created a golang map to handle throttling state per container. This map is updated in `core` package based on the CRI used.

**Checklist:**
- [ ] Commit has unit tests


KubeArmor/feeder/feeder.go
```diff
@@ -744,3 +759,66 @@ func loadTLSCredentials(ip string) (credentials.TransportCredentials, error) {
 	creds, err := cert.NewTlsCredentialManager(&tlsConfig).CreateTlsServerCredentials()
 	return creds, err
 }
+
+func bpfKtimeGetNs() uint64 {
```
daemon1024: Let's move it to common/utils

KubeArmor/BPF/shared.h
```diff
@@ -297,6 +312,83 @@ static __always_inline u32 init_context(event *event_data) {
...
+  }
+
+  u64 throttle_sec = (u64)get_kubearmor_config(_THROTTLE_SEC); 
+  u64 throttle_nsec = throttle_sec * 1000000000L; 
+  u64 maxAlert = (u64)get_kubearmor_config(_MAX_ALERT_PER_SEC); 
+
+  if (state->throttle) {
+    u64 time_difference = current_timestamp - state->first_event_timestamp;
+    if (time_difference < throttle_nsec) {
+      bpf_ringbuf_discard(event_data, 0);
```
daemon1024: Can we avoid usage of bpf_ringbuf_discard? Is it possible to only reserve for submission?

We have seen instances of discard api overloading kernel user space interaction

KubeArmor/common/common.go
```diff
@@ -28,6 +28,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// OuterKey struct
+type OuterKey struct {
+	PidNs uint32
+	MntNs uint32
+}
+
```
daemon1024, line 36: Usage of Global Variable Map 🤔, not sure if there's any better way to handle due to cross module usage

// @DelusionalOptimist

daemon1024 (commented):
@Prateeknandle great work. The changes are clean and easy to understand 👌

I have some open ended suggestions in review. Let's think about it and discuss them.

KubeArmor/BPF/shared.h
Prateeknandle: shouldn't we discard the [reserved space](https://github.com/kubearmor/KubeArmor/blob/main/KubeArmor/BPF/shared.h#L602), otherwise when we do [return retval](https://github.com/kubearmor/KubeArmor/pull/1723/files/20fd95c4c3f8ef9ec08d7740e126d51f19981ca5#diff-bd0b2889e0990fa3966c8718f021a1911672564aadc75b9ccb12410f71cee7c6R700-R702) when we drop alerts, the reserved space will be empty and might create mem overflow?

daemon1024:
need to push bpflsm object file

daemon1024 (approved):
Tried out PR locally 👌🏽 

LGTM. Awesome Clean Work 🙌🏽

rksharma95 (approved):
tested locally LGTM

PR #1753, opened 2024-05-14
fix(core) : owner information in logs

Prateeknandle:
**Purpose of PR?**:
When the pod is not available, we should not update the `dm.OwnerInfo` map. 
The variables which stores owner info were storing info from previous pod, when current pod was not available. Then the `dm.OwnerInfo` map was updated with wrong owner info from previous execution. 

Fixes #

no

on gcp cluster

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/KubeArmor/core/kubeUpdate.go
+++ b/KubeArmor/core/kubeUpdate.go
@@ -550,9 +550,6 @@ func (dm *KubeArmorDaemon) HandleUnknownNamespaceNsMap(container *tp.Container) 
 
 // WatchK8sPods Function
 func (dm *KubeArmorDaemon) WatchK8sPods() {
-	var controllerName, controller, namespace string
-	var err error
-
 	nodeName := os.Getenv("KUBEARMOR_NODENAME")
 	if nodeName == "" {
 		nodeName = cfg.GlobalCfg.Host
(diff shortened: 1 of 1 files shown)


KubeArmor/core/kubeUpdate.go
```diff
@@ -591,29 +588,41 @@ func (dm *KubeArmorDaemon) WatchK8sPods() {
...
+						Namespace: namespace,
+					}
+
+					dm.OwnerInfo[pod.Metadata["podName"]] = owner
+					podOwnerName = controllerName
 				}
-				_, err := K8s.K8sClient.CoreV1().Pods(namespace).Get(context.Background(), event.Object.ObjectMeta.Name, metav1.GetOptions{})
-				if err == nil && (event.Type == "MODIFIED" || event.Type != "DELETED") {
+
+				_, err = K8s.K8sClient.CoreV1().Pods(namespace).Get(context.Background(), event.Object.ObjectMeta.Name, metav1.GetOptions{})
```
DelusionalOptimist, line 612: Looks good to me.
NIT: Can you just add a comment in code as to why we need to check the pod's existence?

DelusionalOptimist (approved):

daemon1024 (approved):

PR #1763, opened 2024-05-22
feat: Add support for containerd and crio hooks

AbdelrahmanElawady:
**Purpose of PR?**:
Support running KubeArmor with containerd using OCI hooks. It need NRI to be enabled on the cluster and [hook-injector](https://github.com/containerd/nri/tree/main/plugins/hook-injector) plugin to be running.

No

Just the getting started example.

**Checklist:**
- [ ] Commit has unit tests


rootxrishabh:
NRI was added in [1674](https://github.com/kubearmor/KubeArmor/pull/1674)

pkg/KubeArmorOperator/hook/main.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func run(state specs.State) error {
```
rksharma95, line 114: need some explanation to understand this, looking for answer of a few questions:
- what's this detached process?
- why it need kubearmor information.

pkg/KubeArmorOperator/cmd/snitch-cmd/main.go
```diff
@@ -130,6 +135,14 @@ func snitch() {
 		Logger.Errorf("Not able to detect runtime")
 		os.Exit(1)
 	}
+	ociHooksLabel := "no"
+	if runtime == "cri-o" || EnableOCIHooks {
+		ociHooksLabel = "yes"
+		if err := applyCRIOHook(socket); err != nil {
```
rksharma95: just to confirm, it's genric way to check OCI hook support for any CRI not limited to cri-o right!

rksharma95 (commented):
@AbdelrahmanElawady can you provide a diagram that explains the design in detail?

pkg/KubeArmorOperator/cmd/snitch-cmd/main.go
rootxrishabh: @rksharma95, I think this was done as NRI support for containerd was not available at the time this PR was created. CRI-O supports hooks out of the box.

pkg/KubeArmorOperator/hook/main.go
rootxrishabh: @rksharma95, detached process is when hook actually starts before kubearmor. Ka.sock is owned by kubearmor and hook waits for socket to start. Needs kubearmor information to know when kubearmor starts.

pkg/KubeArmorOperator/cmd/snitch-cmd/main.go
rksharma95: so do we need to handle anything here?

pkg/KubeArmorOperator/hook/main.go
```diff
@@ -0,0 +1,285 @@
...
+	specBytes, err := os.ReadFile(filepath.Join(state.Bundle, "config.json"))
+	if err != nil {
+		// revert back to annotations
+		containerName := state.Annotations["io.kubernetes.container.name"]
+		appArmorProfile = strings.TrimPrefix(
+			state.Annotations[fmt.Sprintf("container.apparmor.security.beta.kubernetes.io/%s", containerName)],
+			"localhost/",
+		)
+		isKubeArmor = containerName == "kubearmor"
+	} else {
```
rootxrishabh, line 155: @AbdelrahmanElawady else would trigger in case receive an error while reading config.json?
If yes, then should we expect to read `specBytes` on line 157.

AbdelrahmanElawady: else would run if err == nil. error handling is done in line 147

rootxrishabh: oops my bad, thanks for the clarification 🙌

rootxrishabh:
@AbdelrahmanElawady, looking at the hook binary. I think we stop depending on hook stdin and call runtime socket for container info after kubearmor starts. I am not sure whats the logic here. Can you please explain?

Is this done so to get containers started before kubearmor?

AbdelrahmanElawady:
@rootxrishabh it's been a while so might not remember correctly but I think we use stdin to figure out container id, status, pid and the bundle path which are all part of the [state](https://github.com/opencontainers/runtime-spec/blob/main/runtime.md#state) object sent in stdin. this is according to OCI spec as mentioned [here](https://github.com/opencontainers/runtime-spec/blob/main/config.md#posix-platform-hooks).
> The [state](https://github.com/opencontainers/runtime-spec/blob/main/runtime.md#state) of the container MUST be passed to hooks over stdin so that they may do work appropriate to the current state of the container.

We call runtime socket for previous containers but this is separate than hook stdin

pkg/KubeArmorOperator/cmd/snitch-cmd/main.go
```diff
@@ -169,6 +183,64 @@ func snitch() {
 	}
 }
 
+func applyCRIOHook(socket string) error {
+	// TODO: hook path should be fetched from container runtime. This is the default path
+	hookDir := "/usr/share/containers/oci/hooks.d/"
```
rksharma95, line 203: in case of containerd it should be `/etc/containers/oci/hooks.d` 

how it will be handled? should we support a flag with operator to accept this path as we also need to mount this path with snitch, it will be an issue if we dynamically support multiple paths.

pkg/KubeArmorOperator/internal/controller/resources.go
```diff
@@ -328,6 +354,14 @@ func deploySnitch(nodename string, runtime string) *batchv1.Job {
 								Name:      "seccomp-path",
 								MountPath: "/var/lib/kubelet/seccomp",
 							},
+							{
```
rksharma95, line 363: snitch has a command line flag as well to enable OCI Hook, let's handle that as well.

rksharma95 (changes requested):

pkg/KubeArmorOperator/cmd/snitch-cmd/main.go
rootxrishabh: We can auto set this path under the condition of which runtime socket is detected. Right?

rksharma95: this is what I'm concerned with, it will need mounting all possible path with the snitch and if that not exist snitch will be reconciled again after that mount points handled.

rootxrishabh: CRI-O also supports `/etc/containers/oci/hooks.d/` as a path for hook configuration. I think we can use this for both crio and containerd.

pkg/KubeArmorOperator/internal/controller/resources.go
rksharma95: this is not being handled as of now.

rootxrishabh (dismissed):
LGTM 🚀 

Great work @AbdelrahmanElawady !

rootxrishabh (approved):

rksharma95 (approved):

Aryan-sharma11 (approved):

rksharma95:
PR tested locally on both containerd and crio setup. CI is failing due to docker pull limits and few tests are flaky.

PR #1802, opened 2024-07-11
fix: use custom action for pushing charts

DelusionalOptimist:


rksharma95 (approved):

daemon1024 (approved):

kranurag7:
@DelusionalOptimist I was looking into fixing it after noticing the failure on main later today but you already went ahead :) (good work! 🙇🏻) 

I saw your fork, I think this can be solved if we move to storing charts in OCI registry. It's stable from helm side as well. It fixes the auth part and the also the point that we want to store the packaged chart `.tgz` in `x` registry which is located somewhere else on the internet. Are you open to that? 

In the starting, we can think about moving at least latest chart from main to OCI registry and later we can move stable ones as well.

PR #1822, opened 2024-07-30
fix(enforcement/host): fix default posture alerts and streamline host profile handling

daemon1024:


Aryan-sharma11:
partially fixed in #1861

PR #1833, opened 2024-08-11
feat: mount procfs instead of using hostpid (#1688)

yp969803:
Signed-off-by: Yash Patel <yp969803@gmail.com>

**Purpose of PR?**:

Fixes #1688 

**Does this PR introduce a breaking change?**

**If the changes in this PR are manually verified, list down the scenarios covered:**:

**Additional information for reviewer?** :
_Mention if this PR is part of any design or a continuation of previous PRs_

**Checklist:**
- [X] Bug fix. Fixes #<issue number>
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] PR Title follows the convention of  `<type>(<scope>): <subject>`
- [ ] Commit has unit tests
- [ ] Commit has integration tests

PR #1834, opened 2024-08-11
fix: Add non-root user into KubeArmor container(#1401)

### LINKED ISSUE #1401: Add non-root user into KubeArmor container
## Feature Request

Currently, KubeArmor daemonset runs inside a container (with alpine base image) as the root user. To enhance the security and maintain best practices, we need to modify the containerized environment so that the application runs as a non-root user.  We should create a dedicated non-root user within the container to run the KubeArmor application. This user should have minimal privileges necessary for the KubeArmor to function properly.

For rest of the KubeArmor applications including `kubearmor relay server`, `kubearmor controller`, and `kubearmor init container` it will be handled as part of #1201

```[tasklist]
### Tasks
- [ ] Create a non-root user with required capabilities within the target alpine container
- [ ] Test and Validate KubeArmor under non-root user context.
```

yp969803:
Signed-off-by: Yash Patel <yp969803@gmail.com>

**Purpose of PR?**:

Fixes #1401

**Checklist:**
- [X] Bug fix. Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/deployments/get/objects.go
+++ b/deployments/get/objects.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
 )
 
 // GetServiceAccount Function
@@ -413,6 +414,8 @@ func GenerateDaemonSet(env, namespace string) *appsv1.DaemonSet {
 							Image:           "kubearmor/kubearmor:stable",
 							ImagePullPolicy: "Always",
 							SecurityContext: &corev1.SecurityContext{
+								RunAsUser:  ptr.To(int64(1000)),
+								RunAsGroup: ptr.To(int64(1000)),
 								Privileged: &privileged,
 								Capabilities: &corev1.Capabilities{
 									Drop: []corev1.Capability{
--- a/KubeArmor/monitor/syscalls_arm64.go
+++ b/KubeArmor/monitor/syscalls_arm64.go
@@ -63,7 +63,7 @@ const (
 	SocketConnect = 462
 	SocketAccept  = 463
 
-	Capable = 464
+	Capable   = 464
 	DropAlert = 0
 )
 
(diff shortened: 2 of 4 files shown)


yp969803:
@rksharma95 can you review the pr

kranurag7 (commented):
@yp969803 Can you please also update the helm chart for this? 

As of now, I'm not sure if this will work or not. Last time, I checked with non-root user the controller was failing. Having said that, we have refactored the controller a lot so please let me know if you've already verified the changes at runtime. 

EDIT:
Also, I think it would be great if we have something like this `uid=65532(nonroot) gid=65532(nonroot) groups=65532(nonroot)` _(not a blocker for this PR and can be done as a follow up)_

deployments/get/objects.go
```diff
@@ -413,6 +414,8 @@ func GenerateDaemonSet(env, namespace string) *appsv1.DaemonSet {
 							Image:           "kubearmor/kubearmor:stable",
 							ImagePullPolicy: "Always",
 							SecurityContext: &corev1.SecurityContext{
+								RunAsUser:  ptr.To(int64(1000)),
+								RunAsGroup: ptr.To(int64(1000)),
```
daemon1024, line 418: Let's avoid this and change the Dockerfile instead to use non root user

daemon1024 (changes requested):

daemon1024:
Please check with a containerd environment locally if it works or not.

daemon1024:
Please fill the PR template completely.
I don't think this PR was validated at all locally, would appreciate PRs being tested locally before being opened.
Same for #1833. We already see that tests fail so it was not tested out.

rksharma95:
@yp969803 one more point i would like to add is, make sure you first validate if non-root user works with apparmor enforcer.

deployments/get/objects.go
yp969803: Dockerfile already have configuration to use non-root user

rksharma95: only for ubi based image not for alpine one.
PTAL here https://github.com/kubearmor/KubeArmor/blob/6cd782f39ab4280312ecdfd9f228df638aee593c/Dockerfile#L43

PR #1841, opened 2024-08-14
refactor(operator): remove config empty check

carlosrodfern:
The config elements, though optional, are not ptr, so they are initialized with the respective type default values. As result, strconv.Format... will never return an empty string, so the check is unnecessary. I also found that the setting of ThrottleSec was depending on MaxAlertPerSec, but because that one is never empty, the dependency wasn't actually effective anyways.

**Purpose of PR?**:

Clean up

I stumbled into this code while troubleshooting something else. This is a clean up and doesn't impact existing logic.

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/pkg/KubeArmorOperator/internal/controller/cluster.go
+++ b/pkg/KubeArmorOperator/internal/controller/cluster.go
@@ -806,25 +806,19 @@ func UpdateConfigMapData(config *opv1.KubeArmorConfigSpec) bool {
 		}
 	}
 	AlertThrottlingEnabled := strconv.FormatBool(config.AlertThrottling)
-	if AlertThrottlingEnabled != "" {
-		if common.ConfigMapData[common.ConfigAlertThrottling] != AlertThrottlingEnabled {
-			common.ConfigMapData[common.ConfigAlertThrottling] = AlertThrottlingEnabled
-			updated = true
-		}
+	if common.ConfigMapData[common.ConfigAlertThrottling] != AlertThrottlingEnabled {
+		common.ConfigMapData[common.ConfigAlertThrottling] = AlertThrottlingEnabled
+		updated = true
 	}
 	MaxAlertPerSec := strconv.FormatInt(int64(config.MaxAlertPerSec), 10)
-	if MaxAlertPerSec != "" {
-		if common.ConfigMapData[common.ConfigMaxAlertPerSec] != MaxAlertPerSec {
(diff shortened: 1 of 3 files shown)


carlosrodfern:
I'm unable to see why `ci-test-ginkgo` failed with this change. Any idea I can work on?

What I see is that it is expecting an alert to show up, but the logs simply post: `time="2024-08-15T20:14:22Z" level=info msg="event timeout"`

DelusionalOptimist:
> I'm unable to see why `ci-test-ginkgo` failed with this change. Any idea I can work on?
> 
> What I see is that it is expecting an alert to show up, but the logs simply post: `time="2024-08-15T20:14:22Z" level=info msg="event timeout"`

@carlosrodfern yeah... The failure isn't relevant. Tests are a little flakey on the GitHub runners.
`event timeout` indicates that it was trying to look for a matching alert on the gRPC stream that is created for getting alerts from KubeArmor but it didn't find one. These scenarios pass when run locally or on our self hosted runners but due to some performance issues fail on GitHub runners. We have tried doing multiple things in the past but haven't had much success in making the tests less flakier.

pkg/KubeArmorOperator/internal/controller/cluster.go
```diff
@@ -806,25 +806,29 @@ func UpdateConfigMapData(config *opv1.KubeArmorConfigSpec) bool {
...
-	if MaxAlertPerSec != "" {
+	if config.MaxAlertPerSec != 0 {
 		if common.ConfigMapData[common.ConfigMaxAlertPerSec] != MaxAlertPerSec {
 			common.ConfigMapData[common.ConfigMaxAlertPerSec] = MaxAlertPerSec
 			updated = true
 		}
+	} else {
+		common.ConfigMapData[common.ConfigMaxAlertPerSec] = common.DefaultMaxAlertPerSec
+		updated = true
 	}
```
daemon1024:
```suggestion
	if config.MaxAlertPerSec == 0 {
                        MaxAlertPerSec = common.DefaultMaxAlertPerSec
	} 
	if common.ConfigMapData[common.ConfigMaxAlertPerSec] != MaxAlertPerSec {
			common.ConfigMapData[common.ConfigMaxAlertPerSec] = MaxAlertPerSec
			updated = true
	}
```

daemon1024 (changes requested):

daemon1024 (approved):

rksharma95 (approved):

PR #1881, opened 2024-10-22
feat(operator): fetch cluster name from providers

rootxrishabh:
**Purpose of PR?**:

This PR adds fetching cluster name from cluster providers.

No

**Checklist:**
- [X] New feature (non-breaking change which adds functionality)
- [ ] Commit has unit tests

### DIFF
--- a/KubeArmor/core/kubeUpdate.go
+++ b/KubeArmor/core/kubeUpdate.go
@@ -2761,6 +2761,10 @@ func (dm *KubeArmorDaemon) WatchConfigMap() cache.InformerSynced {
 			if cm, ok := obj.(*corev1.ConfigMap); ok && cm.Namespace == cmNS {
 				cfg.GlobalCfg.HostVisibility = cm.Data[cfg.ConfigHostVisibility]
 				cfg.GlobalCfg.Visibility = cm.Data[cfg.ConfigVisibility]
+				cfg.GlobalCfg.Cluster = cm.Data[cfg.ConfigCluster]
+				dm.NodeLock.Lock()
+				dm.Node.ClusterName = cm.Data[cfg.ConfigCluster]
+				dm.NodeLock.Unlock()
 				if _, ok := cm.Data[cfg.ConfigDefaultPostureLogs]; ok {
 					cfg.GlobalCfg.DefaultPostureLogs = (cm.Data[cfg.ConfigDefaultPostureLogs] == "true")
 				}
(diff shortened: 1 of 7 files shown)


pkg/KubeArmorOperator/internal/controller/resources.go
```diff
@@ -469,6 +472,152 @@ func (clusterWatcher *ClusterWatcher) deployControllerDeployment(deployment *app
...
+		}
+	} else if provider == "eks" {
+		clusterWatcher.Log.Infof("Provider is EKS")
+		if clusterName, err := clusterWatcher.fetchClusterNameFromAWS(); err != nil {
+			clusterWatcher.Log.Warnf("Cannot fetch cluster name for EKS %s", err.Error())
+		} else {
+			return clusterName
+		}
+	}
+	// } else if provider == "aks" {
```
rksharma95: any particular reason to skip aks handling ??

rootxrishabh: AKS IMDS does not provide cluster name. Had kept it there as the next action item, but since it's not there. I will be removing it in the next commit.

Prateeknandle (commented):
what happens in case there is no operator, should we have a fallback logic in kubearmor?
cc: @daemon1024

rootxrishabh:
Yes, cluster name goes to "default".

Prateeknandle:
> Yes, cluster name goes to "default".

@rootxrishabh i meant shouldn't we fetch cluster name from providers in case operator is not there?

rksharma95:
@rootxrishabh  can you please add args handling in operator helm chart? It should support any args list values.

Prateeknandle (dismissed):
Tested on both GKE and AWS, working as expected.

rksharma95:
@Prateeknandle @rootxrishabh does kubearmor uses `clusterName` value from cofigmap?

Prateeknandle:
> @Prateeknandle @rootxrishabh does kubearmor uses `clusterName` value from cofigmap?

No, sry totally forgot this point, @rootxrishabh we should update cluster name in kubearmor, it should get updated based on conifgmap value.
https://github.com/Prateeknandle/KubeArmor/blob/main/KubeArmor/core/kubeUpdate.go#L2749

check whether correct cluster name is reflected in alerts/logs or not.

Prateeknandle (dismissed):

KubeArmor/core/kubeUpdate.go
```diff
@@ -2761,6 +2761,7 @@ func (dm *KubeArmorDaemon) WatchConfigMap() cache.InformerSynced {
 			if cm, ok := obj.(*corev1.ConfigMap); ok && cm.Namespace == cmNS {
 				cfg.GlobalCfg.HostVisibility = cm.Data[cfg.ConfigHostVisibility]
 				cfg.GlobalCfg.Visibility = cm.Data[cfg.ConfigVisibility]
+				cfg.GlobalCfg.Cluster = cm.Data[cfg.ConfigCluster]
```
rksharma95, line 2764: we need to update `dm.Node.ClusterName` as well to reflect the changes for the node.

rksharma95 (approved):
LGTM

daemon1024 (approved):

PR #1891, opened 2024-11-10
Improved container and host policy fuzzer

prady0t:
**Purpose of PR?**:

Fixing bugs in fuzzers.
Seed value changed from yaml to json for checking for generation of invalid json inputs. `ContainerPolicy` and `HostPolicy` now returns err for generated errors.

### DIFF
--- a/KubeArmor/core/containerPolicy_fuzz_test.go
+++ b/KubeArmor/core/containerPolicy_fuzz_test.go
@@ -5,6 +5,8 @@ package core
 import (
 	"context"
 	"testing"
+	"encoding/json"
+	tp "github.com/kubearmor/KubeArmor/KubeArmor/types"
 
 	"github.com/kubearmor/KubeArmor/KubeArmor/policy"
 	pb "github.com/kubearmor/KubeArmor/protobuf"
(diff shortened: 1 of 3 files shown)


prady0t:
FuzzHostPolicy still needs to be updated based on the review. @daemon1024 would love your reviews here

prady0t:
Inside FuzzContainerPolicy, I've added a new case : 
```
len(policyEvent.Object.Spec.Selector.MatchLabels) == 0 && res.Status == pb.PolicyStatus_Invalid
```
Based on https://github.com/kubearmor/KubeArmor/blob/2cfc2e24d13eac6d857ad53723f9bd2025892b95/KubeArmor/core/unorchestratedUpdates.go#L234C13-L234C33

This also solved the error we were getting earlier with this value : 
```
"{\"oBjeCt\":{\"metAdAtA\":{\"nAme\":\"0\"}}}"
```
There are a few more cases where fuzzer throws `PolicyStatus_Invalid` status, but are very hard for fuzzers to generate values for such cases. 

https://github.com/kubearmor/KubeArmor/blob/2cfc2e24d13eac6d857ad53723f9bd2025892b95/KubeArmor/core/unorchestratedUpdates.go#L238

What do you think of these conditions?

DelusionalOptimist:
> There are a few more cases where fuzzer throws PolicyStatus_Invalid status, but are very hard for fuzzers to generate values for such cases.

@prady0t Can you share more on why it is hard? Is it because it's not possible to add a new label through them or have the enforcer based check?

prady0t:
Yes because it's hard for fuzzer to generate such conditions. We can still add this condition inside fuzzer if we want

DelusionalOptimist:
> Yes because it's hard for fuzzer to generate such conditions. We can still add this condition inside fuzzer if we want

I think it should be fine. Ideally these should be handled by policy validation.

DelusionalOptimist (approved):

DelusionalOptimist:
Please rebase @prady0t.

PR #1936, opened 2025-01-08
fix(operator): snitch conditional mounts

### LINKED ISSUE #1540: Talos linux support
## Feature Request

**Short Description**

Support for [Talos Linux](https://www.talos.dev/) needs to be validated. Validation has to be done for:
- [x] enforcement supported? What is the LSM used?
- [x] Audit/Observability supported
- [x] Paste the output of karmor probe

As an FYI, the tasks involved would be:
1. Setup a Talos k8s cluster 
2. Install KubeArmor
3. Check sample workloads and verify if following things are working:
   * Policy Enforcement
   * Alerts/Telemetry
4. Get `karmor probe` output reference and attach in this issue
5. Update [Kubearmor support matrix](https://github.com/kubearmor/KubeArmor/blob/main/getting-started/support_matrix.md)

rksharma95:
**Purpose of PR?**:

Snitch Job would be failed to deploy if mount hostpath not exists on nodes with readonly filesystem. 

```
Warning  Failed  11s                
kubelet  Error: failed to generate container "4cb6ea54ef292997cb29e5670d63830015704ec7c1f8fe87e55968280a54e835" 
spec: failed to apply OCI options: failed to mkdir "/etc/apparmor.d/": mkdir /etc/apparmor.d/: read-only file system
```

This PR adds changes to check if snitch job failed due to mount failure in both non-exist path and read-only file system cases. If so re-deploy snitch without these conflicting mounts. 

No

**Checklist:**
- [x] Bug fix.
- [ ] Commit has unit tests

### DIFF
--- a/pkg/KubeArmorOperator/internal/controller/cluster.go
+++ b/pkg/KubeArmorOperator/internal/controller/cluster.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"regexp"
 	"slices"
 	"sort"
 	"strconv"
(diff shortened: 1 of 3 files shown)


pkg/KubeArmorOperator/internal/controller/cluster.go
```diff
@@ -103,6 +104,136 @@ func NewClusterWatcher(client *kubernetes.Clientset, log *zap.SugaredLogger, ext
...
+		return matches[1], true
+	}
+	return "", false
+}
+
+func (clusterWatcher *ClusterWatcher) checkJobStatus(job, runtime, nodename string) {
+	defer func() {
+		clusterWatcher.Log.Infof("checkJobStatus completed for job: %s", job)
+	}()
+
```
daemon1024, line 135: Let's use a 5 minute timeout here just in case this does not exit

daemon1024 (commented):
LGTM, One comment, WDYT?

pkg/KubeArmorOperator/internal/controller/cluster.go
rksharma95: This executes only if job creation is successful.
We check first if job exists, I think It should be okay without sleep as well.

rksharma95: I believe when it gets success response then the job resource has been created persisted in etcd. can you think of any edge case here and we should have a delay in that case.

daemon1024: I mean inside the infinite for loop not in defer function.

rksharma95: I got that, 
I just wanted to mention It should exit in all the cases, In the infinite loop if are you suggesting a delay I think we can have that but 5 minutes is long as in valid cases it will add a long delay.

daemon1024 (approved):
LGTM 👌

rootxrishabh (approved):
LGTM!

Aryan-sharma11 (approved):

PR #1955, opened 2025-01-25
feat(ci): add script to validate each commit in PR

### LINKED ISSUE #616: Add Github action to check if each commit is compiling

inosmeet:
**Purpose of PR?**:
Implement feature that ensures that each commit of PR is separately compiles.
Closes: #616 

Manually tested in fork, with both appropriate commits that should and does compile as well as invalid commits that should and does fail.

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/.github/workflows/ci-validate-commit.yaml
+++ b/.github/workflows/ci-validate-commit.yaml
@@ -18,7 +18,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.23
+          go-version-file: 'KubeArmor/go.mod'
 
       - name: Checkout code
         uses: actions/checkout@v3
--- a/.github/workflows/ci-validate-commit.yaml
+++ b/.github/workflows/ci-validate-commit.yaml
@@ -1,0 +1,48 @@
+name: Validate Commits
+
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  validate-commits:
+    name: Check if KubeArmor compiles for every commit
+    runs-on: ubuntu-20.04
+    timeout-minutes: 60
+    steps:
+      - name: Configure git
+        run: |
+          git config --global user.name "GitHub Actions"
+          git config --global user.email "github-actions@users.noreply.github.com"
+
+      - name: Install Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.23
+
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Validate each commit
+        run: |
+          # Loop through each commit in the PR from oldest to newest
+          for commit in $(git rev-list --reverse ${{ github.event.pull_request.base.sha }}..${{ github.sha }}); do
+            echo "========================================="
+            echo "Checking out commit $commit..."
+            git checkout $commit
+
+            # Attempt to build the project
+            cd KubeArmor
+            echo "Building KubeArmor for commit $commit..."
+            if make build; then
+              echo "✅ Commit $commit compiled successfully."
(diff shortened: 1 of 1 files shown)


inosmeet:
cc @Ankurk99 @achrefbensaad

inosmeet:
Hey @daemon1024!
Can I get a review?
Thanks!

.github/workflows/ci-validate-commit.yaml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+      - name: Configure git
+        run: |
+          git config --global user.name "GitHub Actions"
+          git config --global user.email "github-actions@users.noreply.github.com"
+
+      - name: Install Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.23
```
Ankurk99: Consider using `go-version-file: KubeArmor/go.mod` for defining go version

.github/workflows/ci-validate-commit.yaml
```diff
@@ -0,0 +1,48 @@
...
+          for commit in $(git rev-list --reverse ${{ github.event.pull_request.base.sha }}..${{ github.sha }}); do
+            echo "========================================="
+            echo "Checking out commit $commit..."
+            git checkout $commit
+
+            # Attempt to build the project
+            cd KubeArmor
+            echo "Building KubeArmor for commit $commit..."
+            if make build; then
+              echo "✅ Commit $commit compiled successfully."
```
Ankurk99, line 40: I would also recommend showing the error in the logs

Ankurk99 (changes requested):
Hey, thanks for the PR. I have a few suggestions.
Also, did you test these changes anywhere?

inosmeet:
> did you test these changes anywhere?

Thanks for reviewing!
Yeah, I manually tested these in my fork.
I followed these steps:

1. Made a PR to my fork which included a couple of commits that were appropriate (which passed).
2. On top of the appropriate commits, I deliberately made a few commits that had some faults (which failed as the should).

I'm not sure whether there should be a test for this?

.github/workflows/ci-validate-commit.yaml
inosmeet: GitHub Actions captures all logs from `make build`, do we need to add manual logs?
for example, here's the log from GH action when i was testing manually:
```
Error: ./main.go:40:68: syntax error: unexpected newline in argument list; possibly missing comma or )
Error: ./main.go:41:4: syntax error: unexpected return at end of statement
make: *** [Makefile:26: build] Error 1
❌ Commit e52b12d577936188e5b7aea2c9ddb30fb6b60b17 failed to compile!
=========================================
Error: Process completed with exit code 1.
```

PR #1999, opened 2025-03-11
feat: adding support for excluding labels in csp and ksp for endpoint matching

Prateeknandle:
**Purpose of PR?**:
Support for adding condition for selecting workloads in the policies (ksp & csp) under `matchExpressions`.

Fixes #

**Checklist:**
- [ ] Commit has unit tests


daemon1024 (dismissed):
New set of tests for KSP 

matchExpressions
- label In,NotIn

One more test case
where both of them exist? (Need an answer to this)

KubeArmor/core/kubeUpdate.go
```diff
@@ -1011,32 +1015,34 @@ func matchClusterSecurityPolicyRule(policy tp.SecurityPolicy) bool {
...
 	dm.SecurityPoliciesLock.RLock()
 	defer dm.SecurityPoliciesLock.RUnlock()
 
 	secPolicies := []tp.SecurityPolicy{}
 
 	for _, policy := range dm.SecurityPolicies {
-		if kl.MatchIdentities(policy.Spec.Selector.Identities, identities) || kl.ContainsElement(policy.Spec.Selector.NamespaceList, namespaceName) || matchClusterSecurityPolicyRule(policy) {
+		updateNamespaceListforCSP(policy)
+		// match ksp || csp
+		if (kl.MatchIdentities(policy.Spec.Selector.Identities, endPoint.Identities) && kl.MatchExpIdentities(policy.Spec.Selector, endPoint.Identities)) ||
```
rksharma95, line 1044: when both matchLabel and matchExpression present target workload should satisfy both conditions right!

KubeArmor/core/kubeUpdate.go
```diff
@@ -1249,18 +1294,35 @@ func (dm *KubeArmorDaemon) CreateSecurityPolicy(policyType string, securityPolic
...
+			} else if matchExpression.Key == LabelKey {
+				if matchExpression.Operator == InOperator {
+					for _, label := range matchExpression.Values {
+						secPolicy.Spec.Selector.MatchExpIdentities = append(secPolicy.Spec.Selector.MatchExpIdentities, label)
+					}
+				} else if matchExpression.Operator == NotInOperator {
+					for _, label := range matchExpression.Values {
+						secPolicy.Spec.Selector.NonIdentities = append(secPolicy.Spec.Selector.NonIdentities, label)
+					}
+				}
```
rksharma95: for CSP `In` and `NotIn` operators are being handled differently than KSP any specific reason? iinr in case of ksp `NotIn` rules are being added only if `In` rules are not present which is not the case here with csp.

KubeArmor/core/kubeUpdate.go
```diff
@@ -1070,26 +1076,40 @@ func (dm *KubeArmorDaemon) UpdateSecurityPolicy(action string, secPolicyType str
...
+						}
+						if new {
+							endPoint.SecurityPolicies = append(endPoint.SecurityPolicies, secPolicy)
 						}
-					}
-					if new {
-						endPoint.SecurityPolicies = append(endPoint.SecurityPolicies, secPolicy)
 					}
 				} else if action == "MODIFIED" {
+					// in case new labels are added in the policy, check if identities match, if yes, add policy in endPoint
```
rksharma95, line 1095: ~~it's important to document this behaviour, because when target workloads changed by new selectors it is natural to assume that policy is no more valid for previously targeted workloads but the rules will not be reverted for those workloads right! just that if any rules updated those will not be valid for those workloads right!~~

my bad I missed that. we're updating the policies for each workloads so this should not be an issue.

KubeArmor/core/kubeUpdate.go
rksharma95: consequently this csp policy will target a workload even though both labels are not present. btw logically it matches NotIn rule.
```yaml
kind: KubeArmorClusterPolicy
metadata:
  name: csp-match-expr
spec:
  process:
    matchPaths:
      - path: /usr/bin/sleep
      - path: /bin/sleep
  action: Block    
  selector:
    matchExpressions:
      - key: label
        operator: In
        values:
        - not-exists
      - key: label
        operator: NotIn
        values:
        - also-not-exists
```

KubeArmor/core/kubeUpdate.go
Prateeknandle: yes, that's why we used AND operator here

KubeArmor/core/kubeUpdate.go
Prateeknandle: got it will update for csp, it should be same as ksp, good catch, thanks

Prateeknandle: updated

rksharma95: we need to document this as well as when both `matchLabel` and `matchExpression` present both rules should be matched to select a workload.

Prateeknandle: I've mentioned it here - https://github.com/kubearmor/KubeArmor/pull/1999/files#diff-e871007b073b341af07a21f370a61f1338dd6805746955d03cc674e5bb7acd51R173

rksharma95 (approved):
LGTM :+1: 
waiting for CI to pass.
@Prateeknandle please update the branch.

anurag-rajawat (approved):

PR #2108, opened 2025-07-15
docs(modelarmor): revamp landing page with use cases and flow diagrams

HighnessAtharva:
## This PR is a work-in-progress

**Purpose of PR?**:
This PR revamps the **ModelArmor documentation page** to include:

* Better-defined **use cases** highlighting real-world AI/ML security scenarios.
* **Flow diagrams** explaining attack vectors and how ModelArmor mitigates them.
* Improved **explanations and structure** to showcase what’s actually possible with ModelArmor.
* Cleaned up the existing content for clarity and relevance.

Fixes: *No issue was fixed; this is a content improvement and enhancement.*

---

**Does this PR introduce a breaking change?**
❌ No, this PR only affects documentation.

---

**If the changes in this PR are manually verified, list down the scenarios covered:**

* Validated updated ModelArmor documentation renders correctly on the website.
* Ensured all diagrams and use cases are properly linked and visually clear.
* Confirmed improved navigation and readability of the ModelArmor landing page.

---

**Additional information for reviewer?**:
This PR is part of the ongoing effort to make **ModelArmor documentation more developer-centric** and better showcase its capabilities. It builds on prior content but introduces fresh material for **AI/ML security storytelling** and user understanding.

---

**Checklist:**

* [ ] Bug fix. Fixes #<issue number>
* [x] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [x] This change requires a documentation update
* [x] PR Title follows the convention of  `<type>(<scope>): <subject>`
* [ ] Commit has unit tests
* [ ] Commit has integration tests

PR #2110, opened 2025-07-16
feat(nodeID): add node_id from machine_id variable

vishnusomank:
**Purpose of PR?**:

If worker nodes have the same hostname, it becomes difficult to manage the workloads associated with them. For example, if `node_1` and `node_2` both use the hostname `vmhost_1` and `workload_1` is running on both nodes, we cannot reliably differentiate them based on events alone. If `workload_1` is deleted from `node_2`, it will appear as if it is removed from `vmhost_1`, even though it is still running on `node_1`.

To reduce this discrepancy, we are adding a unique `node_id` along with container and node events sent via state events.

Fixes #

**Checklist:**
- [ ] Commit has unit tests


seswarrajan (dismissed):

KubeArmor/core/kubeArmor.go
```diff
@@ -489,9 +501,17 @@ func KubeArmor() {
...
 	}
+
 	dm.NodeLock.RLock()
 	kg.Printf("Node Name: %s", dm.Node.NodeName)
 	kg.Printf("Node IP: %s", dm.Node.NodeIP)
+	if dm.Node.NodeID == "" {
+		id, _ := os.ReadFile(cfg.GlobalCfg.MachineIDPath)
+		dm.Node.NodeID = strings.TrimSuffix(string(id), "\n")
+	}
+	dm.Node.NodeID = protectedID(dm.Node.NodeID, dm.Node.NodeName)
```
Aryan-sharma11, line 508: We are using a read lock above, but here we are updating the values.

nyrahul:
Hey @vishnusomank , can you explain how we are planning on getting machineID/nodeID? Based on the code, this is what I understand:
1. we read `/etc/machine-id`
2. execute `hostnamectl`

Question is, 
1. what happens if both these avenues fail? There are distros where hostnamectl and machine-id are not present.
2. Do we expect to do a hostpath mount of /etc/ in k8s env? or is this machine-id only used for VM based security?

vishnusomank:
> Hey @vishnusomank , can you explain how we are planning on getting machineID/nodeID? Based on the code, this is what I understand:
> 
> 1. we read `/etc/machine-id`
> 2. execute `hostnamectl`
> 
> Question is,
> 
> 1. what happens if both these avenues fail? There are distros where hostnamectl and machine-id are not present.
> 2. Do we expect to do a hostpath mount of /etc/ in k8s env? or is this machine-id only used for VM based security?

@nyrahul, 
We execute `hostnamectl` initially. As a fallback, we allow specifying a value for the `machineIDPath` flag so that we can read the machine ID from that path.

Currently, this is only used for VMs. This helps us better handle VMs with the same hostname connecting to the control plane by correctly mapping workloads (containers) to their parent nodes.

KubeArmor/core/kubeArmor.go
vishnusomank: addressed @Aryan-sharma11

Aryan-sharma11: we should use write lock here, right?  `dm.NodeLock.Lock()`

KubeArmor/types/types.go
```diff
@@ -230,6 +232,7 @@ type Log struct {
 	// host
 	ClusterName string `json:"clusterName,omitempty"`
 	HostName    string `json:"hostName"`
+	NodeID      string `json:"nodeID"`
```
Aryan-sharma11:
```suggestion
	NodeID      string `json:"nodeID, omitempty"`
```

Aryan-sharma11 (changes requested):

vishnusomank:
https://github.com/kubearmor/KubeArmor/pull/2110#discussion_r2213504009 do we need a lock here? i dont suppose we are writing to this from anywhere else @Aryan-sharma11

Aryan-sharma11:
> [#2110 (comment)](https://github.com/kubearmor/KubeArmor/pull/2110#discussion_r2213504009) do we need a lock here? i dont suppose we are writing to this from anywhere else @Aryan-sharma11

Agreed lock not needed here. 👍

Aryan-sharma11:
To fix CI @vishnusomank, kindly run `make run` in `kubearmor/kubearmor` directory, and add all the object files, it seems like they were deleted, in this PR https://github.com/kubearmor/KubeArmor/pull/2108.

Aryan-sharma11 (approved):

nyrahul (approved):
LGTM

PR #2294, opened 2025-12-05
refactor(core): add worker-pool for containerd event processing

### LINKED ISSUE #2289: Introduce Worker-Pool for Containerd Event Handling to Prevent Blocking and Improve Throughput
### Problem

**MonitorContainerdEvents** currently processes events synchronously.
Since **UpdateContainerdContainer** involves containerd RPC calls, PID/NS lookups, and multiple locks, it can block the subscription loop during heavy workloads.

### This leads to:

subscription loop blocking

reduced throughput under bursts

increased lock contention

delayed event handling

### Proposed Enhancement

Introduce a bounded worker-pool to asynchronously process containerd events.

### Key benefits:

prevents subscription loop from blocking

improves throughput and responsiveness

reduces lock contention inside event handlers

handles bursts of start/stop events more efficiently

### Ordering Model

Depending on the project’s expectations, two approaches are possible:

Therefore, two modes are viable:

**Full parallel mode (default)**

Workers freely process events across all containers.

Maximum throughput.

**Per-container ordering (optional)**

Implemented using consistent hashing:
workerID = hash(containerID) % numWorkers

Ensures events for the same container are always processed by the same worker.

Full parallelism is safe unless ordering requirements change later.

### Implementation Plan

Add a buffered job queue (e.g., size 100)

Spawn N worker goroutines (e.g., 5–10 workers)

Subscription loop only enqueues events

Workers call handleContainerdEvent → UpdateContainerdContainer

Add metrics: queue length, processed/sec, worker busy count

Add helper to dedupe PID/NS retrieval logic

Reduce lock hold-times by computing states outside locks

Graceful shutdown by closing queue on StopChan

### Request

Please review the proposed enhancement approach.
Once approved, I will start implementation and open a PR.

### Maintainer Visibility

Tagging maintainers for input:
kubearmor-maintainers-- @aryansharma , @rksharma95

keshav78-78:
**Purpose of PR?**:

Refactor the containerd handler to process containerd events via a bounded worker-pool and introduce a small PID/namespace cache to improve event throughput and reduce latency under bursty container workloads, without changing existing behavior.

Fixes #2289 

No.  
The external behavior of containerd handling, container maps, endpoints, and policy enforcement remains unchanged.  
Only the internal execution model for containerd events is changed from synchronous to worker-pool based processing.

- **Non-k8s mode with containerd:**
  - Started KubeArmor with:
    - `-k8s=false`
    - `-logPath=stdout`
    - `CRI_SOCKET=unix:///run/k3s/containerd/containerd.sock` (via env)
  - Verified log line:
    - `Started to monitor Containerd events (worker-pool mode)`

- **Container lifecycle via containerd (`ctr`):**
  - Pulled test image:
    - `sudo ctr --address /run/k3s/containerd/containerd.sock -n k8s.io images pull docker.io/library/alpine:latest`
  - Ran short-lived containers:
    - `sudo ctr --address /run/k3s/containerd/containerd.sock -n k8s.io run docker.io/library/alpine:latest ct-test-1 sleep 3`
    - Repeated runs to generate bursts of `/tasks/start`, `/tasks/exit`, `/containers/delete`.
  - Observed in KubeArmor logs:
    - Worker handling:
      - `[containerd-worker] handling event: topic=/tasks/start`
      - `[containerd-worker] handling event: topic=/tasks/exit`
      - `[containerd-worker] handling event: topic=/containers/delete`
    - Container lifecycle:
      - `Detected a container (added/...)`
      - `Detected a container (removed/...)`
    - Queue/worker metrics:
      - `containerd events: queued=<N> processed=<M> busy=<K>`

- **PID / Namespace correctness:**
  - Confirmed that for started containers:
    - `Pid`, `PidNS`, `MntNS` are populated correctly via:
      - cached helper (`getPrimaryPidAndNSCached`) when `eventpid == 0`
      - `/proc/<pid>/ns/{pid,mnt}` lookup when `eventpid != 0`
  - Verified that these values still match the previous behavior (PID/NS enumeration via containerd and `/proc`).

- **Stability & error handling:**
  - Verified that:
    - KubeArmor continues running while processing containerd events in workers.
    - No panics observed due to per-worker `recover` around `processContainerdJob`.
    - On exit events, `UpdateContainerdContainer(..., "destroy")` is only invoked when the exiting PID matches the tracked container PID (same logic as before).

- This change is primarily a **performance / scalability improvement** for containerd event handling:
  - Moves event processing from a single synchronous loop to an **8-worker bounded pool** with a **200-size queue**.
  - Introduces a short-lived PID/NS cache (5s) to avoid repeated `/proc` namespace lookups when events arrive in bursts.
  - Keeps all existing data structures and external behavior intact.
- Concurrency model:
  - Full parallel mode is used (no strict per-container ordering is enforced).
  - This matches current expectations where strict ordering between containerd events for the same container is not required.
- Debug / observability:
  - Additional `[containerd-worker] ...` logs and periodic queue metrics were added to make it easier to profile and monitor behavior in real environments.
  - I already reduced/remove it in this final pr

---

**Checklist:**
- [ ] Commit has unit tests


keshav78-78:
Review Please- @Aryan-sharma11 , @rksharma95, @AryanBakliwal

rksharma95:
Hi @keshav78-78 thanks for your PR,

I have few things to discuss before adopting this worker-pool approach:

first, the only point where no. of events could be large is during startup when existing containers are synced and once it's done Imo we don't expect event burst in that case existing approach looks more simpler without any performance bottleneck.

can we measure the performance impact in both the scenarios?

- throughput with and without worker pool.
- cpu usage in both scenarios

keshav78-78:
Hi @rksharma95 ,
Sorry for the delayed update — I got caught up with an urgent college assignment.
I’ve now completed detailed and controlled testing for the 100-replica burst scenario.

Below are the consolidated results from the experiments:

========================================================
                WORKER-POOL IMPLEMENTATION
========================================================
Test Scenario:
- Namespace: test-burst
- Deployment: event-load
- Scaling Pattern: 1 → 100 replicas
- Measurement: Time from scale-out trigger → all pods Available

Observed Behavior:
- Worker pool processed pod update events in parallel.
- No long event queue buildup; goroutines stayed fully utilized.
- System remained stable throughout the burst.

Final Result:
• WORKER-POOL burst duration (100 replicas): **189 seconds**


========================================================
                BASELINE IMPLEMENTATION
========================================================
Test Scenario:
- Same namespace, same deployment, same measurement method.

Observed Behavior:
- Single-threaded pod event handling created significant sequential lag.
- Event queue buildup observed during high load.
- Higher overall latency during the 100-pod initialization burst.

Final Result:
• BASELINE burst duration (100 replicas): **295 seconds**


========================================================
                    SUMMARY & INSIGHTS
========================================================
- Worker-pool reduces processing time by ~36% for 100 replicas.
- Parallel event handling mitigates bottlenecks during bursts.
- The difference becomes more pronounced as pod scaling increases.

========================================================
                ATTACHED EVIDENCE / SCREENSHOTS
========================================================
(Please find the attached screenshots below showing:)
- Worker-pool burst logs
- Baseline burst logs
- Measured timestamps for both tests
- kubectl scale + wait outputs for both 100-replica runs
- Pod event processing traces


Screenshot 2025-12-09 1412200
Screenshot 2025-12-09 141500
Screenshot 2025-12-09 153024
Screenshot 2025-12-09 155742

keshav78-78:
@rksharma95 
Based on the benchmarks, I believe **it’s reasonable to replace the current implementation with the worker-pool**.  
We’re getting a clear ~36% improvement in the one scenario where event pressure actually matters (100-replica burst), and there were no regressions or instability in normal workloads.

That said, **you have far more experience with KubeArmor’s internals**, so I may be wrong in assuming the worker-pool is the better long-term direction.  
If you feel the existing approach is still preferable from a maintenance or design standpoint, I’m completely open to aligning with that.

Just let me know — I can also make it flag-driven or simplify the pool structure if that fits better with the overall architecture.


Note: The `-189s` appeared because I forgot to use the end marker. 
Actual time = ~189 seconds, only the sign was incorrect.

rksharma95:
@keshav78-78 let's make no of workers configurable using a flag.

keshav78-78:
> @keshav78-78 let's make no of workers configurable using a flag.

Thanks @rksharma95, that makes sense 👍

II’ll make the number of workers configurable using a flag and keep a reasonable default so existing behavior stays the same.
This will allow users to tune it based on their workload if needed.

I’ll send an update soon with:
- a worker count flag
- a documented default value
- no change unless the flag is set

Let me know if you prefer a specific flag name or default value.

keshav78-78:
Hi @rksharma95 
I’ve made the containerd worker-pool size configurable as suggested by you 
via a new
--containerd-worker-pool-size flag.

Details:
- Default value preserves existing behavior
- Flag is wired through config (viper) and validated
- Worker pool now initializes using the configured size
- Invalid values safely fall back to the default

Tested locally in non-Kubernetes mode and verified:
- default worker count
- custom worker count via flag
- safe fallback on invalid input

Screenshot 2025-12-16 211900

PR #2298, opened 2025-12-07
 Bug(fix):-Prevent nil pointer crash during early SystemMonitor initialization

### LINKED ISSUE #2297: Bug: Nil Pointer Panic During KubeArmor Initialization
### **General Information**

- **Environment:** vagrant-dev-env (also reproducible outside vagrant)
- **Kernel:** output of `uname -a`
- **Kubernetes Mode:** Reproducible in both `k8s=false` and `k8s=true`
- **Component Affected:** `core/kubeArmor.go :: KubeArmor()`
- **Context:** Occurs during context-based shutdown testing as well as normal startup
- **BPF Enabled:** Not required (bug occurs even with `enableNRI=false` and failed BPF load)

---

### **Description**

A **nil pointer dereference panic** occurs during initialization when `SystemMonitor` and `Logger` are accessed before they are initialized.

The crash happens consistently at startup:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10]

github.com/kubearmor/KubeArmor/KubeArmor/core.KubeArmor
    /home/vagrant/KubeArmor/KubeArmor/core/kubeArmor.go:670
main.main
    /home/vagrant/KubeArmor/KubeArmor/main.go:63
```

At the time of crash:

```
dm.SystemMonitor == nil
dm.SystemMonitor.Logger == nil
```

But the code still executes:

```go
dm.SystemMonitor.Logger.ContainerNsKey = make(map[string]common.OuterKey)
```

This dereference causes the panic.

---

#### **Case 1 — Non-K8s Mode**
```bash
make build
sudo ./kubearmor --k8s=false --visibility=process,file,network,capabilities
```

#### **Case 2 — K8s Mode**
```bash
sudo ./kubearmor --k8s=true
```

#### **Case 3 — During context-based shutdown testing**
Triggered even before reaching the cancellation logic.

---

### **Actual Behavior**

KubeArmor terminates with a segmentation fault before completing initialization.

---

### **Expected Behavior**

- Initialization should not access `SystemMonitor` until it is fully created  
- Startup should succeed without panic, regardless of:  
  - K8s or non-K8s mode  
  - BPF loaded or not  
  - Context cancellation logic  
- Nil checks or initialization ordering should prevent this crash

---

### **Root Cause**

The line:

```go
dm.SystemMonitor.Logger.ContainerNsKey = make(map[string]common.OuterKey)
```

runs **before** `InitSystemMonitor()` creates both `SystemMonitor` and `Logger`.

Thus both are `nil`, leading to a nil pointer dereference.

---

### **Proposed Fix**

Move the initialization after `InitSystemMonitor()`:

```diff
- dm.SystemMonitor.Logger.ContainerNsKey = make(map[string]common.OuterKey)
+ // after SystemMonitor + Logger are initialized
+ dm.SystemMonitor.Logger.ContainerNsKey = make(map[string]common.OuterKey)
```

Or guard with:

```go
if dm.SystemMonitor != nil && dm.SystemMonitor.Logger != nil {
    dm.SystemMonitor.Logger.ContainerNsKey = make(map[string]common.OuterKey)
}
```

---

### **Additional Information for Reviewer**

- This crash occurs even on a clean environment  
- This is not related to the containerd workerpool PR  
- Issue exists in current main branch  
- Reproducible even when BPF load fails (BPF not required)  
- Bug blocks local development and testing  

---

### **Checklist**

- [x] Bug is reproducible  
- [x] Verified on clean build  
- [x] Stacktrace included  
- [x] Root cause identified  
- [x] Proposed fix provided  
- [x] PR linked

keshav78-78:
## **Purpose of PR**
Fixes #2297
Screenshot 2025-12-05 092547
Screenshot 2025-12-07 170814

This PR addresses a startup-time **nil pointer dereference** in `core/kubeArmor.go`.  
The crash happens because `dm.SystemMonitor.Logger.ContainerNsKey` is initialized **before**  
`SystemMonitor` and its `Logger` are created.

This caused consistent startup panics in:

- Non-K8s mode
- K8s mode
- Early context-based shutdown tests
- Failed BPF-loader scenarios
- Policy/HostPolicy disabled flows

### **Fix Implemented**
Initialization is now performed **after** `InitSystemMonitor()` succeeds, and additionally guarded by:

```go
if dm.SystemMonitor != nil && dm.SystemMonitor.Logger != nil {
    dm.SystemMonitor.Logger.ContainerNsKey = make(map[string]common.OuterKey)
}
```

This ensures safe access regardless of mode or degraded state.

---

### **Startup Stability**
- Proper initialization ordering between **Logger → SystemMonitor → Enforcer**.
- No more startup-time nil pointer crash.
- Daemon cleanly reaches shutdown logic in early-cancel test scenarios.
- K8s mode and non-K8s mode behave consistently.

### **Error Resilience**
- No crash when:
  - BPF loading fails  
  - NRI is disabled  
  - `Policy=false` or `HostPolicy=false`  
  - SystemMonitor is intentionally not created  
- Clean fallback behavior without affecting other subsystems.

### **Memory Overhead**
- Minimal — only a single map allocation when needed.
- No extra goroutines.
- No unbounded growth or high memory usage.

---

## **Additional information for reviewer?**
- This is a **pure bug fix**, not tied to the containerd/NRI worker-pool improvements.
- Reproducible in clean environments with no configuration.
- Bug was blocking local development and CI validations.
- No behavioral changes. No API/interface changes. Fully backward-compatible.

---

## **Checklist:**
- [x] Bug fix. Fixes #<issue-number>
- [ ] Breaking change (existing functionality impacted)
- [ ] Documentation update required
- [x] PR Title follows `<type>(<scope>): <subject>`
- [ ] Commit has unit tests


KubeArmor/core/kubeArmor.go
```diff
@@ -635,6 +635,12 @@ func KubeArmor() {
 		}
 		dm.Logger.Print("Initialized KubeArmor Monitor")
 
+		if dm.SystemMonitor != nil && dm.SystemMonitor.Logger != nil {
+			dm.SystemMonitor.Logger.ContainerNsKey = make(map[string]common.OuterKey)
```
rksharma95, line 639: the reported scenario `(SystemMonitor == nil)` could arise when both policy and hostpolicy are disabled (by default policy is enabled), in that case systemMonitor will not be initialized and will result in panic. 

this fix looks good although we can safely remove `if dm.SystemMonitor != nil && dm.SystemMonitor.Logger != nil` as we're early returning if any of these case is true.

rksharma95 (commented):
@keshav78-78 please rebase the PR. LGTM, just one comment inline.

KubeArmor/core/kubeArmor.go
keshav78-78: Thanks for the update — doing it in a moment. Will push the changes shortly.

PR #2318, opened 2025-12-15
Fix config watcher untracked ns

### LINKED ISSUE #2315: bug: Make untrackedNs flag configurable via kubearmor-config
## Bug Report
Updating untrackedNs currently has no effect, and even restarting the DaemonSet does not apply the changes.

**To Reproduce**

1. Deploy kubearmor and update untrackedNs
2. Notice that the change is not reflected in logs 

**Expected behavior**

The changes should be reflected in kubearmor logs

**Screenshots**

keshav78-78:
**Purpose of PR?**:

This PR fixes an issue where updates to the `untrackedNs` field in the
KubeArmor ConfigMap were not reflected at runtime.

Previously, `untrackedNs` was only read during startup. This change updates
the ConfigMap watcher to detect runtime changes to `untrackedNs` and refresh
`SystemMonitor.UntrackedNamespaces` dynamically, ensuring visibility and
monitoring behavior is updated without requiring a KubeArmor restart.

Fixes #2315 

No. This change is backward-compatible and only affects runtime behavior
when the ConfigMap value is updated.

- Updated `untrackedNs` in `kubearmor-config` while KubeArmor was running
- Verified that changes are picked up at runtime without restarting the pod
- Confirmed that visibility logic respects the updated untracked namespace list

_This PR is a focused bug fix and is not part of a larger design change._

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/KubeArmor/core/kubeUpdate.go
+++ b/KubeArmor/core/kubeUpdate.go
@@ -2929,6 +2929,11 @@ func (dm *KubeArmorDaemon) WatchConfigMap() cache.InformerSynced {
 				dm.NodeLock.Lock()
 				dm.Node.ClusterName = cm.Data[cfg.ConfigCluster]
 				dm.NodeLock.Unlock()
+				if v, ok := cm.Data[cfg.ConfigUntrackedNs]; ok {
+					dm.SystemMonitor.UpdateUntrackedNamespaces(v)
+					dm.Logger.Printf("initial untracked namespaces set: %s", v)
+
+				}
 				if _, ok := cm.Data[cfg.ConfigDefaultPostureLogs]; ok {
 					cfg.GlobalCfg.DefaultPostureLogs = (cm.Data[cfg.ConfigDefaultPostureLogs] == "true")
 				}
@@ -2982,10 +2987,11 @@ func (dm *KubeArmorDaemon) WatchConfigMap() cache.InformerSynced {
 				dm.updateVisibilityWithCM(cm, addEvent)
 			}
 		},
-		UpdateFunc: func(_, new any) {
-			if cm, ok := new.(*corev1.ConfigMap); ok && cm.Namespace == cmNS {
+		UpdateFunc: func(oldObj, newObj any) {
+			if cm, ok := newObj.(*corev1.ConfigMap); ok && cm.Namespace == cmNS {
 				cfg.GlobalCfg.HostVisibility = cm.Data[cfg.ConfigHostVisibility]
 				cfg.GlobalCfg.Visibility = cm.Data[cfg.ConfigVisibility]
+
 				cfg.GlobalCfg.Cluster = cm.Data[cfg.ConfigCluster]
 				cfg.GlobalCfg.DropResourceFromProcessLogs = (cm.Data[cfg.ConfigDropResourceFromProcessLogs] == "true")
 				dm.Node.ClusterName = cm.Data[cfg.ConfigCluster]
@@ -3009,6 +3015,23 @@ func (dm *KubeArmorDaemon) WatchConfigMap() cache.InformerSynced {
 				dm.updatEndpointsWithCM(cm, updateEvent)
 				// update visibility for namespaces
 				dm.updateVisibilityWithCM(cm, updateEvent)
+
+				oldCM, okOld := oldObj.(*corev1.ConfigMap)
+				newCM, okNew := newObj.(*corev1.ConfigMap)
+				if !okOld || !okNew {
+					return
+				}
+
+				oldNs := oldCM.Data[cfg.ConfigUntrackedNs]
+				newNs := newCM.Data[cfg.ConfigUntrackedNs]
+
+				if oldNs != newNs {
+					dm.Logger.Printf("untracked namespaces updated: %s -> %s", oldNs, newNs)
+					dm.SystemMonitor.UpdateUntrackedNamespaces(newNs)
+
+					// re-apply visibility using updated untracked namespace list
+					dm.updateVisibilityWithCM(newCM, updateEvent)
+				}
 
 				if _, ok := cm.Data[cfg.ConfigAlertThrottling]; ok {
 					cfg.GlobalCfg.AlertThrottling = (cm.Data[cfg.ConfigAlertThrottling] == "true")
--- a/KubeArmor/core/kubeUpdate.go
+++ b/KubeArmor/core/kubeUpdate.go
@@ -3013,25 +3013,14 @@ func (dm *KubeArmorDaemon) WatchConfigMap() cache.InformerSynced {
 
 				// update default posture for endpoints
 				dm.updatEndpointsWithCM(cm, updateEvent)
-				// update visibility for namespaces
+
+				// forward untracked namespaces to SystemMonitor
+				dm.SystemMonitor.UpdateUntrackedNamespaces(
+					cm.Data[cfg.ConfigUntrackedNs],
+				)
+
+				// visibility updates are already handled here
 				dm.updateVisibilityWithCM(cm, updateEvent)
-
-				oldCM, okOld := oldObj.(*corev1.ConfigMap)
-				newCM, okNew := newObj.(*corev1.ConfigMap)
-				if !okOld || !okNew {
-					return
-				}
-
-				oldNs := oldCM.Data[cfg.ConfigUntrackedNs]
-				newNs := newCM.Data[cfg.ConfigUntrackedNs]
-
-				if oldNs != newNs {
-					dm.Logger.Printf("untracked namespaces updated: %s -> %s", oldNs, newNs)
-					dm.SystemMonitor.UpdateUntrackedNamespaces(newNs)
-
-					// re-apply visibility using updated untracked namespace list
-					dm.updateVisibilityWithCM(newCM, updateEvent)
-				}
 
 				if _, ok := cm.Data[cfg.ConfigAlertThrottling]; ok {
 					cfg.GlobalCfg.AlertThrottling = (cm.Data[cfg.ConfigAlertThrottling] == "true")
--- a/KubeArmor/core/kubeUpdate.go
+++ b/KubeArmor/core/kubeUpdate.go
@@ -2769,10 +2769,16 @@ func (dm *KubeArmorDaemon) updateVisibilityWithCM(cm *corev1.ConfigMap, _ string
 		return
 	}
 
-	// for each namespace if needed change the visibility
 	for _, ns := range nsList.Items {
-		// if namespace is annotated with visibility annotation don't update on config map change
-		if _, found := ns.Annotations[visibilityKey]; found || kl.ContainsElement(dm.SystemMonitor.UntrackedNamespaces, ns.Name) {
+
+		// 1. If namespace is untracked → explicitly remove visibility
+		if kl.ContainsElement(dm.SystemMonitor.UntrackedNamespaces, ns.Name) {
+			dm.UpdateVisibility(deleteEvent, ns.Name, tp.Visibility{})
+			continue
+		}
+
+		// 2. If namespace has visibility annotation → skip CM-based updates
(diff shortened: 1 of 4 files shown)


keshav78-78:
Review Request -@Aryan-sharma11

KubeArmor/core/kubeUpdate.go
```diff
@@ -3011,6 +3017,23 @@ func (dm *KubeArmorDaemon) WatchConfigMap() cache.InformerSynced {
...
+				oldNs := oldCM.Data[cfg.ConfigUntrackedNs]
+				newNs := newCM.Data[cfg.ConfigUntrackedNs]
+
+				if oldNs != newNs {
+					dm.Logger.Printf("untracked namespaces updated: %s -> %s", oldNs, newNs)
+					dm.SystemMonitor.UpdateUntrackedNamespaces(newNs)
+
+					// re-apply visibility using updated untracked namespace list
+					dm.updateVisibilityWithCM(newCM, updateEvent)
+				}
```
Aryan-sharma11: This would be a simpler implementation wdyt.
```suggestion
				cfg.Globalcfg.ConfigUntrackedNs  = strings.Split(cm.Data[cfg.ConfigUntrackedNs], ",")				
				dm.Logger.Printf("untracked namespaces : %s", cfg.Globalcfg.ConfigUntrackedNs )						
```

keshav78-78: That sounds reasonable
Using the current runtime state instead of comparing old/new ConfigMap objects does simplify this.

I was thinking of updating the UpdateFunc to rely on
`SystemMonitor.UntrackedNamespaces` as the old value and compare it with
the new ConfigMap value, something like this:

```go
newNs := cm.Data[cfg.ConfigUntrackedNs]
oldNs := strings.Join(dm.SystemMonitor.UntrackedNamespaces, ",")

if newNs != oldNs {
    dm.SystemMonitor.UpdateUntrackedNamespaces(newNs)
    dm.updateVisibilityWithCM(cm, updateEvent)
}

Comment (Aryan-sharma11) [reply KubeArmor/core/kubeUpdate.go]: Still, in this flow, we first convert oldNs from an array to a string and then convert newNs back into an array. This introduces a redundant conversion step, especially since the namespaces may be the same but in a different order. Moreover, we already handle visibility updates for every change with `dm.updateVisibilityWithCM`, so this extra conversion isn’t necessary.

Comment (keshav78-78) [reply KubeArmor/core/kubeUpdate.go]: Got it — thanks for the detailed explanation

You're right: converting the runtime slice to a string and then parsing it
back introduces redundant conversions and can also lead to false updates
when the namespace order differs.

I’ll simplify the UpdateFunc to always forward the new ConfigMap value
directly to `UpdateUntrackedNamespaces` and move the normalization and
comparison logic inside `SystemMonitor`, keeping the watcher logic minimal.

Something along these lines:

```go
// UpdateFunc
newNs := cm.Data[cfg.ConfigUntrackedNs]
dm.SystemMonitor.UpdateUntrackedNamespaces(newNs)

// visibility updates are already handled via updateVisibilityWithCM

Comment (keshav78-78) [reply KubeArmor/core/kubeUpdate.go]: please Review @Aryan-sharma11

File: KubeArmor/monitor/systemMonitor.go [hunk 2]
```diff
@@ -329,6 +331,30 @@ func (mon *SystemMonitor) UpdateThrottlingConfig() {
...
+	// Normalize order to avoid false updates
+	sort.Strings(namespaces)
+
+	// No-op if nothing changed
+	if slices.Equal(sm.UntrackedNamespaces, namespaces) {
+		return
+	}
+
+	sm.UntrackedNamespaces = namespaces
+}
```
Aryan-sharma11: Hi @keshav78-78, my reasoning for not comparing against the existing `untrackedNs` is that we already update visibilities on every ConfigMap change. Instead of performing the comparison, we can directly assign the new values. Then, within the `updateVisibilityWithCM` function, we can explicitly remove visibility configurations from the BPF maps for the namespaces listed in `untrackedNs`. Currently, we don’t add any visibility for `untrackedNs`, but there could be scenarios where visibility already exists and needs to be cleaned up.
Feel free to ask if you have any questions or need clarification.

keshav78-78: Thanks for the explanation.

Just to confirm my understanding:
you’re saying we don’t need to compare untrackedNs in the watcher.
Instead, on every ConfigMap update, we can rely on updateVisibilityWithCM
to handle visibility correctly, including removing any existing visibility
for namespaces that are now untracked.

If this is correct, I’ll keep the UpdateFunc simple and move all the
cleanup logic there.

keshav78-78: @Aryan-sharma11 
I checked the code again and confirmed that there is no old/new comparison
of untrackedNs in the watcher.

The watcher stays simple and just passes the latest ConfigMap values.
I added explicit cleanup of existing visibility from the BPF maps for
namespaces that become untracked inside updateVisibilityWithCM, without
adding any diff-based logic.

This keeps all visibility handling in one place and does not change the
existing annotation-based behavior.

KubeArmor/monitor/systemMonitor.go
```diff
@@ -329,6 +331,30 @@ func (mon *SystemMonitor) UpdateThrottlingConfig() {
...
+	namespaces := make([]string, 0, len(parts))
+	for _, ns := range parts {
+		ns = strings.TrimSpace(ns)
+		if ns != "" {
+			namespaces = append(namespaces, ns)
+		}
+	}
+
+	// Normalize order to avoid false updates
+	sort.Strings(namespaces)
```
Aryan-sharma11: no need to sort here. we are using `kl.ContainsElement(dm.SystemMonitor.UntrackedNamespaces, ns.Name)`.

KubeArmor/monitor/systemMonitor.go
```diff
@@ -329,6 +331,30 @@ func (mon *SystemMonitor) UpdateThrottlingConfig() {
...
+
+	// No-op if nothing changed
+	if slices.Equal(sm.UntrackedNamespaces, namespaces) {
```
Aryan-sharma11: This is not required, we can directly assign the namespaces.

KubeArmor/monitor/systemMonitor.go
```diff
@@ -329,6 +331,30 @@ func (mon *SystemMonitor) UpdateThrottlingConfig() {
...
+func (sm *SystemMonitor) UpdateUntrackedNamespaces(v string) {
+	parts := strings.Split(v, ",")
+
+	namespaces := make([]string, 0, len(parts))
+	for _, ns := range parts {
+		ns = strings.TrimSpace(ns)
+		if ns != "" {
+			namespaces = append(namespaces, ns)
+		}
+	}
```
Aryan-sharma11:
```suggestion
for _, ns := range parts {
	if ns = strings.TrimSpace(ns); ns != "" {
		namespaces = append(namespaces, ns)
	}
}
```

KubeArmor/monitor/systemMonitor.go
```diff
@@ -329,6 +331,30 @@ func (mon *SystemMonitor) UpdateThrottlingConfig() {
...
+
+	// Normalize order to avoid false updates
+	sort.Strings(namespaces)
+
+	// No-op if nothing changed
+	if slices.Equal(sm.UntrackedNamespaces, namespaces) {
+		return
+	}
+
```
Aryan-sharma11:
```suggestion
cfg.Globalcfg.untrackedNamespaces = namespaces
```

Aryan-sharma11 (changes requested):

keshav78-78:
Just doing in a moment

keshav78-78:
@Aryan-sharma11  I’ve updated the code based on your feedback. The watcher remains simple and always forwards the latest ConfigMap value, and the visibility cleanup for untracked namespaces is now handled in the visibility update path.

Let me know if you’d like any further adjustments.

KubeArmor/monitor/systemMonitor.go
```diff
@@ -329,6 +329,21 @@ func (mon *SystemMonitor) UpdateThrottlingConfig() {
...
+func (sm *SystemMonitor) UpdateUntrackedNamespaces(v string) {
+	parts := strings.Split(v, ",")
+
+	namespaces := make([]string, 0, len(parts))
+	for _, ns := range parts {
+		if ns = strings.TrimSpace(ns); ns != "" {
+			namespaces = append(namespaces, ns)
+		}
+	}
+
```
Aryan-sharma11: let's add a log here 
`dm.Logger.Printf("initial untracked namespaces set: %v", namespaces)`

Aryan-sharma11: need to update `cfg.Globalcfg.untrackedns` as well
```suggestion
cfg.Globalcfg.untrackedns = namespaces
```

KubeArmor/core/kubeUpdate.go
```diff
@@ -3008,7 +3020,13 @@ func (dm *KubeArmorDaemon) WatchConfigMap() cache.InformerSynced {
 
 				// update default posture for endpoints
 				dm.updatEndpointsWithCM(cm, updateEvent)
-				// update visibility for namespaces
+
+				// forward untracked namespaces to SystemMonitor
+				dm.SystemMonitor.UpdateUntrackedNamespaces(
```
Aryan-sharma11: This update should be applied only when cm.Data[cfg.ConfigUntrackedNs] exists in the ConfigMap. Otherwise, the default values could be unintentionally reset to empty.

KubeArmor/core/kubeUpdate.go
```diff
@@ -2931,6 +2937,11 @@ func (dm *KubeArmorDaemon) WatchConfigMap() cache.InformerSynced {
 				dm.NodeLock.Lock()
 				dm.Node.ClusterName = cm.Data[cfg.ConfigCluster]
 				dm.NodeLock.Unlock()
+				if v, ok := cm.Data[cfg.ConfigUntrackedNs]; ok {
+					dm.SystemMonitor.UpdateUntrackedNamespaces(v)
+					dm.Logger.Printf("initial untracked namespaces set: %s", v)
```
Aryan-sharma11: let's handle the logging in `UpdateUntrackedNamespaces` function

keshav78-78: I’ll update UpdateUntrackedNamespaces to handle the logging internally and also keep cfg.GlobalCfg.untrackedNs in sync there.

In the watcher, I’ll guard the call so it’s applied only when cm.Data[cfg.ConfigUntrackedNs] exists, to avoid unintentionally resetting the value to empty when the key is missing.

Please let me know if this approach looks good before I push the update.

Aryan-sharma11: Yes this approach makes sense. 
Can you also upload the screenshots after testing it locally?

keshav78-78: Yeah, of course I’ll upload the screenshots after testing it locally

keshav78-78: I’ve pushed the latest changes on this PR.
This includes the mutex fix in systemMonitor and runtime syncing of untrackedNs from the ConfigMap watcher and updated the feeder test also.

AryanBakliwal (commented):
@Aryan-sharma11 I tried to test this, 

1. Policies are not enforced when ns is untracked, as expected.
2. But after a ns is added to untracked ns list, policies previously applied to it should be removed and not enforced anymore?
3. Policies (KSPs) already applied to an untracked ns can be deleted, need to add

Aryan-sharma11 (approved):

AryanBakliwal (approved):

Aryan-sharma11:
Thanks a lot for the PR @keshav78-78 :rocket:

PR #2376, opened 2026-01-15
fix(log): make DEBUG env var case-insensitive and add unit tests

As1agi:
Currently, setting `DEBUG=True` (or `TRUE`) is ignored because the environment variable check is case-sensitive. This commit uses `strings.ToLower()` to handle case sensitivity and adds unit tests to verify logger configuration

**Purpose of PR?**:

Fixes case-sensitive environment variable handling for the logger. Previously, users attempting to enable debug mode with `DEBUG=True` (common in other ecosystems) would fail to see debug logs. This PR ensures robust handling of the flag and improves test coverage by adding `logger_test.go`

No

- Ran `go test -v ./KubeArmor/log/` (All tests passed)
- Verified defaults (variable unset) remain in INFO mode

This PR adds the first unit tests for the `log` package to prevent future regressions

**Checklist:**
- [] Bug fix. Fixes #
- [x] Commit has unit tests

### DIFF
--- a/KubeArmor/config/config.go
+++ b/KubeArmor/config/config.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"sync/atomic"
 
 	kg "github.com/kubearmor/KubeArmor/KubeArmor/log"
 	"github.com/spf13/viper"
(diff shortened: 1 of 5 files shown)


AryanBakliwal:
Hey @As1agi, your  effort in adding tests and helping improve coverage is appreciated.
That said, this test only checks the logger level toggle using the `DEBUG` env var, which is more of a configuration thing and not very meaningful as a unit test on its own. If you could add tests for other important,  currently untested functionality as well where actual behavior/logic is used, that would be much more useful.

Let us know if you need any help with that.

As1agi:
> Hey @As1agi, your effort in adding tests and helping improve coverage is appreciated. That said, this test only checks the logger level toggle using the `DEBUG` env var, which is more of a configuration thing and not very meaningful as a unit test on its own. If you could add tests for other important, currently untested functionality as well where actual behavior/logic is used, that would be much more useful.
> 
> Let us know if you need any help with that.

Thanks for the feedback @AryanBakliwal , Agreed, the env check is too trivial.

I'll close this PR and open a fresh one that targets actual business logic/validation in the core packages to improve meaningful coverage

Actually, since you offered help , is there a specific package or functional area you consider most critical to cover right now? I'm happy to hunt down the coverage gaps wherever they are most needed

Aryan-sharma11:
Since this PR is no longer required, we’ll be closing it. @As1agi , feel free to work on other issues, including improving test coverage. Thanks!

PR #2447, opened 2026-02-16
refactor(core): use raw machine ID as node ID

### LINKED ISSUE #2444: use machine ID as node id
## Feature Request

**Short Description**

use machine ID as node id directly, currently the node id is encoded string (node name+ machine id).  

https://github.com/kubearmor/KubeArmor/blob/dbc21da1bea9faae60cdc76965000d99f96eaed4/KubeArmor/core/kubeArmor.go#L560

- [ ] use machine id as node id
- [ ] verify it across most common distros

parthivsaikia:
**Purpose of PR?**:

Removes the protectedID function. The Node ID now uses the actual machine ID (from hostnamectl or /etc/machine-id) instead of hashing it with the node name.

Fixes #2444 

No

Verified locally on Arch Linux (systemd):

    1. Ran KubeArmor in host mode: sudo -E go run main.go -k8s=false

    2. Confirmed via logs that Node ID now matches the raw UUID from /etc/machine-id (e.g., 8d1096...) instead of the previous SHA256 hash.

    3. Verified that the logic correctly falls back to reading /etc/machine-id if hostnamectl data isn't immediately available.

**Checklist:**
- [x] Bug fix. Fixes #24444
- [ ] Commit has unit tests

### DIFF
--- a/KubeArmor/core/kubeArmor.go
+++ b/KubeArmor/core/kubeArmor.go
@@ -6,12 +6,10 @@ package core
 
 import (
 	"context"
-	"crypto/hmac"
-	"crypto/sha256"
-	"encoding/hex"
 	"fmt"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"strings"
 	"sync"
 	"syscall"
@@ -83,7 +81,7 @@ type KubeArmorDaemon struct {
 	HostSecurityPolicies     []tp.HostSecurityPolicy
 	HostSecurityPoliciesLock *sync.RWMutex
 
-	//DefaultPosture (namespace -> postures)
+	// DefaultPosture (namespace -> postures)
 	DefaultPostures     map[string]tp.DefaultPosture
 	DefaultPosturesLock *sync.Mutex
 
@@ -199,7 +197,7 @@ func (dm *KubeArmorDaemon) DestroyKubeArmorDaemon() {
 	}
 
 	if dm.USBDeviceHandler != nil {
-		//close USB device handler
+		// close USB device handler
 		if dm.CloseUSBDeviceHandler() {
 			dm.Logger.Print("Stopped USB Device Handler")
 		}
@@ -212,7 +210,7 @@ func (dm *KubeArmorDaemon) DestroyKubeArmorDaemon() {
 	}
 
 	if dm.StateAgent != nil {
-		//go dm.StateAgent.PushNodeEvent(dm.Node, state.EventDeleted)
+		// go dm.StateAgent.PushNodeEvent(dm.Node, state.EventDeleted)
 		if err := dm.CloseStateAgent(); err != nil {
 			kg.Errf("Failed to destroy StateAgent: %s", err.Error())
 		} else {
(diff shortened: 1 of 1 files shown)


nyrahul:
One problem I see is that there is no handling if the MachineIDPath file does not exist. This problem was present in the previous code as well, but since we are no longer affixing an additional random value, the ID becomes empty.

rksharma95:
@parthivsaikia can we validate it across the [kubearmor supported distributions](https://github.com/kubearmor/KubeArmor/blob/main/getting-started/support_matrix.md#supported-linux-distributions)

parthivsaikia:
@nyrahul  @rksharma95 I have added two files (`cfg.GlobalCfg.MachineIDPath`, `/var/lib/dbus/machine-id`) to check for the machine-id. And if both the files aren't present then dm.Node.NodeName is assigned as the node id.

rksharma95:
> @nyrahul @rksharma95 I have added two files (`cfg.GlobalCfg.MachineIDPath`, `/var/lib/dbus/machine-id`) to check for the machine-id. And if both the files aren't present then dm.Node.NodeName is assigned as the node id.

@parthivsaikia were you able to validate it on the distros I shared in previous comment?

parthivsaikia:
> > @nyrahul @rksharma95 I have added two files (`cfg.GlobalCfg.MachineIDPath`, `/var/lib/dbus/machine-id`) to check for the machine-id. And if both the files aren't present then dm.Node.NodeName is assigned as the node id.
> 
> @parthivsaikia were you able to validate it on the distros I shared in previous comment?

@rksharma95 I haven't spun up full VM for every distro but I have investigated their architectures. 

- Standard Systemd (Amazon Linux 2, Alibaba, Oracle/UEK, Bottlerocket): These all adhere to the systemd spec and are guaranteed to have /etc/machine-id.

- Talos Linux: While not systemd-based, I confirmed their documentation states they explicitly generate /etc/machine-id for compatibility, so the primary path works there as well.

- Fallback: For any non-standard environments (like Alpine), the code now checks /var/lib/dbus/machine-id.

- Safety Net: As a final guarantee, defaulting to NodeName (hostname) ensures we never return an empty ID.

AryanBakliwal (commented):
@parthivsaikia I checked on KubeArmor's supported distros, the machine-id file seems to be present on all. Can you please handle the failing test and squash the commits?

parthivsaikia:
Hey @AryanBakliwal I have pushed the changes to fix the failing test. Can you check it.

KubeArmor/core/kubeArmor.go
```diff
@@ -547,17 +545,26 @@ func KubeArmor() {
...
+					break
+				}
+			}
+		}
+	}
+
+	if dm.Node.NodeID == "" {
+		dm.Node.NodeID = dm.Node.NodeName
 	}
-	dm.Node.NodeID = protectedID(dm.Node.NodeID, dm.Node.NodeName)
```
vishnusomank, line 560:
```suggestion

dm.Node.NodeID = dm.Node.NodeName

for _, path := range machineIDPaths {
		if id, err := os.ReadFile(filepath.Clean(path)); err == nil {
				cleanID := strings.TrimSpace(string(id))
				if cleanID != "" {
					dm.Node.NodeID = cleanID
					break
				}
			}
}
```

vishnusomank (changes requested):
Can't we do this?

KubeArmor/core/kubeArmor.go
parthivsaikia: `dm.Node.NodeID` is already assigned a value at this line 

```
if strings.Contains(line, "Machine ID") {
				dm.Node.NodeID = strings.Split(line, ": ")[1]
}
```
so I think checking if the value is empty is needed to avoid unwanted override.

vishnusomank (approved):

AryanBakliwal (approved):
LGTM!

PR #2487, opened 2026-03-03
feat(operator): Improve CRI socket detection

### LINKED ISSUE #2360: improve CRI socket detection using operator
## Feature Request

in case if multiple CRI socket file being detected on the node, with the current design snitch identify CRI socket on FCFS basis. there should be a improved way to identify the CRI socket that is currently being used. 

i.e. either using kubelet configuration file i.e. `/etc/kubernetes/kubelet.conf` to identify the active socket file with a flag (`--config-dir`) to define path or 
support  `--socket-file` flag to set default socket file location to check.  i.e. `--socket-file /run/k3s/containerd/containerd.sock`

AryanBakliwal:
**Purpose of PR?**:

Fixes #2360

No

**Checklist:**
- [ ] Bug fix.
- [ ] Commit has unit tests


pkg/KubeArmorOperator/runtime/runtime.go
```diff
@@ -25,7 +35,75 @@ func DetectNRI(pathPrefix string) (string, error) {
...
+
+	return true
+}
+
+func runtimeSockHasContainer(containerID, sockPath string) bool {
+	if strings.Contains(sockPath, "containerd") {
+		return containerdSockHasContainer(containerID, sockPath)
+	} else if strings.Contains(sockPath, "docker") {
+		return dockerSockHasContainer(containerID, sockPath)
+	} else if strings.Contains(sockPath, "crio") || strings.Contains(sockPath, "dockershim") {
```
rksharma95: is it validated? dockershim socket exposes crio compatible apis?

here we're using containerd 
https://github.com/kubearmor/KubeArmor/blob/185c35c41c85d9df01772da998d6e2cbc937374c/KubeArmor/common/common.go#L434-L440

AryanBakliwal: I checked into this, dockershim does support CRI gRPC APIs like cri-o but the support is only limited to v1alpha2. The client we are using for cri-o uses v1, so I will make changes to support v1alpha2 as well for dockershim socket

pkg/KubeArmorOperator/runtime/runtime.go
```diff
@@ -25,7 +35,99 @@ func DetectNRI(pathPrefix string) (string, error) {
...
+
+func containerdSockHasContainer(containerID, sockPath string) bool {
+	client, err := containerd.New(sockPath)
+	if err != nil {
+		log.Warnf("Error in creating containerd client: %v", err)
+		return false
+	}
+	defer client.Close()
+
+	ctx := namespaces.WithNamespace(context.Background(), "k8s.io")
```
achrefbensaad: This does not takes into account that the condition where docker is using containerd as runtime the namespace will be `moby` in that case

achrefbensaad: also even if this is handled in a seperate function , we should not hardcode the namespace; move it into a constant then use it here

pkg/KubeArmorOperator/runtime/runtime.go
```diff
@@ -25,7 +35,99 @@ func DetectNRI(pathPrefix string) (string, error) {
...
+func runtimeSockHasContainer(containerID, sockPath string) bool {
+	if strings.Contains(sockPath, "containerd.sock") {
+		return containerdSockHasContainer(containerID, sockPath)
+	} else if strings.Contains(sockPath, "docker.sock") {
+		return dockerSockHasContainer(containerID, sockPath)
+	} else if strings.Contains(sockPath, "crio.sock") {
+		return crioSockHasContainer(containerID, sockPath)
+	} else if strings.Contains(sockPath, "dockershim.sock") {
+		return dockershimSockHasContainer(containerID, sockPath)
+	}
```
achrefbensaad: you should not try to guess the runtime by looking at the socket, as this will fail when customers are using custom runtime wrappers. 
instead, use try logic.

achrefbensaad (changes requested):

pkg/KubeArmorOperator/runtime/runtime.go
AryanBakliwal: updated with checking in both namespaces

pkg/KubeArmorOperator/runtime/runtime.go
AryanBakliwal: updated

rksharma95 (changes requested):
we'll need to handle it with kubearmor (k8s mode) as well, as snitch generate daemonset with nodeselector labels containing both runtime and socket we can make use of it. or either we can have another flag in addition to criSocket and can use that.

https://github.com/kubearmor/KubeArmor/blob/5d622d362d8211d0b3c85c633d6adefc00f56171/KubeArmor/core/kubeArmor.go#L782-L883

pkg/KubeArmorOperator/runtime/runtime.go
```diff
@@ -25,7 +40,136 @@ func DetectNRI(pathPrefix string) (string, error) {
...
+
+	// get container ID of snitch
+	podName := os.Getenv("POD_NAME")
+	namespace := os.Getenv("POD_NAMESPACE")
+	pod, err := cl.CoreV1().Pods(namespace).Get(context.TODO(), podName, metav1.GetOptions{})
+	if err != nil {
+		log.Errorf("Error in getting pod info: %v\n", err)
+		return "NA", "NA", "NA"
+	}
+	id := pod.Status.ContainerStatuses[0].ContainerID // <type>://<container_id>, for example containerd://abcdef123456.....
```
Aryan-sharma11: let's have a check here for `len(pod.Status.ContainerStatuses[0])>0` to avoid index out of bounds panic

pkg/KubeArmorOperator/runtime/runtime.go
```diff
@@ -25,7 +40,136 @@ func DetectNRI(pathPrefix string) (string, error) {
...
+	return true
+}
+
+func dockershimSockHasContainer(containerID, sockPath string) bool {
+	conn, err := grpc.Dial("unix://"+sockPath, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		log.Warnf("Error in creating dockershim client: %v", err)
+		return false
+	}
+
```
Aryan-sharma11, line 94:
```suggestion
defer conn.Close()
```

pkg/KubeArmorOperator/runtime/runtime.go
```diff
@@ -25,7 +40,136 @@ func DetectNRI(pathPrefix string) (string, error) {
...
+	return true
+}
+
+func crioSockHasContainer(containerID, sockPath string) bool {
+	conn, err := grpc.NewClient("unix://"+sockPath, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		log.Warnf("Error in creating crio client: %v", err)
+		return false
+	}
+
```
Aryan-sharma11, line 118:
```suggestion
defer conn.close()
```

pkg/KubeArmorOperator/runtime/runtime.go
```diff
@@ -25,7 +40,136 @@ func DetectNRI(pathPrefix string) (string, error) {
...
+	if err != nil {
+		log.Warn("Container not found in Docker namespace")
+		return false
+	}
+
+	return true
+}
+
+func dockershimSockHasContainer(containerID, sockPath string) bool {
+	conn, err := grpc.Dial("unix://"+sockPath, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
Aryan-sharma11: let's use grpc.Newclient() here as `grpc.Dial` is deprecated

pkg/KubeArmorOperator/runtime/runtime.go
```diff
@@ -39,47 +183,53 @@ func DetectRuntimeViaMap(pathPrefix string, k8sRuntime string, explicitSocket st
...
+
 	if k8sRuntime != "" {
 		for _, path := range common.ContainerRuntimeSocketMap[k8sRuntime] {
 			if _, err := os.Stat(pathPrefix + path); err == nil || os.IsPermission(err) {
-				if (k8sRuntime == "docker" && strings.Contains(path, "containerd")) || k8sRuntime == "containerd" {
-					if nriPath, err := DetectNRI(pathPrefix); err == nil {
-						return k8sRuntime, path, nriPath
-					} else {
-						log.Warnf("%s", err)
+				found, k8sRuntime := runtimeSockHasContainer(containerID, pathPrefix+path)
```
Aryan-sharma11: we have already declared k8sRuntime above 
```suggestion
				found, detectedRuntime := runtimeSockHasContainer(containerID, pathPrefix+path)
```

pkg/KubeArmorOperator/runtime/runtime.go
```diff
@@ -25,7 +40,136 @@ func DetectNRI(pathPrefix string) (string, error) {
...
+		log.Warnf("Error in creating dockershim client: %v", err)
+		return false
+	}
+
+	dockershimClient := criV1alpha2.NewRuntimeServiceClient(conn)
+
+	req := &criV1alpha2.ContainerStatusRequest{
+		ContainerId: containerID,
+		Verbose:     true,
+	}
```
Aryan-sharma11, line 102:
```suggestion
	req := &criV1alpha2.ContainerStatusRequest{
		ContainerId: containerID,
		Verbose:     false
	}
```
`Verbose: true` adds overhead, we are not using the response for the request only checking success or fail.

Aryan-sharma11 (changes requested):

rksharma95 (dismissed):
LGTM 👍

Aryan-sharma11 (dismissed):

achrefbensaad (changes requested):
Will reveiw by EOD

achrefbensaad (dismissed):
LGTM, it handles custom runtimes as well , GJ

rksharma95 (approved):

Aryan-sharma11 (approved):

PR #2499, opened 2026-03-16
Enforce rules using inode+device (handle symlink and original path)

### LINKED ISSUE #2319: kubearmor should enforce on both symlink and original link

asmit27rai:
- Make policy lookup canonical by using the file identity (device + inode) in BPF and userspace, so accesses via symlink or original path match the same rule.
- Preserve existing string-based directory/prefix/name matching to retain current behavior for directory and recursive rules.

Fix : #2319


asmit27rai:
@AryanBakliwal @rksharma95 Tests added.
Please have a look

KubeArmor/enforcer/bpflsm/mapHelpers.go
```diff
@@ -29,8 +29,14 @@ type NsKey struct {
...
-	Source [200]byte
+	KeyType uint8
+	_pad    [7]byte
+	Ino     uint64
+	Dev     uint64
+	SrcIno  uint64
+	SrcDev  uint64
+	Path    [200]byte
+	Source  [200]byte
 }
```
AryanBakliwal, line 42: On kernel version <= 5.10, key size cannot be larger than max stack size that is 512 bytes, which is why the `InnerKey` size is kept to 400 bytes so `ArgumentsKey` and `ArgListKey`'s total size does not exceed 512 bytes.

KubeArmor/enforcer/bpflsm/rulesHandling.go
```diff
@@ -7,12 +7,40 @@ import (
...
 	tp "github.com/kubearmor/KubeArmor/KubeArmor/types"
 )
 
+// makeInnerKey builds an InnerKey from path and optional source path.
+// It populates inode and device fields (following symlinks) and also
+// copies the original path strings for directory and alerting support.
+func makeInnerKey(pathStr, srcStr string) InnerKey {
+	var key InnerKey
+	if pathStr != "" {
+		if st, err := os.Stat(pathStr); err == nil {
```
AryanBakliwal: This will not work, this runs inside the KubeArmor container which does not have access to the host filesystem

AryanBakliwal (changes requested):
The unit tests would still pass because they run directly on the host, not inside the container. To properly validate this, we need integration tests

PR #2523, opened 2026-04-04
fix(enforcer): correct key deletion in RemoveDeletedEntriesForNodeSelector

### LINKED ISSUE #2520: bug: RemoveDeletedEntriesForNodeSelector fails to  remove entries due to iterating values instead of keys
## Bug Report

**General Information**
- Environment: Any Kubernetes cluster with KubeArmor Operator deployed
- Kernel version: N/A (code-level bug, not environment specific)
- Orchestration: Any
- Relevant code: pkg/KubeArmorOperator/internal/controller/cluster.go line 1016
- Target: KubeArmor Operator

**To Reproduce**
1. Deploy KubeArmor operator
2. Set a globalNodeSelector or component-specific nodeSelector in KubeArmorConfig
3. Remove a nodeSelector key from KubeArmorConfig (this marks it as "-" internally)
4. Observe that the removed nodeSelector key persists in the daemonset/deployment node selectors it is never cleaned up

**Expected behavior**
When a nodeSelector entry is marked for deletion (value set to "-"), `RemoveDeletedEntriesForNodeSelector` should remove it from the map. Instead, due to a bug where the map value is collected instead of the key,`delete()` is called with "-" as the key which never exists, so nothing is deleted.

Buggy code (line 1016):
```go
for _, v := range ns {
    if v == "-" {
        deletedEntries = append(deletedEntries, v) // appends value "-", not the key
    }
}
// delete(ns, "-") is called — key "-" doesn't exist, nothing deleted
```

Fix:
```go
for k, v := range ns {
    if v == "-" {
        deletedEntries = append(deletedEntries, k) // correctly appends the key
    }
}
```

This function is called in 4 places inside `UpdateKubeArmorImages()`:
- line 583: KubeArmorNodeSelector
- line 656: KubeArmorRelayNodeSelector  
- line 713: KubeArmorControllerNodeSelector
- line 738: GlobalNodeSelectors

Every config update leaves stale "-" entries in node selector maps which get propagated to daemonsets and deployments incorrectly.

A unit test confirming the bug and validating the fix will be included in the PR.

I'd like to work on this fix if maintainers agree with the approach.

**Screenshots**
N/A — code-level bug. Relevant function:
https://github.com/kubearmor/KubeArmor/blob/main/pkg/KubeArmorOperator/internal/controller/cluster.go#L1016

Athang69:
**Purpose of PR?**:
This PR fixes a logical error in the RemoveDeletedEntriesForNodeSelector function where the code incorrectly attempted to delete map entries using the value (-) as the key. This prevented the intended cleanup of node selector entries.

Fixes #2520

No

1. Single deletion: Verified that passing a map with one entry set to "-" results in that entry being removed.
2. No deletion: Verified that a map with standard key-value pairs remains untouched.
3. Multiple entries: Verified that only the entries marked with "-" are removed while others persist.
4. Unit Tests: Added `TestRemoveDeletedEntriesForNodeSelector` to automate these scenarios.

**Checklist:**
- [x] Bug fix. Fixes #2520
- [x] Commit has unit tests

**Output of the test before editing the function**
Screenshot from 2026-04-04 23-08-00

**Output of the test after editing the function**
Screenshot from 2026-04-04 23-07-23

### DIFF
--- a/pkg/KubeArmorOperator/internal/controller/cluster.go
+++ b/pkg/KubeArmorOperator/internal/controller/cluster.go
@@ -1014,17 +1014,15 @@ func AddOrUpdateEnv(dst *[]corev1.EnvVar, src []corev1.EnvVar) {
 }
 
 func RemoveDeletedEntriesForNodeSelector(ns map[string]string) {
-	deletedEntries := []string{}
-
-	for _, v := range ns {
-		if v == "-" {
-			deletedEntries = append(deletedEntries, v)
-		}
-	}
-
-	for _, k := range deletedEntries {
-		delete(ns, k)
-	}
+    deletedEntries := []string{}
+    for k, v := range ns {        
+        if v == "-" {
+            deletedEntries = append(deletedEntries, k)  
+        }
+    }
+    for _, k := range deletedEntries {
+        delete(ns, k)
+    }
 }
 
 func RemoveDeletedEntriesForEnv(env *[]corev1.EnvVar) {
--- a/pkg/KubeArmorOperator/internal/controller/cluster.go
+++ b/pkg/KubeArmorOperator/internal/controller/cluster.go
@@ -1014,16 +1014,10 @@ func AddOrUpdateEnv(dst *[]corev1.EnvVar, src []corev1.EnvVar) {
 }
 
 func RemoveDeletedEntriesForNodeSelector(ns map[string]string) {
-	deletedEntries := []string{}
-
-	for _, v := range ns {
+	for k, v := range ns {
 		if v == "-" {
-			deletedEntries = append(deletedEntries, v)
+			delete(ns, k)
 		}
-	}
-
-	for _, k := range deletedEntries {
-		delete(ns, k)
 	}
 }
 
--- a/pkg/KubeArmorOperator/internal/controller/cluster_test.go
+++ b/pkg/KubeArmorOperator/internal/controller/cluster_test.go
@@ -5,7 +5,7 @@ package controller
 
 import (
 	"testing"
-
+	"reflect"
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
 )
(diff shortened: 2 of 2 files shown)


Athang69:
@Aryan-sharma11 @AryanBakliwal tagging you for a review on this PR. Could you let me know if this approach aligns with the project's standards, or if there are specific changes you'd like to see? Happy to iterate!

Athang69:
Hi @Aryan-sharma11 @AryanBakliwal , gentle ping on this PR would appreciate your review when possible
Thanks!

pkg/KubeArmorOperator/internal/controller/cluster.go
```diff
@@ -1014,17 +1014,11 @@ func AddOrUpdateEnv(dst *[]corev1.EnvVar, src []corev1.EnvVar) {
 }
 
 func RemoveDeletedEntriesForNodeSelector(ns map[string]string) {
-	deletedEntries := []string{}
-
-	for _, v := range ns {
+	for k, v := range ns {
```
tuxerrante, line 1017:
```go
maps.DeleteFunc(ns, func(k string, v string) bool {
		return ns[k] == "-"
	})

```

tuxerrante (approved):
Hi @Athang69, @achrefbensaad 
I independently reproduced #2520 against the current `main` branch and confirmed that this PR addresses the root cause.

Additional validation I ran:

- Applied the new regression test to the parent implementation: it fails because the map retains the deleted entry as `"key":"-"`.
- Ran the test with the fix: it passes, including empty-map and nil-map cases.
- Ran the Operator `fmt`, `vet`, and `build` Make targets successfully.
- Built the Operator image locally and exercised a live `KubeArmorConfig` node-selector add → remove → re-add lifecycle on Kubernetes v1.35.1 with containerd.
- After removal, the managed Deployment retained its unrelated OS selector, remained Ready with zero restarts, and showed no repeated update churn during a 90-second observation window.

I also have an equivalent implementation using `maps.DeleteFunc` and the extra empty/nil test cases, but I do not want to duplicate or supersede your contribution. Would you or the maintainers like those test cases or the sanitized runtime evidence added here?

Possible follow-ups, kept separate from this focused fix, are multi-node cloud validation, documenting the exact-image prerequisite for `imagePullPolicy: Never`, and profiling high-cardinality selector updates/node churn in a dedicated test cluster.

PR #2527, opened 2026-04-05
fix(core): use TaskExit type to unmarshal containerd exit events

Athang69:
**Purpose of PR?**:

/tasks/exit handler was using apievents.TaskStart instead of apievents.TaskExit causing GetPid() to always return 0, so destroy was never triggered leaving stale entries in dm.Containers and eBPF NsMap.

Fixes #2526 

**Does this PR introduce a breaking change?**

No

**If the changes in this PR are manually verified, list down the scenarios covered:**
Container exit event correctly unmarshaled, GetPid() returns correct pid, destroy path triggers correctly, /tasks/start handling unaffected.

**Additional information for reviewer?** :
One line fix - apievents.TaskStart to apievents.TaskExit in /tasks/exit case. ContainerID is field1 in both types so container lookup worked fine making the bug subtle - only the pid comparison was silently broken. Added containerdHandler_test.go with 4 unit tests proving the mismatch and verifying the fix.

Issue

**Checklist:**
- [x] Bug fix. Fixes #2526 
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [x] PR Title follows the convention of  `<type>(<scope>): <subject>`
- [x] Commit has unit tests
- [ ] Commit has integration tests

PR #2535, opened 2026-04-08
fix: use correct protobuf type for /tasks/exit in containerd handler

### LINKED ISSUE #2526: bug: /tasks/exit handler unmarshals into wrong protobuf type (TaskStart instead of TaskExit), causing stale container entries
In `handleContainerdEvent` (`KubeArmor/core/containerdHandler.go`), the `/tasks/exit` case unmarshals the event payload into `apievents.TaskStart` instead of `apievents.TaskExit`.

**General Information**
- Environment description: Any environment using containerd as CRI (GKE, kubeadm, minikube, microk8s)
- Kernel version: Any (this is a userspace protobuf bug, not kernel specific)
- Orchestration: Any Kubernetes version using containerd runtime
- Affected file: `KubeArmor/core/containerdHandler.go` in `handleContainerdEvent`
- Target containers: All containers managed via containerd

**To Reproduce**
1. Run KubeArmor with containerd as the CRI
2. Start a container - KubeArmor detects it correctly via `/tasks/start`
3. Stop/exit the container - containerd fires a `/tasks/exit` event
4. KubeArmor unmarshals the exit event into the wrong type `apievents.TaskStart`
5. `GetPid()` returns `0` instead of the real pid due to proto field mismatch
6. The guard `if pid == exitTask.GetPid()` fails silently
7. `UpdateContainerdContainer` destroy is never called
8. Stale entries remain in `dm.Containers` and the eBPF NsMap

**Root Cause**

`TaskStart` and `TaskExit` have different protobuf field numbers:

| Field | TaskStart | TaskExit |
|-------|-----------|----------|
| 1 | ContainerID | ContainerID |
| 2 | Pid | ID (string) |
| 3 | - | Pid |
| 4 | - | ExitStatus |

When a `TaskExit` payload is unmarshaled into `TaskStart`, the `Pid` (field 3 in `TaskExit`) has no counterpart in`TaskStart` so protobuf drops it silently - `GetPid()` always returns `0`.

**Expected behavior**

When a container exits, KubeArmor should correctly read the pid from the exit event, pass the `pid == exitTask.GetPid()` guard, and call `UpdateContainerdContainer` with action `destroy` - cleaning up `dm.Containers` and the eBPF NsMap properly.

**Screenshots**

The following test proves the field mismatch - `TaskStart.GetPid()` returns `0` for a `TaskExit` payload where real pid is `9999`, while `TaskExit.GetPid()` correctly returns `9999`:

Image

**Fix**

One line change in `handleContainerdEvent`:
```go
// before (wrong)
exitTask := &apievents.TaskStart{}

// after (correct)
exitTask := &apievents.TaskExit{}
```

I have a fix ready with 4 unit tests. Will submit a PR.

Bahtya:
## Summary

Fixes #2526

The `/tasks/exit` case in `handleContainerdEvent` (`KubeArmor/core/containerdHandler.go`) was unmarshaling the event payload into `apievents.TaskStart` instead of `apievents.TaskExit`.

## Root Cause

`TaskStart` and `TaskExit` have different protobuf field numbers:

| Field | TaskStart | TaskExit |
|-------|-----------|----------|
| 1 | ContainerID | ContainerID |
| 2 | Pid | ID (string) |
| 3 | - | Pid |
| 4 | - | ExitStatus |

When a `TaskExit` payload is unmarshaled into `TaskStart`, the `Pid` field (field 3 in TaskExit) has no counterpart in `TaskStart`, so protobuf silently drops it. `GetPid()` always returns `0`.

## Impact

- The guard `if pid == exitTask.GetPid()` always fails
- `UpdateContainerdContainer` with action `destroy` is never called
- Stale entries remain in `dm.Containers` and the eBPF NsMap

## Fix

One-line change: `apievents.TaskStart{}` → `apievents.TaskExit{}`

## Testing

Existing containerd integration tests should be verified. The fix is a straightforward protobuf type correction.

### DIFF
--- a/KubeArmor/core/containerdHandler.go
+++ b/KubeArmor/core/containerdHandler.go
@@ -752,7 +752,7 @@ func (dm *KubeArmorDaemon) handleContainerdEvent(envelope *events.Envelope, cont
 		}
 
 	case "/tasks/exit":
-		exitTask := &apievents.TaskStart{}
+		exitTask := &apievents.TaskExit{}
 
 		err := proto.Unmarshal(envelope.Event.GetValue(), exitTask)
 		if err != nil {


Aryan-sharma11:
Hi @Bahtya, thanks for the PR. We currently already have PR #2527 that addresses this issue.

Bahtya:
Thanks for the heads up @Aryan-sharma11! Closing this since #2527 already addresses the issue. Happy to contribute elsewhere.

PR #2537, opened 2026-04-10
fix(operator): guard InitContainers[0] access when init container is disabled

### LINKED ISSUE #2536: Unconditional access to InitContainers[0] causes panic on BTF-enabled nodes
## Bug Report
`generateDaemonset` in `resources.go` unconditionally accesses `InitContainers[0]` after explicitly emptying `InitContainers` to an empty slice, causing an index out of range panic at runtime on any node where BTF is present and `initDeploy` is false which is the common case on modern kernels.
```
if btfPresent != "no" && !initDeploy {
    daemonset.Spec.Template.Spec.InitContainers = []corev1.Container{}
}
```
the slice is genuinely emptied. Immediately after (past the TODO comment), the code unconditionally does:
```
AddOrUpdateEnv(&daemonset.Spec.Template.Spec.InitContainers[0].Env, common.GlobalEnv)
AddOrUpdateEnv(&daemonset.Spec.Template.Spec.InitContainers[0].Env, common.KubeArmorInitEnv)
```
And again near the bottom of the same function:
```
daemonset.Spec.Template.Spec.InitContainers[0].Image = common.GetApplicationImage(common.KubeArmorInitName)
daemonset.Spec.Template.Spec.InitContainers[0].ImagePullPolicy = corev1.PullPolicy(common.KubeArmorInitImagePullPolicy)
```
Both are guaranteed panics. The same pattern exists in `UpdateKubeArmorImages` where init container env updates occur outside the existing `len(InitContainers) != 0` guard
```
if len(ds.Spec.Template.Spec.InitContainers) != 0 {
    // image/args/secrets updates — correctly guarded
}

// BUG: these two are outside the guard
AddOrUpdateEnv(&ds.Spec.Template.Spec.InitContainers[0].Env, common.GlobalEnv)
if len(common.KubeArmorInitEnv) > 0 {
    AddOrUpdateEnv(&ds.Spec.Template.Spec.InitContainers[0].Env, common.KubeArmorInitEnv)
}
```

**General Information**
- Environment: Any Kubernetes environment running on a modern Linux kernel with BTF (BPF Type Format) support this includes GKE, EKS, AKS, kubeadm clusters on kernels ≥ 5.8, minikube, microk8s, etc.
- Kernel version: Any kernel where BTF is present (≥ 5.8 in most distributions; check with ls /sys/kernel/btf/vmlinux)
- Orchestration: Any supported Kubernetes version (kubectl version)
- Affected code: `pkg/KubeArmorOperator/internal/controller/resources.go` (generateDaemonset) and `pkg/KubeArmorOperator/internal/controller/clusterWatcher.go` (UpdateKubeArmorImages)

**To Reproduce**
1. Deploy KubeArmor operator on any cluster whose nodes have BTF present (`/sys/kernel/btf/vmlinux` exists standard on kernels ≥ 5.8)
2. Ensure `initDeploy` is `false` (the default)
3. Trigger daemonset generation e.g. by adding a node or having the operator reconcile
4. Observe the operator pod crash with `runtime error: index out of range [0] with length 0`

**Expected behavior**
The operator should correctly skip init container configuration when `InitContainers` has been cleared (BTF present, no init deploy needed), consistent with the guard pattern already used elsewhere in both functions. No panic should occur.

saschabuehrle:
Fixes #2536

On BTF-enabled nodes with `initDeploy=false`, the daemonset intentionally clears `InitContainers`, but a few unconditional `InitContainers[0]` updates still execute and can panic. This change adds length guards around all remaining init-container env/image updates in both daemonset generation and image refresh paths.

Greetings, saschabuehrle

### DIFF
--- a/pkg/KubeArmorOperator/internal/controller/cluster.go
+++ b/pkg/KubeArmorOperator/internal/controller/cluster.go
@@ -600,12 +600,14 @@ func (clusterWatcher *ClusterWatcher) UpdateKubeArmorImages(images []string) err
 						ds.Spec.Template.Spec.Tolerations = append(ds.Spec.Template.Spec.Tolerations, common.KubeArmorInitTolerations...)
 					}
 
-					// update with global env
-					AddOrUpdateEnv(&ds.Spec.Template.Spec.InitContainers[0].Env, common.GlobalEnv)
-					// add/override with kubearmor-init specific env
-					if len(common.KubeArmorInitEnv) > 0 {
-						defer RemoveDeletedEntriesForEnv(&common.KubeArmorInitEnv)
-						AddOrUpdateEnv(&ds.Spec.Template.Spec.InitContainers[0].Env, common.KubeArmorInitEnv)
+					if len(ds.Spec.Template.Spec.InitContainers) != 0 {
+						// update with global env
+						AddOrUpdateEnv(&ds.Spec.Template.Spec.InitContainers[0].Env, common.GlobalEnv)
+						// add/override with kubearmor-init specific env
+						if len(common.KubeArmorInitEnv) > 0 {
+							defer RemoveDeletedEntriesForEnv(&common.KubeArmorInitEnv)
+							AddOrUpdateEnv(&ds.Spec.Template.Spec.InitContainers[0].Env, common.KubeArmorInitEnv)
+						}
 					}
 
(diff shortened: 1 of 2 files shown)


Athang69:
Hey @saschabuehrle, I'm the one who filed issue #2536. Thanks for picking this up quickly the fix looks correct to me. I had a partial fix ready but missed the third location (the unconditional `Image/ImagePullPolicy` assignments at the bottom of `generateDaemonset`), so your fix is more complete.

PR #2543, opened 2026-04-12
fix(operator): return error on non-200 metadata response and fix log format verbs

Athang69:
**Purpose of PR?**:
Fixes two bugs in `fetchClusterNameFromGKE` and `fetchClusterNameFromAWS` in `resources.go`:

1. When the metadata server returns a non-200 status code, both functions were returning `("", nil)` because `err` is nil after a successful HTTP round-trip. The caller `GetClusterName` checks `if err != nil` to trigger fallback to `"default"`, so a 404 or 500 from the metadata server would silently return an empty cluster name instead of falling back correctly. Fixed by returning `fmt.Errorf("unexpected status code: %d", resp.StatusCode)`.

2. Two `Warnf` calls were using `%d` (integer format verb) to format `error` values, producing garbled log output like `%!d(*errors.errorString=&{...})`. Fixed to use `%s` with `err.Error()`.

No.

- Metadata server returns non-200 status: `GetClusterName` now correctly falls back to `"default"` instead of returning an empty cluster name
- Error log lines in `fetchClusterNameFromAWS` now produce readable output

Four lines changed across two functions in `resources.go`. No logic changes outside of error handling.

**Checklist:**
- [x] Bug fix. Fixes #
- [ ] Commit has unit tests

### DIFF
--- a/pkg/KubeArmorOperator/internal/controller/resources.go
+++ b/pkg/KubeArmorOperator/internal/controller/resources.go
@@ -686,7 +686,7 @@ func (clusterWatcher *ClusterWatcher) fetchClusterNameFromGKE(providerHostname, 
 	// Check for a successful response
 	if resp.StatusCode != http.StatusOK {
 		clusterWatcher.Log.Warnf("failed to fetch from metadata, status code: %d", resp.StatusCode)
-		return "", err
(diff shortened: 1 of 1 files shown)


Athang69:
Hi @Aryan-sharma11, gentle ping on this PR would appreciate your review when possible. Thanks!

PR #2579, opened 2026-05-06
test(tests): add network default-deny posture assertions in block posture suite

### LINKED ISSUE #2578: Missing network block posture tests in tests/k8s_env/blockposture/
KubeArmor supports namespace-level default-deny posture via annotations (kubearmor-network-posture=block, etc.). When an Allow policy is defined for one resource class (e.g. files) but silent on others, the namespace default should block the rest.
Current gap:
The suite annotates the nginx-posture namespace with both file and network block posture, then applies ksp-nginx-allow-file.yaml (file Allow only). All test assertions only check File operations. There are zero tests for:

Outbound network traffic being blocked (curl/wget/nc)
DefaultPosture Network alert emission
Removing the network block annotation to restore egress

Impact: Network posture can regress completely and the test suite will still pass - a silent failure in Zero-Trust environments.
Repro: Run the suite + curl -m 5 1.1.1.1 from the pod → traffic is blocked, but tests pass.
Expected: Add assertions verifying network denial under default block posture, alert generation, and connectivity restoration when the annotation is removed.

nXtCyberNet:
### Purpose of PR

Closes the network default-deny posture assertion gap in the block posture integration test suite.

Close : #2578 

### Does this PR introduce a breaking change?

No.

### If the changes in this PR were manually verified, list the scenarios covered:

* Verified that network egress is blocked and a `DefaultPosture` network alert is emitted when `kubearmor-network-posture=block` is set and no network `Allow` rule exists
* Verified that no unexpected network alerts are generated during allowed file operations (negative test case)
* Verified that egress connectivity is restored after removing the `kubearmor-network-posture=block` annotation
* Tested on a Kind cluster with the AppArmor enforcer enabled

### Additional information for reviewers

This PR continues the block posture test improvements.

Changes included:

* Added a reusable `probeNetworkBinary()` helper to gracefully handle minimal/distroless container images
* Increased alert timeout to 15s to reduce CI flakiness under heavy load

### Checklist

* [x] Bug fix / Test improvement
* [ ] New feature
* [ ] Breaking change
* [ ] Documentation update required
* [x] PR title follows the convention: `<type>(<scope>): <subject>`
* [ ] Unit tests
* [x] Integration tests

### DIFF
--- a/tests/k8s_env/blockposture/block_test.go
+++ b/tests/k8s_env/blockposture/block_test.go
@@ -4,6 +4,7 @@
 package blockposture
 
 import (
+	"strings"
 	"time"
 
 	"github.com/kubearmor/KubeArmor/tests/util"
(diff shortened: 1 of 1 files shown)


Aryan-sharma11:
Hi @nXtCyberNet , we already have network block posture tests 
https://github.com/kubearmor/KubeArmor/blob/main/tests/k8s_env/networktests/network_test.go#L245

nXtCyberNet:
Hi @Aryan-sharma11 , thanks for the review. Sorry for not having seen that file earlier.

These tests are complementary to the existing networking tests, however the following two cases are not covered in `networktests` currently:

- **Annotation removal test** — confirming egress is re-allowed after `kubearmor-network-posture=block` is removed.
- **Cross-operation noise check** — ensuring a file operation does not accidentally trigger a Network alert.

If you think these are valuable cases, I can add them directly into `network_test.go` instead of keeping them in a separate package.

Also, while I was trying to run it in my k3s setup on Ubuntu, I faced an issue with port-level control not being supported. I would like to know if you think this is a worthwhile advancement — if so, I can open a PR for the same. I have more details in #2580.

Thanks!

PR #2598, opened 2026-05-13
fix(core): Use k8s informers to watch host and network policies

### LINKED ISSUE #1057: Deprecate k8s watchers in favor of k8s informers
## Feature Request

**Short Description**

Currently we are making use of kubernetes watchers to listen for resource events.
This has two side effect:
- Performance impact as we querying the api server at a constant rate even if there is no change in the data.
- Event loss.

We should make use of k8s informers in order to solve the above mentioned problems.

- [ ] Use client go to make API calls #531 
- [ ] Implement share informers in order to watch the required resources

AryanBakliwal:
**Purpose of PR?**:
When running in K8s mode, after around 30-40 min. of deploying we get these logs
```
timeout while monitoring host security policies, kubearmorhostpolicies CRD not found
timeout while monitoring network security policies, kubearmornetworkpolicies CRD not found
```
Even though the CRDs are present in the cluster. Cause explained [here](https://stackoverflow.com/a/33487394).
This PR fixes this by using K8s informers for watching host and network policies, similar to watching cluster security policies.

Also fixes #1057 

No

**If the changes in this PR are manually verified, list down the scenarios covered**:

**Checklist:**
- [x] Bug fix. Fixes #1057 
- [ ] Commit has unit tests


KubeArmor/core/kubeArmor.go
```diff
@@ -985,77 +985,109 @@ func KubeArmor() {
...
+			clusterSecurityPoliciesSynced cache.InformerSynced
+			hostSecurityPoliciesSynced    cache.InformerSynced
+			networkSecurityPoliciesSynced cache.InformerSynced
+			defaultPostureSynced          cache.InformerSynced
+			configMapSynced               cache.InformerSynced
+		)
+
+		if cfg.GlobalCfg.Policy {
+			// watch security policies
+			securityPoliciesSynced = dm.WatchSecurityPolicies()
```
rksharma95: let's add timeout to `WatchSecurityPolicies()` as well.

rksharma95 (changes requested):
LGTM, just a suggestion commented inline.

KubeArmor/core/kubeArmor.go
AryanBakliwal: added

rksharma95 (approved):

Aryan-sharma11 (approved):

PR #2601, opened 2026-05-13
skip AppArmor/OCI-hook mounts on BPF-only nodes

### LINKED ISSUE #2592: KubeArmor snitch fails on Bottlerocket nodes (ContainerStatusUnknown, FailedMount)
General Information

Environment description: Amazon EKS cluster with Bottlerocket nodes, KubeArmor installed via Helm chart.

Kernel version:
  Bottlerocket 1.57.0 nodes: Linux kernel 6.12.73
  Bottlerocket 1.60.0 nodes: Linux kernel 6.12.79

Orchestration system version in use: Kubernetes v1.34.4-eks-f69f56f

Target containers/pods:
Pods matching kubearmor-snitch in namespace kubearmor
Image: kubearmor-snitch

Problem Summary

In this EKS Bottlerocket cluster, many kubearmor-snitch pods are repeatedly created by the operator and fail with ContainerStatusUnknown.
The failures are associated with hostPath mount errors:

MountVolume.SetUp failed for volume kubearmor-dir: mkdir /usr/share/kubearmor: read-only file system
MountVolume.SetUp failed for volume hook-dir: mkdir /usr/share/containers: read-only file system
MountVolume.SetUp failed for volume apparmor-path: hostPath type check failed: [apparmor.d] is not a directory

For failing pods, the container often shows:

Reason: ContainerStatusUnknown
Message: The container could not be located when the pod was terminated
Exit code: 137
Important Correlation Observed

Nodes on Bottlerocket 1.57.0 are missing kubearmor.io/lsm label.
Nodes on Bottlerocket 1.60.0 usually have kubearmor.io/lsm=bpf.
On nodes without kubearmor.io/lsm, snitch pods include apparmor-path, hook-dir, and kubearmor-dir mounts and fail.
On nodes with kubearmor.io/lsm=bpf, some snitch pods complete successfully with a reduced set of mounts.
This suggests node capability detection and snitch mount generation may be inconsistent in mixed Bottlerocket versions.

To Reproduce

Create or use an EKS cluster with Bottlerocket nodes, including at least one older node where [apparmor.d] is not present and /usr/share paths are read-only.
Install KubeArmor operator/chart version aligned with KubeArmor 1.6.16 images.
Ensure kubearmor-snitch jobs are triggered by the operator for nodes joining or restarting.
Observe snitch pods in namespace kubearmor:
many pods fail with ContainerStatusUnknown
events show FailedMount for apparmor-path, hook-dir, and kubearmor-dir.

Expected behavior

On Bottlerocket nodes, snitch should not attempt incompatible hostPath mounts such as /etc/apparmor.d, /usr/share/kubearmor, or /usr/share/containers/oci/hooks.d when those paths are unsupported or read-only.
Snitch jobs should complete deterministically without entering continuous failed retries.
ContainerStatusUnknown should not occur due to mount-path incompatibility.
Observed behavior

Operator continuously creates snitch jobs.
Many snitch pods fail with mount errors and end as ContainerStatusUnknown.
Cluster gets noisy and partially unstable from repeated retries and failed transient pods.

Screenshots

Image

Labels add in nodes:
But these labels I enforce in node:
- kubearmor.io/runtime: containerd
- kubearmor.io/lsm: bpf
Image

System Info:

Image

Mandalorian7773:
Fixes #2592 

On Bottlerocket (and similar minimal distributions), multiple hostPath mounts (`apparmor-path`, `hook-dir`, `kubearmor-dir`) fail simultaneously. The existing `checkJobStatus` handler only removed **one** failing volume per retry cycle (it `break`ed after the first match), requiring multiple successful retry cycles. Combined with the pod being terminated quickly (exit 137 / `ContainerStatusUnknown`), the handler could never converge, causing an infinite loop of failed snitch jobs.

This PR fixes `checkJobStatus` to collect **all** `FailedMount` / `ReadOnly-FS` events in a single pass and remove every failing volume at once, so the retry Job succeeds on the first attempt.

No.

- [x] Deployed the modified operator on a local k3s cluster and verified the snitch job completes successfully without errors.
- [x] Verified operator logs show clean job watching with no `FailedMount` events or crash loops.

Revised approach based on feedback from @rksharma95. Instead of trying to pre-check the node's enforcer label (chicken-and-egg problem since snitch creates that label), this fix improves the existing `checkJobStatus` mount-failure handler to handle multiple simultaneous failures in one cycle.

### Local Verification Evidence:

Screenshot 2026-05-14 at 2 01 42 PM

**Checklist:**
- [x] Bug fix. Fixes #2592
- [ ] Commit has unit tests

### DIFF
--- a/pkg/KubeArmorOperator/internal/controller/cluster.go
+++ b/pkg/KubeArmorOperator/internal/controller/cluster.go
@@ -170,8 +170,13 @@ func (clusterWatcher *ClusterWatcher) checkJobStatus(job, runtime, nodename stri
 			}
 
 			for _, pod := range podsList.Items {
-				mountFailure := false
-				failedMount := ""
+				// Collect ALL failing mounts from events instead of just the first one.
+				// On distributions like Bottlerocket, multiple hostPath mounts
+				// (apparmor-path, hook-dir, kubearmor-dir) fail simultaneously.
+				// Removing them all in one retry avoids an infinite loop where
+				// the pod is terminated before the next poll can catch subsequent
+				// failures (fixes #2592).
+				var failedMounts []string
 				events, err := clusterWatcher.Client.CoreV1().Events(common.Namespace).List(context.TODO(), v1.ListOptions{
 					FieldSelector: fmt.Sprintf("involvedObject.name=%s", pod.Name),
 				})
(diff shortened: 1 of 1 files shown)


Mandalorian7773:
Hi @rksharma95  please review

rksharma95 (changes requested):
@Mandalorian7773 the problem with this approach is that, snitch itself adds the `kubearmor.io/enforcer` label to the node, so until unless snitch gets deployed on the node it would not work.

we're handling the mountFailure issue here, I would suggest to check why this failed to handle the same on bottlerocket env. https://github.com/kubearmor/KubeArmor/blob/bcfdb951aa037cecac8a58d0f68e3412278ed7db/pkg/KubeArmorOperator/internal/controller/cluster.go#L140

Mandalorian7773:
@rksharma95  Thanks for the feedback! You're right, I've revised the approach. Please take another look!

rksharma95 (approved):

Mandalorian7773:
Hi, @Aryan-sharma11 and @AryanBakliwal, its been 3 months. Can you please drop a final review and merge this PR.

Aryan-sharma11 (approved):

PR #2627, opened 2026-05-25
fix(core): fix lock-order inversion deadlock in UpdateDefaultPosture

### LINKED ISSUE #2524: Potential deadlock in default-posture update path due to lock-order inversion
There appears to be a deadlock risk in the default-posture/config update flow.

The code acquires `EndPointsLock` and `DefaultPosturesLock` in inconsistent order across different update paths:

- `UpdateDefaultPosture()` takes `DefaultPosturesLock` first, then later uses `EndPointsLock`
- `UpdateGlobalPosture()` takes `EndPointsLock` first, then `DefaultPosturesLock`
- `updatEndpointsWithCM()` also takes `EndPointsLock` first, then `DefaultPosturesLock`

Because these functions are triggered by different informers, two goroutines can end up waiting on each other.

## Affected paths

- `core/kubeUpdate.go`
  - `UpdateDefaultPosture()`
  - `UpdateGlobalPosture()`
  - `updatEndpointsWithCM()`
- `core/kubeArmor.go`
  - `WatchDefaultPosture()`
  - `WatchConfigMap()`

## Why this matters

`WatchDefaultPosture()` and `WatchConfigMap()` are both started during normal daemon startup. That means this is a real runtime path, not just a theoretical concurrency issue.

A namespace posture update and a configmap posture update happening at the same time can potentially deadlock posture propagation.

## Repro scenario

1. Start KubeArmor with Kubernetes policy/default posture handling enabled.
2. Trigger a namespace-level default posture update.
3. At the same time, trigger a `kubearmor-configmap` update that changes global posture.
4. Both informer handlers can enter posture-update code concurrently.

## Expected behavior

Default posture updates should complete without blocking each other, and endpoint posture/enforcement state should converge correctly.

## Actual behavior

There is a lock-order inversion between the namespace posture path and the configmap posture path, which can potentially deadlock the handlers and stall posture propagation.

## Code details

Current lock ordering is inconsistent:

- `UpdateDefaultPosture()`: `DefaultPosturesLock` -> `EndPointsLock`
- `UpdateGlobalPosture()`: `EndPointsLock` -> `DefaultPosturesLock`
- `updatEndpointsWithCM()`: `EndPointsLock` -> `DefaultPosturesLock`

## Suggested fix

Standardize lock ordering across all default-posture update paths, or refactor these paths so both locks are not held together.

It would also be useful to add a concurrency test that updates namespace posture and configmap posture in parallel.

asmit27rai:
Fixed a potential deadlock caused by inconsistent lock acquisition order between `DefaultPosturesLock` and `EndPointsLock` during concurrent namespace posture and configmap posture updates. The issue occurred because `UpdateDefaultPosture()` acquired locks in the reverse order compared to `UpdateGlobalPosture()` and `updatEndpointsWithCM()`.

To resolve this, the locking strategy was standardized across all related functions by always acquiring `EndPointsLock` first, followed by `DefaultPosturesLock`. The fix includes refactoring `UpdateDefaultPosture()` in `kubeUpdate.go`, adding documentation comments for the lock-order convention, and introducing the regression test `TestNoDeadlockOnConcurrentPostureUpdates`.

Fix : #2524


asmit27rai:
@rksharma95 Please have a look at this.

PR #2635, opened 2026-05-28
fix(test): use KarmorGetTargetAlert in blockposture to fix intermittent CI failures

### LINKED ISSUE #2634: test(blockposture): KarmorGetLogs 10s timeout causes intermittent CI failures on loaded runners
## Bug Report

**General Information**

- Environment: `oracle-vm-16cpu-64gb-x86-64` self-hosted GitHub Actions runner
- Orchestration: k3s with containerd runtime
- Affected test: `tests/k8s_env/blockposture/block_test.go:89`
- Affected suite: `Auto-testing Framework / oracle-vm-16cpu-64gb-x86-64 / containerd`

**To Reproduce**

1. Raise any PR that triggers `ci-test-ginkgo` workflow
2. Wait for the `Auto-testing Framework / oracle-vm-16cpu-64gb-x86-64 / containerd` matrix job to run
3. Observe the `Test KubeArmor using Ginkgo` step failing in the `blockposture` suite with `event timeout`

**Expected behavior**

The test `can whitelist certain files accessed by a package while blocking all other sensitive content` should pass. The block itself works correctly as seen in the logs: `cat: can't open 'docker-entrypoint.sh': Permission denied`. KubeArmor is enforcing the policy but `KarmorGetLogs(10*time.Second, 1)` times out before the policy violation alert arrives over gRPC, returning 0 alerts. The assertion `Expect(len(alerts)).To(BeNumerically(">=", 1))` then fails with `Expected <int>: 0 to be >= <int>: 1`.

The test retries 9 times and fails on every attempt totalling 322 seconds before giving up. The failure is intermittent and runner-load-dependent. The same test passes on other CI runs when the runner is less loaded, confirming this is a timing issue in the eBPF event delivery pipeline rather than a functional KubeArmor bug.

The fix is to increase the `KarmorGetLogs` timeout from `10*time.Second` to `60*time.Second` in `block_test.go:89`:

```go
// before
_, alerts, err := KarmorGetLogs(10*time.Second, 1)

// after
_, alerts, err := KarmorGetLogs(60*time.Second, 1)
```

**Screenshots**
Image

Athang69:
**Purpose of PR?**: The `blockposture` suite fails intermittently on the `containerd` CI job due to a race condition in how `KarmorGetLogs` handles mixed event types.

**Fixes:** #2634

**Root cause**: `KarmorGetLogs` counts both `Log` and `Alert` events toward the `maxEvents` limit. With `maxEvents: 1`, the first event off the channel is a `Log` from the allowed `cat /usr/share/nginx/html/index.html` command. This hits `evtCnt >= 1` and triggers early exit before the `Alert` for the blocked `cat docker-entrypoint.sh` arrives, returning an empty alerts slice. The 10s timeout is never the bottleneck - the function exits immediately on the first log event regardless of runner load.

**Fix**: Switch to `KarmorGetTargetAlert` which ignores `Log` events entirely and only returns when it finds an alert matching `PolicyName` and `Action`, or the timeout fires. The original 10s timeout is kept unchanged. No CI time increase.

**Does this PR introduce a breaking change?** No.

**If the changes in this PR are manually verified, list down the scenarios covered?**: The fix is verifiable by code inspection - `KarmorGetTargetAlert` cannot exit early on a log event by construction. `go build` and `go vet` pass cleanly on the changed file. The intermittent nature of the original failure (runner-load-dependent) makes local reproduction impractical, but the root cause is confirmed by reading `KarmorGetLogs` in `tests/util/karmorlog.go`: the `evtCnt` counter increments on every event type, not just alerts.

**Additional information for reviewer?**: All other `KarmorGetLogs` call sites use `maxEvents: 50` and collect logs in bulk, so they are unaffected by this bug. Only `blockposture` and `throttling` use `maxEvents: 1` while caring only about alerts - `blockposture` is fixed here. The `throttling` tests may have the same latent issue.

**Checklist:**
- [x] Bug fix. Fixes #2634
- [ ] Commit has unit tests

### DIFF
--- a/tests/k8s_env/blockposture/block_test.go
+++ b/tests/k8s_env/blockposture/block_test.go
@@ -6,6 +6,7 @@ package blockposture
 import (
 	"time"
 
+	pb "github.com/kubearmor/KubeArmor/protobuf"
 	"github.com/kubearmor/KubeArmor/tests/util"
 	. "github.com/kubearmor/KubeArmor/tests/util"
 	. "github.com/onsi/ginkgo/v2"
@@ -84,11 +85,12 @@ var _ = Describe("Posture", func() {
 				MatchRegexp("<!DOCTYPE html>((?:.*\r?\n?)*)</html>"), true,
 			)
 			// check policy violation alert
-			_, alerts, err := KarmorGetLogs(10*time.Second, 1)
+			res, err := KarmorGetTargetAlert(10*time.Second, &pb.Alert{
+				PolicyName: "DefaultPosture",
+				Action:     "Block",
+			})
 			Expect(err).To(BeNil())
-			Expect(len(alerts)).To(BeNumerically(">=", 1))
-			Expect(alerts[0].PolicyName).To(Equal("DefaultPosture"))
-			Expect(alerts[0].Action).To(Equal("Block"))
+			Expect(res.Found).To(BeTrue())
 		})
 	})
 


Athang69:
@Aryan-sharma11 @rksharma95 PTAL Whenever convenient.
Thankss!!

Aryan-sharma11:
@Athang69 Increasing the timeout by 6x, especially given that we already have 10 retries, would significantly increase the overall CI execution time. The alert should ideally be generated within the existing 10-second timeout. If it isn't, I think we should investigate why the alert is taking that long to appear rather than increasing the timeout.

Athang69:
> @Athang69 Increasing the timeout by 6x, especially given that we already have 10 retries, would significantly increase the overall CI execution time. The alert should ideally be generated within the existing 10-second timeout. If it isn't, I think we should investigate why the alert is taking that long to appear rather than increasing the timeout.

@Aryan-sharma11 You're right, I investigated further and the timeout is not the actual problem.
Root cause: `KarmorGetLogs` counts both `Log` and `Alert` events toward the `maxEvents` limit. With `maxEvents: 1`, the first event off the channel is a `Log` from the allowed `cat /usr/share/nginx/html/index.html` command, which hits `evtCnt >= 1` and triggers early exit before the `Alert` for the blocked `cat docker-entrypoint.sh` arrives. The function returns an empty alerts slice immediately - the 10s timeout never even fires.
I've updated the PR to switch to `KarmorGetTargetAlert` instead, which ignores `Log` events entirely and only exits on a matching alert or timeout. The original 10s timeout is kept unchanged, no CI time increase.
Also worth noting - `throttling_test.go` uses `KarmorGetLogs(5*time.Second, 1)` with the same pattern and may have the same latent issue, but I've left that for a separate fix to keep this PR scoped.
PTAL.
Thanks!

achrefbensaad:
This PR lgtm, rebased it, waiting for CI to complete

achrefbensaad (approved):
LGTM

Aryan-sharma11 (approved):

PR #2658, opened 2026-06-08
fix(operator): allow statx in daemonset seccomp profiles

### LINKED ISSUE #2594: statx syscall missing from daemonset seccomp profile
## Bug Report

**General Information**

- Environment description: EKS 1.34 bottlerocket
- Kernel version: 6.12.79 x86_64 x86_64 x86_64 GNU/Linux
- Orchestration system version in use: k8s 1.34
- Link to relevant artifacts: https://github.com/kubearmor/KubeArmor/blob/v1.7.0/pkg/KubeArmorOperator/seccomp/seccomp.go
- Target containers/pods: kubearmor-bpf-containerd-xxxxx DaemonSet

**To Reproduce**

1. Install kubearmor via operator and then controller & relay via KubeArmorConfig
2. Restart the operator pod

**Expected behavior**

All pods should start without any issues

**Actual behaviour**
The dameonset pods were failing, and stuck with following error:

`Error: failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error closing exec fds: get handle to /proc/thread-self/fd: check "thread-self" component is not overmounted: get root mount id: statx(STATX_MNT_ID_...) fsmount:fscontext:proc/: could not get mount id: operation not permitted`

Valyrian-Code:
## Purpose

The operator generates two default-deny (`ActErrno`) seccomp profiles in `pkg/KubeArmorOperator/seccomp/seccomp.go`:

- `kubearmor-seccomp.json` (KubeArmor DaemonSet container)
- `kubearmor-init-seccomp.json` (init container)

Both allow the stat-family syscalls (`stat`, `fstat`, `newfstatat`, `statfs`) but omit `statx`. Newer runc on 6.12 kernels (EKS Bottlerocket, per #2594) calls `statx(STATX_MNT_ID_...)` during container init, so the missing entry blocks it and the KubeArmor DaemonSet pods fail to start:

```
error during container init: error closing exec fds: ... statx(STATX_MNT_ID_...) ... operation not permitted
```

This adds `statx` to both allow-lists, next to the other stat-family syscalls.

## Does this PR introduce a breaking change?

No — it only permits one additional syscall in the generated profiles.

Fixes #2594


Valyrian-Code:
Friendly ping, this small seccomp fix is ready whenever someone has a chance to review. Thanks!

PR #2690, opened 2026-06-23
cert: fix GenerateCA returning nil error on self-signed cert failure

### LINKED ISSUE #2689: cert/cert.go: GenerateCA returns nil error on self-signed cert generation failure
## What

In the \GenerateCA\ function within the certificate management module, when the underlying self-signed certificate generation fails, the function does not propagate the error back to its caller. Instead, it returns an empty certificate structure while reporting success via a nil error value. This causes any caller of \GenerateCA\ to incorrectly believe that a valid CA certificate and key pair was successfully created.

## Why

The root cause is a single incorrect return statement in the error handling path. After receiving an error from the self-signed certificate generation call, the function returns an empty result with a nil error instead of forwarding the actual error. This means the failure is silently consumed — logged and discarded — rather than surfaced to the calling code.

This contrasts with the immediately preceding error handling block in the same function, which correctly propagates the error, suggesting this was an oversight or copy-paste mistake.

## How to Reproduce

1. Set up the KubeArmor binary with any configuration that triggers certificate generation (e.g., default self-signed TLS setup).
2. Induce a failure in the self-signed certificate generation path. This can be done by providing an invalid certificate configuration that causes key generation or certificate template creation to fail.
3. Observe that \GenerateCA\ returns without error despite the failure.
4. Observe that the returned certificate and key are empty or zero-value bytes.

The downstream effect manifests in any codepath that depends on \GenerateCA\ output — for example, the operator's PKI initialization and the gRPC server's TLS setup will proceed with empty certificates, leading to runtime failures or insecure connections that are difficult to diagnose because the original error was silenced.

## Impact

- The gRPC log server and state server may start with invalid or empty TLS certificates
- The operator may deploy with broken PKI
- The actual failure appears as a hard-to-trace symptom (e.g., TLS handshake failures) rather than a clear certificate generation error
- Debugging requires tracing through multiple layers since the root cause was discarded

## Approach to Fix

The fix is a single change in the error return statement: replace the hardcoded nil error with the actual error variable. This ensures that when self-signed certificate generation fails, the error propagates correctly to the caller, which can then handle it appropriately — either by retrying, falling back, or failing with a clear diagnostic message.

No structural changes are needed — only the return value needs correction.

Aamod007:
## What

\GenerateCA()\ silently swallows errors from \GenerateSelfSignedCert()\. When self-signed certificate generation fails, the function returns an empty \CertBytes\ struct with a \
il\ error, causing callers to believe certificate generation succeeded.

## Why

The error return was hardcoded to \
il\ instead of propagating the actual error. This was a copy-paste mistake — the preceding error handling block correctly returns the error.

## Impact

Callers like the operator's PKI initialization and gRPC TLS setup proceed with zero-value cert/key bytes, leading to runtime TLS failures that are difficult to diagnose because the root cause error was discarded.

## How to Reproduce

Trigger a failure in \GenerateSelfSignedCert\ (e.g., invalid cert config). \GenerateCA\ logs the error but returns nil error, leaving callers with an empty certificate.

## Fix

Changed \
eturn &CertBytes{}, nil\ to \
eturn &CertBytes{}, err\ — a one-line change.

Fixes #2689

### DIFF
--- a/KubeArmor/cert/cert.go
+++ b/KubeArmor/cert/cert.go
@@ -208,7 +208,7 @@ func GenerateCA(cfg *CertConfig) (*CertBytes, error) {
 	}
 	crtBytes, err := GenerateSelfSignedCert(crtTemp, cfg)
 	if err != nil {
-		return &CertBytes{}, nil
+		return &CertBytes{}, err
 	}
 	return &CertBytes{
 		Crt: crtBytes.Crt,
--- a/KubeArmor/cert/cert_test.go
+++ b/KubeArmor/cert/cert_test.go
@@ -1,0 +1,81 @@
+package cert
+
+import (
+	"crypto/x509"
+	"testing"
+)
+
+func TestGenerateCA_Success(t *testing.T) {
+	cfg := &CertConfig{
+		CN:           "test-ca",
+		Organization: "test-org",
+		IsCa:         true,
+		KeyUsage:     x509.KeyUsageCertSign,
+	}
+
+	certBytes, err := GenerateCA(cfg)
+	if err != nil {
+		t.Fatalf("GenerateCA() expected no error, got: %v", err)
+	}
+	if len(certBytes.Crt) == 0 {
+		t.Error("GenerateCA() returned empty certificate")
+	}
+	if len(certBytes.Key) == 0 {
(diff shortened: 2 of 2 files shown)


Aamod007:

PR #2692, opened 2026-06-23
fix: return admission error when pod fetch fails in binding handler

### LINKED ISSUE #2691: pod_mutation.go: nil pod pointer used after failed API call in PodAnnotator Handle
## What

In the Kubernetes controller's pod mutation webhook handler, when processing a pod binding event, the code fetches the pod object from the Kubernetes API server. If that API call fails, the error is logged but execution continues unconditionally. The resulting nil or zero-value pod pointer is then passed to the AppArmor annotation function, which will panic when attempting to access fields on a nil object.

## Why

On line 52, after the Kubernetes API call to get the pod fails, the error is only logged. There is no early return or error response sent back to the admission webhook. The code proceeds to read the target node name from the binding, evaluate cluster state, and ultimately call the annotation helper with a pod pointer that may be nil.

A Kubernetes API call can fail for many transient reasons — network interruption, API server timeout, resource contention, or RBAC misconfiguration. In all such cases, the webhook handler should return an error response rather than continuing with invalid state.

## How to Reproduce

1. Deploy the KubeArmor admission controller with the pod mutation webhook enabled.
2. Cause the Kubernetes API server to be unreachable or return an error when the webhook attempts to fetch pod metadata (e.g., network partition, RBAC denial, or throttling).
3. Submit a pod binding request.
4. The webhook catches the error, logs it, but does not return an admission error response.
5. A nil pod pointer is passed to AppArmor annotation logic, causing a nil pointer dereference panic in the webhook handler.

## Impact

- The admission webhook process crashes with a nil pointer dereference panic
- All subsequent pod admission requests are rejected or ignored until the webhook pod restarts
- In the worst case, this can cause a denial of service for pod scheduling in the cluster
- The root cause is masked by the crash dump, making debugging harder than necessary

## Approach to Fix

After the error check on the pod fetch result, return an admission errored response instead of only logging. This ensures that when the API call fails:
- The admission request is properly rejected with a descriptive error
- The webhook continues serving other requests without crashing
- The caller receives immediate feedback about the failure

The fix is adding a single return statement after the error log.

Aamod007:
## What

In the pod mutation webhook's binding event handler, when the Kubernetes API call to fetch the pod fails, the handler logs the error but continues execution. The resulting nil pod pointer is then passed to \AppArmorAnnotator\, which immediately accesses \pod.ObjectMeta\, causing a nil pointer dereference panic.

## Why

The error path after the API call was missing a return statement. The error was only logged, but no admission error response was sent back, allowing the code to proceed with an invalid pod pointer.

## Fix

Added \
eturn admission.Errored(http.StatusInternalServerError, err)\ after logging the error. This ensures the webhook properly rejects the admission request instead of panicking.

Fixes #2691

### DIFF
--- a/KubeArmor/cert/cert.go
+++ b/KubeArmor/cert/cert.go
@@ -208,7 +208,7 @@ func GenerateCA(cfg *CertConfig) (*CertBytes, error) {
 	}
 	crtBytes, err := GenerateSelfSignedCert(crtTemp, cfg)
 	if err != nil {
-		return &CertBytes{}, nil
+		return &CertBytes{}, err
 	}
 	return &CertBytes{
 		Crt: crtBytes.Crt,
--- a/KubeArmor/cert/cert_test.go
+++ b/KubeArmor/cert/cert_test.go
@@ -1,0 +1,81 @@
+package cert
+
+import (
+	"crypto/x509"
+	"testing"
+)
+
+func TestGenerateCA_Success(t *testing.T) {
+	cfg := &CertConfig{
+		CN:           "test-ca",
+		Organization: "test-org",
+		IsCa:         true,
+		KeyUsage:     x509.KeyUsageCertSign,
+	}
+
+	certBytes, err := GenerateCA(cfg)
+	if err != nil {
+		t.Fatalf("GenerateCA() expected no error, got: %v", err)
+	}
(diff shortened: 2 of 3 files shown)


Aamod007:

PR #2694, opened 2026-06-23
fix: return error when /etc/passwd cannot be opened in podman hook

### LINKED ISSUE #2693: podman Hook: nil pointer dereference when /etc/passwd cannot be opened
## What

In the Podman OCI hook binary, when resolving the user's home directory from /etc/passwd, if the file cannot be opened the error is logged but execution continues. A deferred Close() call on the nil file pointer causes a nil pointer dereference panic, crashing the hook process.

## Why

When os.Open("/etc/passwd") fails, the returned file handle is nil. The error is logged but there is no early return, so the deferred passwdFile.Close() executes on a nil pointer. A deferred method call on a nil interface value in Go produces a panic because the method dispatch requires a non-nil receiver.

This can happen in container environments with minimal filesystem images, permission restrictions, or when the hook runs before the root filesystem is fully mounted.

## How to Reproduce

1. Build and install the Podman OCI hook binary.
2. Run it in an environment where /etc/passwd does not exist or is unreadable (e.g., a minimal container with no passwd file, or a restricted mount namespace).
3. Ensure the HOME environment variable is not set so the code enters the affected path.
4. The hook crashes with a nil pointer dereference panic instead of handling the missing file gracefully.

## Impact

When this triggers:
- The OCI hook process crashes, preventing container creation from proceeding correctly
- Podman container operations that depend on the hook fail silently or produce cryptic errors
- The crash occurs during container lifecycle hooks, which can leave containers in an inconsistent state

## Approach to Fix

After logging the open failure, return an error immediately instead of falling through to the nil defer. This ensures the function exits cleanly and the caller can handle the failure appropriately.

The fix is adding a return statement inside the error branch after the log.

Aamod007:
## What

In the Podman OCI hook binary, the `run()` function attempts to determine the user's home directory when the `HOME` environment variable is not set. It falls back to parsing `/etc/passwd`. If opening `/etc/passwd` fails, the error is logged but execution continues. A deferred `Close()` call on the nil file pointer causes a nil pointer dereference panic, crashing the hook process.

## Root Cause

When `os.Open("/etc/passwd")` fails, the returned `*os.File` is nil. The code logs the error:

```
log.Printf("Failed to open /etc/passwd: %v", err)
```

but does not return, so execution reaches:

```
defer passwdFile.Close()
```

In Go, calling a method on a nil interface value produces a panic because the method dispatch requires a non-nil receiver. The `Close()` method is never called because the function panics first.

## Impact

- The OCI hook process crashes with a nil pointer dereference instead of gracefully handling the error
- Container creation operations that depend on the hook fail
- The crash occurs during a container lifecycle hook, which can leave containers in an inconsistent state

## Fix

Added `return err` after the error log so the function exits cleanly when `/etc/passwd` cannot be opened. This ensures:

1. The error is properly propagated to the caller
2. The nil `passwdFile` is never used for deferred close or scanning
3. The hook fails gracefully instead of crashing

Fixes #2693

### DIFF
--- a/KubeArmor/cert/cert.go
+++ b/KubeArmor/cert/cert.go
@@ -208,7 +208,7 @@ func GenerateCA(cfg *CertConfig) (*CertBytes, error) {
 	}
 	crtBytes, err := GenerateSelfSignedCert(crtTemp, cfg)
 	if err != nil {
-		return &CertBytes{}, nil
+		return &CertBytes{}, err
 	}
 	return &CertBytes{
 		Crt: crtBytes.Crt,
--- a/KubeArmor/cert/cert_test.go
+++ b/KubeArmor/cert/cert_test.go
@@ -1,0 +1,81 @@
+package cert
+
+import (
+	"crypto/x509"
+	"testing"
+)
+
+func TestGenerateCA_Success(t *testing.T) {
+	cfg := &CertConfig{
+		CN:           "test-ca",
+		Organization: "test-org",
+		IsCa:         true,
+		KeyUsage:     x509.KeyUsageCertSign,
+	}
+
+	certBytes, err := GenerateCA(cfg)
+	if err != nil {
+		t.Fatalf("GenerateCA() expected no error, got: %v", err)
+	}
+	if len(certBytes.Crt) == 0 {
+		t.Error("GenerateCA() returned empty certificate")
+	}
+	if len(certBytes.Key) == 0 {
+		t.Error("GenerateCA() returned empty key")
+	}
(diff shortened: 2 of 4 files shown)


Aamod007:

vee1e:
@Aamod007 i think you've pushed 2 unrelated fixes into one PR. the fixes from: 
- https://github.com/kubearmor/KubeArmor/pull/2692
- https://github.com/kubearmor/KubeArmor/pull/2690

are both applied here too.

PR #2720, opened 2026-06-29
Fix(KA002 + KA001(part-B) ) : Implmenting mTLS on by default and restructured grpc architecture (control palne and data palne)

sagarkhandagre998:
```
┌─────────────────────┬──────────────────────────┬──────────────────────────────────────┬──────────────────────────┐
│ Mode                │ Log Server (observability)│ Management Server                   │ Notes                    │
├─────────────────────┼──────────────────────────┼──────────────────────────────────────┼──────────────────────────┤
│ K8s                 │ Created on port 32767     │ Not created                         │ Policy via K8s CRDs      │
│ (K8sEnabled=true)   │ with TLS                 │ — no port 32768                     │ CA from operator/helm    │
│                     │ CA: /.../tls/log/ca.crt  │ No attack surface                   │ at /.../tls/log/ca.crt   │
├─────────────────────┼──────────────────────────┼──────────────────────────────────────┼──────────────────────────┤
│ VM without TLS      │ Created on port 32767    │ Created on port 32768               │ karmor vm/probe use      │
│ (TLSEnabled=false)  │ without TLS              │ without TLS                         │ insecure credentials     │
├─────────────────────┼──────────────────────────┼──────────────────────────────────────┼──────────────────────────┤
│ VM with TLS         │ Created on port 32767    │ Created on port 32768               │ Requires CA from         │
│ (TLSEnabled=true)   │ with TLS                 │ with TLS                            │ karmor vm init           │
│                     │ CA: /.../tls/log/ca.crt  │ CA: /.../tls/management/ca.crt      │                          │
└─────────────────────┴──────────────────────────┴──────────────────────────────────────┴──────────────────────────┘

```


sagarkhandagre998:
hey @Aryan-sharma11,  Please take a look . Design docs is shared on ticket .

sagarkhandagre998:
I have a question regarding the new TLS setup. Since the client will now connect over TLS on a new port instead of the existing log server port, do we also need to update the client-side callers? 

Specifically, should we modify kubearmorController, or are there any other components/clients that currently depend on the existing log server port and need to be updated to use the new TLS endpoint? 

I want to make sure I don't miss any client that communicates with the log server or new server.

sagarkhandagre998:
The current implementation expects `ca.crt and ca.key` to exist before `CreateTlsServerCredentials()` is called. This works when certificates are provisioned by the operator/daemonset, but CI and local development fail because those files are absent. Is the expectation that CI provisions these certificates before starting the daemon, or should the cert package bootstrap a development CA automatically when it doesn't find one?

sagarkhandagre998:
@Aryan-sharma11 
 
 While implementing the mTLS support, we ran into an issue with local development and CI.

The current `self` certificate provider expects an existing CA (`ca.crt`/`ca.key`) to already be present under the configured TLS directory. This works in Kubernetes where the operator/DaemonSet provisions the certificates, but in local development and CI the daemon failed to start because those files don't exist.

To address this, we introduced a new **`dev`** certificate provider intended only for development and CI environments.

The current behavior is:

* `tlsCertProvider=self` → Existing behavior (expects a pre-existing CA).
* `tlsCertProvider=external` → Existing behavior.
* `tlsCertProvider=dev` → On first startup, if `ca.crt`/`ca.key` are missing, automatically generate a development CA, persist it under the configured TLS directory, and then use that CA to issue the server certificate.

With this approach, the server starts successfully in local environments without requiring manual certificate bootstrapping, while the production providers (`self` and `external`) remain unchanged.

**I've verified that In My Local Setup:**

* The `dev` provider is selected correctly.
* The development CA is generated automatically.
* KubeArmor starts successfully with TLS enabled.
* The new `ManagementServer` is created successfully and listens on its dedicated port.

Before we continue with client certificate generation for full mTLS, we'd like to confirm whether this is the direction you'd prefer.

Does introducing a dedicated `dev` certificate provider for automatic development/CI bootstrapping align with the project's intended certificate management model, or would you prefer a different approach for handling local and CI environments?

sagarkhandagre998:
mTLS ports Design decision is pending .

sagarkhandagre998:
Now We are implimenting the new design restructured architecture for observability and management services.

sagarkhandagre998:
@Aryan-sharma11 Its ready for review.

sagarkhandagre998:
@Aryan-sharma11 I will start working on karmor after this will be approved.

sagarkhandagre998:
@Aryan-sharma11 Do we have any update ?

sagarkhandagre998:
**Backward-compatibility audit: observability/management plane separation + mTLS-on-by-default**

P0 - tlsEnabled default flipped false→true (config.go)

Any deployment not explicitly passing -tlsEnabled=false (non-K8s/systemd, bare-metal, custom manifests, older operator) now serves TLS + requires client certs (tls.RequireAndVerifyClientCert). Existing plaintext clients break.
Compounding issue: default provider is still self, which doesn't bootstrap certs — it reads a pre-existing /var/lib/kubearmor/tls/ca.crt. A fresh non-K8s install with defaults crash-loops at startup. CI only passes because systemd tests explicitly set -tlsCertProvider=dev.

P1 - Management plane moved off :32767 (non-K8s only)

PolicyService/ProbeService/StateAgent now on :32765/UDS. In-repo test clients updated, but the external karmor CLI still defaults to :32767 → VM-mode policy push/probe will connection-refuse until kubearmor-client changes land in lockstep.

P3 - Management health server never reports SERVING

grpc_health_v1 is registered on the management server, but nothing sets a status on it; SetHealthStatus only touches the :32766 plaintext health server. Probes against the management plane would falsely report unhealthy.

P4 - Reflection removed from :32767 - in K8s mode it's gone entirely (management server doesn't run there). grpcurl/introspection breaks; intended hardening but needs documenting.

P5 - Cert SANs expanded - KubeArmorServerSANs adds localhost/host/namespace names. Net-compatible, but external/management deployments need separate certs under <tlsCertPath>/management.

Unchanged (good): :32766 health port behavior, mTLS semantics itself, observability :32767 LogService + keepalive, helm tls.enabled=false default.

sagarkhandagre998:
I think we need to serve Management health server , WDYT @Aryan-sharma11 ? 

Also share your thoughts on the above issues.

KubeArmor/config/config.go
```diff
@@ -146,10 +156,13 @@ func readCmdLineParams() {
 	hostStr := flag.String(ConfigHost, hostname, "host name")
 
 	grpcStr := flag.String(ConfigGRPC, "32767", "gRPC port number")
+	managementGRPCStr := flag.String(ConfigManagementGRPC, "32768", "management gRPC port number")
```
Aryan-sharma11, line 159: let's use 32765,32766 , 32767. 
```suggestion
	managementGRPCStr := flag.String(ConfigManagementGRPC, "32765", "management gRPC port number")
```

KubeArmor/core/kubeArmor.go
```diff
@@ -672,7 +711,7 @@ func KubeArmor() {
 		}
 		dm.Logger.Print("Initialized State Agent Server")
 
-		pb.RegisterStateAgentServer(dm.Logger.LogServer, dm.StateAgent)
+		pb.RegisterStateAgentServer(dm.ManagementServer.Server, dm.StateAgent)
```
Aryan-sharma11, line 714: now since management server will be conditionally initialized, let's add a `nil` check here as well.

sagarkhandagre998:
@Aryan-sharma11 Can we try to do the review this week , It will be appreciated if someone else can take over .

.github/workflows/ci-test-systemd.yml
```diff
@@ -60,6 +60,39 @@ jobs:
...
 
+      - name: Configure KubeArmor mTLS for systemd tests
+        run: |
+          sudo systemctl stop kubearmor
+          sudo install -d -m 0750 /var/lib/kubearmor/tls
+          sudo install -d /etc/systemd/system/kubearmor.service.d
+          sudo tee /etc/systemd/system/kubearmor.service.d/ci-mtls.conf >/dev/null <<'EOF'
+          [Service]
+          ExecStart=
+          ExecStart=/opt/kubearmor/kubearmor -tlsCertProvider=dev -tlsCertPath=/var/lib/kubearmor/tls
```
Aryan-sharma11, line 71:
```suggestion
          ExecStart=/opt/kubearmor/kubearmor -tlsCertProvider=dev -tlsCertPath=/var/lib/kubearmor/tls
```

KubeArmor/grpcutil/grpcutil.go
Aryan-sharma11, line 1: Let's fix lint warnings as well

KubeArmor/management/management.go
Aryan-sharma11, line 1: Let's fix the lint warnings here.

sagarkhandagre998:
@Aryan-sharma11 I'll fix as soon as possible, Thanks for your time.

.github/workflows/ci-test-systemd.yml
```diff
@@ -54,6 +54,7 @@ jobs:
           sudo install -d /etc/systemd/system/kubearmor.service.d
           sudo tee /etc/systemd/system/kubearmor.service.d/ci-mtls.conf >/dev/null <<'EOF'
           [Service]
+          ExecStart=
```
sagarkhandagre998, line 57: @Aryan-sharma11 
I think this is important to keep it because without it, systemctl restart fails. Here's the mechanics:

- The base unit from the deb (/usr/lib/systemd/system/kubearmor.service) already has ExecStart=/opt/kubearmor/kubearmor.
- our drop-in adds a second ExecStart= with the TLS flags. Drop-ins merge with the base unit - they don't replace it.
- systemd allows only one ExecStart= for normal services. Two of them = "bad unit file setting", restart refused.
- The empty ExecStart= line means "clear the inherited one first" - it's the documented reset pattern (man systemd.unit, overriding vendor settings). After it, your flagged line becomes the single effective command.

sagarkhandagre998 (commented):
Please take a look

PR #2721, opened 2026-06-30
fix(feeder): rate-limit drop logs to prevent memory spikes

AryanBakliwal:
**Purpose of PR?**:
During high-event loads (e.g., thousands of events per second) where connected clients cannot consume the stream fast enough, the Feeder's internal client broadcast channels fill up. The previous implementation executed

https://github.com/kubearmor/KubeArmor/blob/27ded4a9cfa8aae52177cfc7e3a3b9e3fcb7802e/KubeArmor/feeder/feeder.go#L926

for every single dropped event.
Spamming the standard output logger thousands of times per second resulted in the KubeArmor's memory usage spike exponentially.

This PR adds a counter to track dropped logs, printing a single warning for every 10,000 dropped logs to prevent memory spikes.

Fixes #

No

image

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/KubeArmor/feeder/feeder.go
+++ b/KubeArmor/feeder/feeder.go
@@ -219,6 +219,9 @@ type BaseFeeder struct {
 	// True if feeder and its workers are working
 	Running bool
 
+	// Number of dropped logs
+	DroppedLogs uint64
+
 	// LogServer //
 
 	// port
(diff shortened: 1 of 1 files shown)


KubeArmor/feeder/feeder.go
Aryan-sharma11, line 924: What if the channel recovers before reaching the 10,000-drop threshold? Should we wait until the counter reaches 10,000 the next time logs are dropped? wdyt?
```suggestion
			case fd.EventStructs.LogStructs[uid].Broadcast <- pbLog:
		if fd.DroppedLogs > 0 {
			kg.Warnf("Dropped %d logs.", fd.DroppedLogs)
			fd.DroppedLogs = 0
		}
```

achrefbensaad (approved):

KubeArmor/feeder/feeder.go
AryanBakliwal: @Aryan-sharma11 I tested this change, although this did not increase memory usage %, it caused a lot of extra logs like below
image

Aryan-sharma11 (approved):

nyrahul (approved):
LGTM

PR #2742, opened 2026-07-07
fix(ci): fix ci-merge-coverage workflow failing with git rev-parse error

bhuvan-somisetty:
## Summary
- `ci-merge-coverage` was failing on every run on `main` (e.g. run for 106f8de) with `fatal: not a git repository (or any of the parent directories): .git` and exit code 128.
- Root cause: the "Check if all required workflows completed successfully" step calls `git rev-parse HEAD` before the `actions/checkout` step has run, so no git repo exists yet in the runner workspace.
- Fix: move `actions/checkout` and `actions/setup-go` before the check step, and use `github.event.workflow_run.head_sha` instead of `git rev-parse HEAD` (this workflow is triggered by `workflow_run`, so that's the correct commit reference anyway).
- Also sanitized the per-workflow status env var name (`ci-test-ginkgo_status` -> `ci_test_ginkgo_status`, since hyphenated env context keys are unreliable) and switched to `GH_TOKEN` for `gh` CLI auth.

## Test plan
- [x] Reviewed the failing run logs on `main` confirming `git rev-parse HEAD` fails pre-checkout
- [x] YAML validated locally
- [ ] Verify `ci-merge-coverage` passes on the next `ci-test-ginkgo` completion after merge

### DIFF
--- a/.github/workflows/ci-merge-coverage.yaml
+++ b/.github/workflows/ci-merge-coverage.yaml
@@ -17,16 +17,27 @@ jobs:
       actions: read
       contents: read
     steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          submodules: true
+
+      - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version-file: "KubeArmor/go.mod"
+
       - name: Check if all required workflows completed successfully
         id: check-workflows
         run: |
           workflows=("ci-test-ginkgo")
           all_completed=true
 
-          commit_sha=$(git rev-parse HEAD)
+          commit_sha="${{ github.event.workflow_run.head_sha }}"
 
           for workflow in "${workflows[@]}"; do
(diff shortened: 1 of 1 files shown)


bhuvan-somisetty:
This fixes the \`ci-merge-coverage\` failures on \`main\` (e.g. run for 106f8de) — the "Check if all required workflows completed successfully" step ran \`git rev-parse HEAD\` before \`actions/checkout\`, so there was no git repo yet and it failed with \`fatal: not a git repository\` (exit 128). Moved checkout/setup-go earlier and switched to \`github.event.workflow_run.head_sha\`, which is the correct commit reference for a \`workflow_run\`-triggered job anyway. Also sanitized the per-workflow env var name and switched to \`GH_TOKEN\` for \`gh\` CLI auth.

Note: \`security/snyk (KubeArmor)\` is failing with "You have used your limit of private tests" — that's a Snyk org billing/quota limit unrelated to this change, and it's not a required check.

@Aryan-sharma11 @AryanBakliwal @achrefbensaad requesting review, thanks!

PR #2749, opened 2026-07-07
fix(policy): validate nil policy requests

aaa-aashna:
## Overview

Adds validation to policy gRPC handlers to safely handle nil or empty requests and prevent potential panics.

## Changes

* Validate inputs in:

  * `ContainerPolicy`
  * `HostPolicy`
  * `NetworkPolicy`
* Return `PolicyStatus_Invalid` for invalid requests
* Add regression test for nil policy requests

## Why

Improves robustness by ensuring malformed inputs are handled safely without affecting valid request behavior.

## Testing

```bash
go test ./policy
```

All existing policy tests pass, and the new regression test validates the added input validation.


aaa-aashna:
@Aryan-sharma11 @AryanBakliwal PTAL

PR #2775, opened 2026-07-15
fix(state): guard StateEventChans iteration with StateEventChansLock

### LINKED ISSUE #2774: state agent: Push*Event iterate StateEventChans without StateEventChansLock — daemon crash (concurrent map write / send on closed channel)
## Bug Report

**General Information**
- Environment description: any environment running KubeArmor with the state agent enabled (`-stateAgent` / `gRPC` state service), e.g. non-k8s / systemd mode or wherever `WatchState` clients connect
- Kernel version: independent of kernel
- Orchestration system version: independent of orchestrator
- KubeArmor version: `main` (also present in v1.7)

**To Reproduce**
1. Run KubeArmor with the state agent enabled.
2. Have a `WatchState` gRPC client connect/disconnect repeatedly (e.g. `karmor` probes, kubearmor-relay reconnects, or any client churn).
3. Generate container events at the same time (start/stop containers).
4. The daemon eventually crashes with `fatal error: concurrent map iteration and map write` or `panic: send on closed channel`.

A deterministic reproducer as a Go test (fails under `go test -race` and even panics without `-race`) is below.

**Expected behavior**
Broadcasting state events must be safe against clients connecting/disconnecting concurrently; a client disconnect must never crash the daemon.

## Details

`PushContainerEvent`, `PushNodeEvent` and `PushNamespaceEvent` in `KubeArmor/state/data.go` iterate `sa.StateEventChans` **without holding `StateEventChansLock`**:

```go
// KubeArmor/state/data.go (PushContainerEvent, same pattern in the other two)
if sa.StateEventChans == nil {
        return
}

for uid, conn := range sa.StateEventChans {   // <-- no StateEventChansLock held
        select {
        case conn <- containerEvent:
        ...
```

Meanwhile `addStateEventChan` / `removeStateEventChan` (called on every `WatchState` client connect/disconnect in `KubeArmor/state/stateAgent.go`) mutate the same map — and `removeStateEventChan` also **closes the channel** — under the lock:

```go
// KubeArmor/state/stateAgent.go
func (sa *StateAgent) removeStateEventChan(uid string) {
        sa.StateEventChansLock.Lock()
        close(sa.StateEventChans[uid])   // Push* may be sending on this right now
        delete(sa.StateEventChans, uid)  // Push* may be iterating right now
        sa.StateEventChansLock.Unlock()
}
```

The push functions are invoked from dedicated goroutines by the runtime handlers (`go dm.StateAgent.PushContainerEvent(...)` in `core/containerdHandler.go`, `core/dockerHandler.go`, `core/hook_handlier_non_k8s.go`), so the map read and the map write/close genuinely run concurrently. Two failure modes, both fatal for the whole daemon:

1. `fatal error: concurrent map iteration and map write` — map mutated while the broadcast loop iterates it (this one is not recoverable even with a `recover()`).
2. `panic: send on closed channel` — `removeStateEventChan` closes a channel while a `Push*Event` goroutine is selected on sending to it.

## Reproducer

```go
func TestPushContainerEventRace(t *testing.T) {
	sa := NewStateAgent(&tp.Node{NodeName: "n"}, new(sync.RWMutex), map[string]tp.Container{}, new(sync.RWMutex))

	var wg sync.WaitGroup
	stop := make(chan struct{})

	wg.Add(1)
	go func() { // WatchState clients connecting/disconnecting
		defer wg.Done()
		for {
			select {
			case <-stop:
				return
			default:
			}
			uid, conn := sa.addStateEventChan()
			select {
			case <-conn:
			default:
			}
			sa.removeStateEventChan(uid)
		}
	}()

	wg.Add(1)
	go func() { // container events from runtime handlers
		defer wg.Done()
		for i := 0; i < 5000; i++ {
			sa.PushContainerEvent(tp.Container{ContainerID: fmt.Sprintf("c%d", i), ContainerName: "nginx", NamespaceName: "default"}, EventAdded)
		}
		close(stop)
	}()

	wg.Wait()
}
```

Observed output on current `main`:

```
WARNING: DATA RACE
Read at 0x00c0002c93c0 by goroutine 11:
  ...state.(*StateAgent).PushContainerEvent()
      KubeArmor/state/data.go:85
Previous write at 0x00c0002c93c0 by goroutine 10:
  runtime.closechan()
  ...state.(*StateAgent).removeStateEventChan()
      KubeArmor/state/stateAgent.go:92
...
panic: send on closed channel
goroutine 23 [running]:
...state.(*StateAgent).PushContainerEvent(...)
	KubeArmor/state/data.go:85
```

## Suggested fix

Take `StateEventChansLock.RLock()` around the broadcast loop in all three `Push*Event` functions. The sends are already non-blocking (`select`/`default`), so holding the read lock during fan-out is cheap, and since `removeStateEventChan` closes channels under the write lock, the read lock also makes the send-on-closed-channel impossible.

I have a fix with a regression test ready and will open a PR.

Aamod007:
**Purpose of PR?**:

Fixes #2774

`PushContainerEvent`, `PushNodeEvent` and `PushNamespaceEvent` in `KubeArmor/state/data.go` iterate `sa.StateEventChans` without holding `StateEventChansLock`, while `addStateEventChan`/`removeStateEventChan` mutate the map and close the channels under that lock on every `WatchState` client connect/disconnect. Since the runtime handlers push events from separate goroutines (`go dm.StateAgent.PushContainerEvent(...)`), a client disconnecting mid-broadcast crashes the daemon with `fatal error: concurrent map iteration and map write` or `panic: send on closed channel`.

This PR takes `StateEventChansLock.RLock()` around the broadcast loop in all three functions. The sends are already non-blocking (`select`/`default`), so holding the read lock during fan-out is cheap, and because `removeStateEventChan` closes channels under the write lock, the read lock also rules out the send-on-closed-channel panic.

No.

- Added `TestPushContainerEventRace` which simulates `WatchState` clients connecting/disconnecting concurrently with container event pushes. On current `main` it fails under `go test -race` (data race between `PushContainerEvent` and `removeStateEventChan`) and also panics with `send on closed channel`. With this fix, `go test -race -count=1 -run TestPushContainerEventRace ./state/` passes.
- `go vet ./state/` clean.

Failure output on `main` without the fix:

```
WARNING: DATA RACE
Read at ... by goroutine 11:
  ...state.(*StateAgent).PushContainerEvent()  KubeArmor/state/data.go:85
Previous write at ... by goroutine 10:
  runtime.closechan()
  ...state.(*StateAgent).removeStateEventChan() KubeArmor/state/stateAgent.go:92
...
panic: send on closed channel
```

**Checklist:**
- [x] Bug fix. Fixes #2774
- [x] Commit has unit tests

### DIFF
--- a/KubeArmor/state/data_race_test.go
+++ b/KubeArmor/state/data_race_test.go
@@ -1,0 +1,66 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Authors of KubeArmor
+
+package state
+
+import (
+	"fmt"
+	"sync"
+	"testing"
+
+	tp "github.com/kubearmor/KubeArmor/KubeArmor/types"
+)
+
+func newTestStateAgent() *StateAgent {
+	node := &tp.Node{NodeName: "test-node"}
+	return NewStateAgent(node, new(sync.RWMutex), map[string]tp.Container{}, new(sync.RWMutex))
+}
+
+// TestPushContainerEventRace demonstrates the unsynchronized read of
+// sa.StateEventChans in PushContainerEvent racing with clients
+// connecting/disconnecting (addStateEventChan/removeStateEventChan).
+// Run with -race to observe the data race.
+func TestPushContainerEventRace(t *testing.T) {
+	sa := newTestStateAgent()
+
+	var wg sync.WaitGroup
+	stop := make(chan struct{})
+
+	// simulate WatchState clients connecting and disconnecting
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for i := 0; ; i++ {
+			select {
+			case <-stop:
+				return
+			default:
+			}
+			uid, conn := sa.addStateEventChan()
+			// drain a bit like WatchState would
+			select {
+			case <-conn:
+			default:
+			}
+			sa.removeStateEventChan(uid)
+			_ = i
+		}
+	}()
+
+	// simulate container events pushed by the runtime handlers
+	// (containerdHandler/dockerHandler call these via `go ...`)
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for i := 0; i < 5000; i++ {
+			sa.PushContainerEvent(tp.Container{
+				ContainerID:   fmt.Sprintf("c%d", i),
+				ContainerName: "nginx",
+				NamespaceName: "default",
+			}, EventAdded)
+		}
+		close(stop)
+	}()
+
+	wg.Wait()
+}
--- a/KubeArmor/state/data_race_test.go
+++ b/KubeArmor/state/data_race_test.go
@@ -30,7 +30,7 @@ func TestPushContainerEventRace(t *testing.T) {
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		for i := 0; ; i++ {
+		for {
 			select {
 			case <-stop:
 				return
@@ -43,7 +43,6 @@ func TestPushContainerEventRace(t *testing.T) {
 			default:
 			}
 			sa.removeStateEventChan(uid)
-			_ = i
 		}
 	}()
 
--- a/KubeArmor/state/data.go
+++ b/KubeArmor/state/data.go
@@ -75,6 +75,9 @@ func (sa *StateAgent) PushContainerEvent(container tp.Container, event string) {
 		Object: containerBytes,
 	}
 
+	sa.StateEventChansLock.RLock()
+	defer sa.StateEventChansLock.RUnlock()
+
 	// skip sending message as no state receiver is connected
 	if sa.StateEventChans == nil {
 		return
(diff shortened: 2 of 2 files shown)


KubeArmor/state/data_race_test.go
```diff
@@ -0,0 +1,66 @@
...
+			default:
+			}
+			uid, conn := sa.addStateEventChan()
+			// drain a bit like WatchState would
+			select {
+			case <-conn:
+			default:
+			}
+			sa.removeStateEventChan(uid)
+			_ = i
```
vee1e: here `i` is already used in the for loop, so the compiler doesn't consider it unused, making the assignment pointless. It neither suppresses a compiler error nor does anything at runtime. i know its pedantic but still worth pointing out dead code.

Aamod007: Good catch — the loop variable wasn't used at all, so I removed both the counter and the `_ = i` line and made it a plain `for` loop. Fixed in 2bdeb7d6.

Aamod007:
@vee1e i implemented the changes, please look at pr in your free hours, Thank you

PR #2781, opened 2026-07-19
fix(core): resolve AppArmor profile leak on non-k8s container destroy

### LINKED ISSUE #2780: bug : AppArmor profile leak on container destroy in non-k8s (systemd) mode
### Bug Summary

In non-Kubernetes (systemd) mode, destroying a container fails to unregister and unload its
AppArmor profile, leaving orphaned profile files in `/etc/apparmor.d/` and still loaded in the
kernel after the container is gone.

### Root Cause

There are two related issues in `KubeArmor/core/hook_handlier_non_k8s.go`, inside
`UpdateContainer()` under `action == "destroy"`:

**1. Ordering bug (in-memory cleanup never runs)**

`dm.MatchandRemoveContainerFromEndpoint(containerID)` is called *before* the loop that looks up
and removes the matching entry from `endPoint.AppArmorProfiles`. That lookup depends on
`kl.ContainsElement(endPoint.Containers, container.ContainerID)` — but `containerID` has already
been stripped out of `endPoint.Containers` by the earlier call. So this check always evaluates to
`false`, and the intended cleanup loop (written with the comment `// remove apparmor profile for
that endpoint`) never actually removes anything. It's dead code in its current position.

**2. Missing enforcer call (kernel/disk profile never unloaded)**

Separately, even if issue 1 is fixed, `hook_handlier_non_k8s.go` never calls
`dm.RuntimeEnforcer.UpdateAppArmorProfiles(..., "DELETED", ...)` on container destroy — unlike
k8s mode, where `KubeArmor/core/kubeUpdate.go` calls this on pod deletion to trigger
`apparmor_parser -R` and remove the profile file. Without this call, the profile stays loaded in
the kernel and the file stays on disk indefinitely.

### Observable Effect

After a container is destroyed in non-k8s mode, its AppArmor profile file
(`/etc/apparmor.d/kubearmor_<container_name>`) and the corresponding kernel-loaded policy persist
until the KubeArmor daemon itself restarts — at which point `NewAppArmorEnforcer()`'s startup scan
sweeps up unused profiles. On a long-running host with frequent container churn, this leaves an
accumulating number of orphaned profiles between restarts.

**Note on severity:** this is a resource/cleanup leak, not a security bypass — the orphaned
profile continues enforcing whatever rules it already had; it's just not removed promptly when it
should be.

### Environment

- KubeArmor version: v1.7.5-rc1
- Mode: systemd (non-k8s), Docker runtime
- LSM: AppArmor (forced via `-lsm=apparmor`; also reproduces under default `bpf,apparmor,selinux`
  order once AppArmor is the active enforcer)

### Suggested Fix

Reorder the destroy logic so the AppArmor profile lookup runs *before*
`MatchandRemoveContainerFromEndpoint`, and add the missing
`dm.RuntimeEnforcer.UpdateAppArmorProfiles(container.ContainerName, "DELETED", ...)` call, mirroring
the k8s-mode behavior in `kubeUpdate.go`. Happy to open a PR with this fix plus a regression test.

VirajMandavkar:
**Purpose of PR?**:
This PR resolves an AppArmor profile leak occurring during container destruction in non-Kubernetes (systemd / standalone CRI) mode (KubeArmor/core/hook_handlier_non_k8s.go).

*Root Causes & Fixes:*
1. **Ordering Defect in Memory Cleanup:** In UpdateContainer (action == "destroy"), dm.MatchandRemoveContainerFromEndpoint(containerID) was executed at line 218 before line 226's loop checked kl.ContainsElement(endPoint.Containers, container.ContainerID) to remove the profile. Because line 218 had already stripped containerID from endPoint.Containers, ContainsElement always evaluated to false, causing endPoint.AppArmorProfiles to retain stale profile entries in memory.
    - Fix: Reordered the AppArmorProfiles lookup loop to run prior to MatchandRemoveContainerFromEndpoint.
    
2. **Missing Runtime Enforcer Unregister Call:** The non-K8s destroy path never invoked dm.RuntimeEnforcer.UpdateAppArmorProfiles(...), causing /etc/apparmor.d/kubearmor_<container_name> profile files and kernel-enforced rules to persist indefinitely after container termination.
    - Fix: Added dm.RuntimeEnforcer.UpdateAppArmorProfiles(container.ContainerName, "DELETED", appArmorProfiles, privilegedProfiles) to unregister and unload profiles on container destroy.

Fixes #2780 

 No 

1. **Unit Test Verification:** Added unit test TestUpdateContainerNonK8sDestroyAppArmorCleanup in KubeArmor/core/hook_handlier_non_k8s_test.go to test container create -> destroy lifecycle state transitions. Verified endPoint.AppArmorProfiles receives the profile on container creation and is cleanly purged on destroy.
2. **Code Quality Checks:** Ran go vet ./core/... (passed with 0 warnings) and go test -v ./core/ -run TestUpdateContainerNonK8sDestroyAppArmorCleanup (PASS).
3. **Endpoint State Verification**: Verified that kl.ContainsElement(endPoint.Containers, container.ContainerID) evaluates to true now that profile removal executes before container ID removal.

This PR addresses issue #2780. The fix mirrors the profile deletion pattern used in Kubernetes mode (kubeUpdate.go:929) adapted for non-K8s container descriptors.

**Checklist:**
- [x] Bug fix. Fixes #2780 
- [x] Commit has unit tests

### DIFF
--- a/KubeArmor/core/hook_handlier_non_k8s.go
+++ b/KubeArmor/core/hook_handlier_non_k8s.go
@@ -215,13 +215,7 @@ func (dm *KubeArmorDaemon) UpdateContainer(containerID string, container tp.Cont
 			return fmt.Errorf("container not found for removal: %s", containerID)
 		}
 		dm.EndPointsLock.Lock()
-		dm.MatchandRemoveContainerFromEndpoint(containerID)
-		dm.EndPointsLock.Unlock()
-		delete(dm.Containers, containerID)
-		dm.ContainersLock.Unlock()
-
-		dm.EndPointsLock.Lock()
-		// remove apparmor profile for that endpoint
+		// remove apparmor profile for that endpoint BEFORE removing container ID from endpoint
 		for idx, endPoint := range dm.EndPoints {
 			if endPoint.NamespaceName == container.NamespaceName && endPoint.EndPointName == container.EndPointName && kl.ContainsElement(endPoint.Containers, container.ContainerID) {
 
@@ -236,7 +230,20 @@ func (dm *KubeArmorDaemon) UpdateContainer(containerID string, container tp.Cont
 				break
 			}
 		}
+		dm.MatchandRemoveContainerFromEndpoint(containerID)
 		dm.EndPointsLock.Unlock()
+		delete(dm.Containers, containerID)
+		dm.ContainersLock.Unlock()
+		if dm.RuntimeEnforcer != nil && container.AppArmorProfile != "" {
+			appArmorProfiles := map[string]string{
+				container.ContainerName: container.AppArmorProfile,
+			}
+			privilegedProfiles := map[string]struct{}{}
+			if container.Privileged {
+				privilegedProfiles[container.AppArmorProfile] = struct{}{}
+			}
+			dm.RuntimeEnforcer.UpdateAppArmorProfiles(container.ContainerName, "DELETED", appArmorProfiles, privilegedProfiles)
+		}
 		// delete endpoint if no security rules and containers
 		idx := 0
 		endpointsLength := len(dm.EndPoints)
--- a/KubeArmor/core/hook_handlier_non_k8s.go
+++ b/KubeArmor/core/hook_handlier_non_k8s.go
@@ -217,7 +217,7 @@ func (dm *KubeArmorDaemon) UpdateContainer(containerID string, container tp.Cont
 		dm.EndPointsLock.Lock()
 		// remove apparmor profile for that endpoint BEFORE removing container ID from endpoint
 		for idx, endPoint := range dm.EndPoints {
-			if endPoint.NamespaceName == container.NamespaceName && endPoint.EndPointName == container.EndPointName && kl.ContainsElement(endPoint.Containers, container.ContainerID) {
+			if endPoint.NamespaceName == container.NamespaceName && endPoint.EndPointName == container.ContainerName && kl.ContainsElement(endPoint.Containers, container.ContainerID) {
 
 				// update apparmor profiles
 				for idxA, profile := range endPoint.AppArmorProfiles {
--- a/KubeArmor/core/hook_handlier_non_k8s_test.go
+++ b/KubeArmor/core/hook_handlier_non_k8s_test.go
@@ -1,0 +1,90 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Authors of KubeArmor
+
+package core
+
+import (
+	"sync"
+	"testing"
+
+	cfg "github.com/kubearmor/KubeArmor/KubeArmor/config"
+	fd "github.com/kubearmor/KubeArmor/KubeArmor/feeder"
+	mon "github.com/kubearmor/KubeArmor/KubeArmor/monitor"
+	tp "github.com/kubearmor/KubeArmor/KubeArmor/types"
+)
+
+func TestUpdateContainerNonK8sDestroyAppArmorCleanup(t *testing.T) {
+	// Disable policy and state agent to isolate the container update state machine
+	cfg.GlobalCfg.Policy = false
+	cfg.GlobalCfg.StateAgent = false
+
+	dm := &KubeArmorDaemon{
+		EndPoints:            []tp.EndPoint{},
+		EndPointsLock:        &sync.RWMutex{},
+		ContainersLock:       &sync.RWMutex{},
+		Containers:           make(map[string]tp.Container),
+		SecurityPoliciesLock: &sync.RWMutex{},
+		SecurityPolicies:     []tp.SecurityPolicy{},
+		SystemMonitor: &mon.SystemMonitor{
+			NsMap:            make(map[mon.NsKey]string),
+			NsMapLock:        &sync.RWMutex{},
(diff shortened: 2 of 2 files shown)


KubeArmor/core/hook_handlier_non_k8s.go
```diff
@@ -236,7 +230,20 @@ func (dm *KubeArmorDaemon) UpdateContainer(containerID string, container tp.Cont
...
+		dm.ContainersLock.Unlock()
+		if dm.RuntimeEnforcer != nil && container.AppArmorProfile != "" {
+			appArmorProfiles := map[string]string{
+				container.ContainerName: container.AppArmorProfile,
+			}
+			privilegedProfiles := map[string]struct{}{}
+			if container.Privileged {
+				privilegedProfiles[container.AppArmorProfile] = struct{}{}
+			}
+			dm.RuntimeEnforcer.UpdateAppArmorProfiles(container.ContainerName, "DELETED", appArmorProfiles, privilegedProfiles)
```
vee1e, line 245: here `UpdateAppArmorProfiles` only drops in-memory pod associations, but it doesn't unload apparmor.d or kernel profiles as claimed.

with a non-empty first argument it only removes that name from the in-memory pod list and returns. full unload only happens when `podName == ""` (enforcer destroy).

KubeArmor/core/hook_handlier_non_k8s_test.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	// Verify container removed from container map
+	dm.ContainersLock.RLock()
+	if _, ok := dm.Containers[containerID]; ok {
+		t.Errorf("container %s still exists in dm.Containers map", containerID)
+	}
+	dm.ContainersLock.RUnlock()
+
+	// Verify AppArmor profile entry was removed from the endpoint
```
vee1e: here the regression test is ineffective. after destroy, the endpoint is removed completely when it has zero containers and zero security policies, so the post-destroy code finds nothing whether or not `AppArmorProfiles` was cleaned up first. this assertion would pass on the pre-fix code as well.

few more things:
- the fixture never sets `EndPointName`. create stores the container as-is, and destroy matches `endPoint.EndPointName == container.EndPointName`. so the profile-removal loop never matches in this test.
- `RuntimeEnforcer` is never set (and policy is off), so the new `UpdateAppArmorProfiles` path is never exercised

VirajMandavkar:
@vee1e thanks , you're right on both points.

on the test: fixing -- adding a SecurityPolicy to the fixture so the endpoint survives container destroy (previously it was pruned to zero which made the assertion pass vacuously). now asserting on a surviving endpoint with zero AppArmorProfiles remaining.

on UpdateAppArmorProfiles: confirmed you're right that non-empty podName only updates the in-memory pod-tracking list, not the kernel/disk profile. i traced the k8s-mode call (kubeUpdate.go:929) and it has the same behavior - non-empty podName never falls through to the apparmor_parser / file-delete branch there either. so this PR brings non-k8s parity with k8s's existing behavior for this call. neither mode triggers immediate kernel/disk unload on
a single container/pod destroy - only on enforcer shutdown (DestroyAppArmorEnforcer passes
podName="") and orphan cleanup happens at daemon startup (NewAppArmorEnforcer GC scan).

is that intentional (e.g. profile kept warm for reuse if same container name restarts) or is
there a separate cleanup path i'm missing? want to make sure i'm not scoping this PR to fix
parity when there's a deeper gap that should be its own issue.

vee1e:
@VirajMandavkar sounds good re: test. keeping a `SecurityPolicy` so the endpoint survives destroy, then asserting it still exists with an empty `AppArmorProfiles` list is exactly what that check needed. That should make the ordering fix real instead of vacuous.

on `UpdateAppArmorProfiles` i agree as well. good catch on the k8s path. if non-k8s now matches k8s  then this PR is the right scope: parity + the ordering bug, not a full disk/kernel unload redesign.

i don’t think you’re missing a separate per-destroy unload path in the current design. profiles can be shared/reused, so full unload on every destroy would be a bigger behavior change. if we want true last-user unload (or always drop orphans without waiting for restart), that should be its own issue. not a blocker for this PR.

once the test fix is in, I’m good on these two comments.

VirajMandavkar:
> @VirajMandavkar sounds good re: test. keeping a `SecurityPolicy` so the endpoint survives destroy, then asserting it still exists with an empty `AppArmorProfiles` list is exactly what that check needed. That should make the ordering fix real instead of vacuous.
> 
> on `UpdateAppArmorProfiles` i agree as well. good catch on the k8s path. if non-k8s now matches k8s then this PR is the right scope: parity + the ordering bug, not a full disk/kernel unload redesign.
> 
> i don’t think you’re missing a separate per-destroy unload path in the current design. profiles can be shared/reused, so full unload on every destroy would be a bigger behavior change. if we want true last-user unload (or always drop orphans without waiting for restart), that should be its own issue. not a blocker for this PR.
> 
> once the test fix is in, I’m good on these two comments.

@vee1e  yess I'll be modifying the unit test and might also add some test verifying the edge cases . It might take a while
Thank You!

VirajMandavkar:
@vee1e  also if time allows can you take a look at my other pr's 
#2770 #2761

PR #2802, opened 2026-07-24
cert: fix GenerateCA returning nil error on self-signed cert failure

### LINKED ISSUE #2689: cert/cert.go: GenerateCA returns nil error on self-signed cert generation failure

bhuvan-somisetty:
**Purpose of PR?**:

In the `GenerateCA` function in `KubeArmor/cert/cert.go`, when `GenerateSelfSignedCert` fails, the error was logged and swallowed, returning an empty `CertBytes` struct with a `nil` error value (`return &CertBytes{}, nil`). Calling code was thus unable to detect certificate generation failures and proceeded with uninitialized or empty TLS certificates.

This change ensures that `GenerateCA` propagates the error returned by `GenerateSelfSignedCert`.

Fixes #2689

No

1. Added unit test `TestGenerateCA_Success` verifying CA certificate generation under normal conditions.
2. Added unit test `TestGenerateCA_ErrorPropagationOnSelfSignedCertFailure` verifying that errors during self-signed certificate creation are properly returned to the caller.

**Checklist:**
- [x] Bug fix. Fixes #2689
- [x] Commit has unit tests

### DIFF
--- a/KubeArmor/cert/cert.go
+++ b/KubeArmor/cert/cert.go
@@ -208,7 +208,7 @@ func GenerateCA(cfg *CertConfig) (*CertBytes, error) {
 	}
 	crtBytes, err := GenerateSelfSignedCert(crtTemp, cfg)
 	if err != nil {
-		return &CertBytes{}, nil
+		return &CertBytes{}, err
 	}
 	return &CertBytes{
 		Crt: crtBytes.Crt,
@@ -246,6 +246,10 @@ func GenerateCert(cfg *CertConfig) (*CertKeyPair, error) {
 
 // GenerateSelfSignedCert func generates cert and key signed by provided CA
 func GenerateSelfSignedCert(ca *CertKeyPair, cfg *CertConfig) (*CertBytes, error) {
+	if ca == nil || ca.Crt == nil || ca.Key == nil {
+		return nil, fmt.Errorf("invalid CA certificate or key")
+	}
+
 	certKeyPair, err := GenerateCert(cfg)
 	if err != nil {
 		return nil, err
--- a/KubeArmor/cert/cert.go
+++ b/KubeArmor/cert/cert.go
@@ -208,6 +208,7 @@ func GenerateCA(cfg *CertConfig) (*CertBytes, error) {
 	}
 	crtBytes, err := GenerateSelfSignedCert(crtTemp, cfg)
 	if err != nil {
+		klog.Errorf("error generating self-signed ca cert: %s\n", err)
 		return &CertBytes{}, err
 	}
 	return &CertBytes{
--- a/KubeArmor/cert/cert_test.go
+++ b/KubeArmor/cert/cert_test.go
@@ -1,0 +1,64 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Authors of KubeArmor
+
+package cert
+
+import (
+	"testing"
+	"time"
+)
+
+func TestGenerateCA_Success(t *testing.T) {
+	cfg := &DefaultKubeArmorCAConfig
+	cfg.NotAfter = time.Now().Add(24 * time.Hour)
+
+	caBytes, err := GenerateCA(cfg)
+	if err != nil {
+		t.Fatalf("expected no error generating CA, got: %v", err)
+	}
+
+	if len(caBytes.Crt) == 0 {
+		t.Errorf("expected non-empty CA certificate bytes")
+	}
+
+	if len(caBytes.Key) == 0 {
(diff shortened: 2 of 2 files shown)


bhuvan-somisetty:
Hi @Aryan-sharma11 @achrefbensaad @AryanBakliwal @daemon1024 @nyrahul — could someone please take a look at this PR when you get a chance? It fixes #2689 by propagating errors from GenerateSelfSignedCert in GenerateCA() rather than swallowing them. Also added unit tests in cert_test.go. Thanks!

KubeArmor/cert/cert.go
```diff
@@ -208,7 +208,7 @@ func GenerateCA(cfg *CertConfig) (*CertBytes, error) {
 	}
 	crtBytes, err := GenerateSelfSignedCert(crtTemp, cfg)
 	if err != nil {
-		return &CertBytes{}, nil
+		return &CertBytes{}, err
```
vee1e, line 212: this path now correctly propagates `err`, but returns it silently. For consistency with the existing style in this function and the rest of the file, log here too:

```go
if err != nil {
    klog.Errorf("error generating self-signed ca cert: %s\n", err)
    return &CertBytes{}, err
}
```

KubeArmor/cert/cert_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+package cert
+
+import (
+	"testing"
+	"time"
+)
+
+func TestGenerateCA_Success(t *testing.T) {
+	cfg := &DefaultKubeArmorCAConfig
```
vee1e: this mutates a package-level config var, permanently changing `DefaultKubeArmorCAConfig` for the rest of the process, which can affect later tests if more coverage is added.

Take a value copy instead (same pattern as production callers in `pkg/KubeArmorOperator/cert/certgen.go`):

```go
cfg := DefaultKubeArmorCAConfig
```

KubeArmor/cert/cert.go
bhuvan-somisetty: Good catch @vee1e, fixed in c547c447. Added the `klog.Errorf("error generating self-signed ca cert: %s\n", err)` before the return so this path logs like the rest of the function instead of failing silently.

KubeArmor/cert/cert_test.go
bhuvan-somisetty: Makes sense @vee1e, fixed in c547c447. Switched to a value copy (`cfg := DefaultKubeArmorCAConfig`) in both tests, so `DefaultKubeArmorCAConfig` is no longer mutated for the rest of the process. `go test ./cert/...` still passes.

AryanBakliwal (approved):

Aryan-sharma11 (approved):

Issue #2683, opened 2026-06-21
Ran KubeArmor's system_monitor.bpf.o across a kernel matrix — compatibility evidence (useful as a CI lane?)


ErenAri:
Hi KubeArmor team,

I maintain [bpfcompat](https://github.com/Kernel-Guard/bpfcompat) (Apache-2.0): it boots real distro kernels in disposable QEMU/KVM VMs, load/attach-validates a compiled `.bpf.o`, and reports a per-kernel pass/fail matrix with classified failure reasons (missing BTF, CO-RE relocation, unsupported map/program/attach type).

I built `system_monitor.bpf.o` from the repo (`make -C KubeArmor/BPF`) and ran it through bpfcompat across 5 kernels (Ubuntu 5.4 / 5.15, Debian 6.1, Ubuntu 6.8, AlmaLinux 8 / 4.18). Two findings — one a loader contract, then a real matrix once I honored it.

**1. As-is, a generic libbpf load fails identically on every kernel** — including 6.8 with BTF — at:

> Map `kubearmor_visibility` failed to create: `Invalid argument (-22)`

Failing the *same way even on the newest kernel with BTF* means it's not a kernel issue: `kubearmor_visibility` is a `BPF_MAP_TYPE_HASH_OF_MAPS` whose inner per-namespace map your Go (cilium/ebpf) loader installs at load time. A naive libbpf load has no inner-map prototype, so map creation fails. That's a loader contract, not a compatibility result.

**2. Honoring the contract, here's the real matrix.** I declared the inner-map prototype in a small bpfcompat manifest (key `struct outer_key`, inner `scope → u32`):

```yaml
name: kubearmor-system-monitor
maps:
  - name: kubearmor_visibility
    inner_map:
      type: hash
      key_size: 4
      value_size: 4
      max_entries: 64
```

With that, `system_monitor.bpf.o` (sha256 `7fda0e91…d285c1`) **loads on all five kernels**:

| Kernel | Host kernel | BTF | Load | Attach (best-effort) |
|---|---|---|---|---|
| Ubuntu 20.04 (5.4) | `5.4.0-216` | yes | **pass** | 52/55 |
| Ubuntu 22.04 (5.15) | `5.15.0-181` | yes | **pass** | 52/55 |
| Debian 12 (6.1) | `6.1.0-49` | yes | **pass** | 52/55 |
| Ubuntu 24.04 (6.8) | `6.8.0-117` | yes | **pass** | 52/55 |
[...]

PR #2809, opened 2026-07-26
ci: non-blocking build-on-target eBPF kernel-compatibility lane (system_monitor)

ErenAri:
This is the small, non-blocking proof-of-concept lane I offered in #2683.

It answers one narrow question on a schedule: does `system_monitor.bpf.o` still
build and load across a set of real vendor distro kernels? For each kernel it
boots a disposable VM, installs the toolchain, builds `system_monitor.bpf.o`
against **that kernel's own BTF** (matching KubeArmor's per-node build model -
the BPF Makefile builds with CO-RE access-index preservation off), then loads it
with the `kubearmor_visibility` `HASH_OF_MAPS` inner map installed as the loader
would, and produces a classified JSON/Markdown report.

Green run on GitHub-hosted runners (my fork), all four kernels:
https://github.com/ErenAri/KubeArmor/actions/runs/30216558482
- almalinux-8 (4.18, RHEL-family backport)
- ubuntu-22.04 (5.15)
- debian-12 (6.1)
- ubuntu-24.04 (6.8)

Scope and honesty, following our discussion in #2683:
- **Complementary to lvh, not a replacement.** lvh provisions kernels for full
  KubeArmor functional/E2E tests; this narrowly checks that the BPF object still
  builds+loads on vendor kernels and classifies *why* if it doesn't.
- **Non-blocking**: schedule + `workflow_dispatch` only, no PR/push trigger.
- **Loader caveat**: the load uses bpfcompat's libbpf-based validator, whereas
  KubeArmor's runtime loader is cilium/ebpf. A libbpf load-pass is a strong
  signal but not identical. A natural follow-up is to swap in a small
  KubeArmor-owned cilium/ebpf self-test via the same command mode (verdict =
  exit code), which would exercise the real loader path.
- **Ubuntu 20.04 (5.4) is intentionally omitted**: focal's bundled bpftool
  predates raw-BTF dump, so `bpftool btf dump file /sys/kernel/btf/vmlinux
  format c` fails there - KubeArmor's own build hits the same on focal.

Disclosure: I maintain bpfcompat. Happy to adjust the kernel set or cadence,
rework it as the cilium/ebpf self-test variant, or drop it - whatever is most
useful. No obligation either way.

### DIFF
--- a/.github/bpfcompat/kernel-matrix.yaml
+++ b/.github/bpfcompat/kernel-matrix.yaml
@@ -1,0 +1,15 @@
+# Kernels for the build-on-target BPF compatibility lane. Each is built against
+# its OWN kernel BTF inside the guest, matching KubeArmor's per-node build model.
+# (Ubuntu 20.04/5.4 is intentionally omitted: focal's bundled bpftool predates
+# raw-BTF dump, so `bpftool btf dump file /sys/kernel/btf/vmlinux` fails there -
+# KubeArmor's own build hits the same on focal.)
+name: kubearmor-buildontarget
+profiles:
+  - id: almalinux-8-4.18   # RHEL family backports many features onto 4.18
+    required: true
+  - id: ubuntu-22.04-5.15
+    required: true
+  - id: debian-12-6.1
+    required: true
+  - id: ubuntu-24.04-6.8
+    required: true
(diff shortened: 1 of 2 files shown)


ErenAri:
hello can you look at it @parisnakitakejser @legorie @asharma0703 @asifalix

PR #2814, opened 2026-07-29
fix(core): guard strings.Split index in AppArmor annotation parsing

### LINKED ISSUE #2813: core: unguarded strings.Split index in AppArmor annotation parsing can panic the daemon
## Bug Report

**General Information**

- Component: `KubeArmor/core/kubeUpdate.go` (pod update handler, AppArmor annotation parsing)
- Applies to: any Kubernetes deployment mode with the AppArmor enforcer enabled
- Kernel version: N/A (pure Go logic bug, not enforcer/kernel specific)

**Description**

`UpdateContainer` (and the equivalent path for pods) parses each pod annotation whose key starts with `container.apparmor.security.beta.kubernetes.io` in order to build the `appArmorAnnotations` map:

```go
for k, v := range pod.Annotations {
    if strings.HasPrefix(k, "container.apparmor.security.beta.kubernetes.io") {
        if v == "unconfined" {
            containerName := strings.Split(k, "/")[1]
            appArmorAnnotations[containerName] = v
        } else {
            containerName := strings.Split(k, "/")[1]
            appArmorAnnotations[containerName] = strings.Split(v, "/")[1]
        }
    }
}
```
(https://github.com/kubearmor/KubeArmor/blob/main/KubeArmor/core/kubeUpdate.go#L843-L853)

Both branches index into the result of `strings.Split` without checking its length:

- `strings.Split(k, "/")[1]` assumes the annotation key always contains a `/` separating the prefix from the container name.
- `strings.Split(v, "/")[1]` assumes the annotation value is always in the `localhost/<profile>` form.

`strings.Split` returns a 1-element slice when the separator is absent, so either line panics with `index out of range [1] with length 1` the moment a pod annotation doesn't match the expected shape — e.g. the bare key `container.apparmor.security.beta.kubernetes.io` with no container suffix, or a value that isn't `unconfined` and doesn't contain a `/` (a typo, a non-`localhost/` profile reference, or anything set by a mutating webhook/user that doesn't follow the expected convention). Since this loop runs on every pod add/update event, a single non-conforming annotation on any pod in the cluster can crash the KubeArmor daemon goroutine handling pod updates.

**To Reproduce**

1. Deploy KubeArmor with the AppArmor enforcer enabled on a cluster.
2. Create a pod with a non-conforming AppArmor annotation, e.g.:
   ```yaml
   metadata:
     annotations:
       container.apparmor.security.beta.kubernetes.io/nginx: "not-a-valid-profile-string"
   ```
   (any value without a `/`, and not equal to `unconfined`)
3. Observe the KubeArmor daemon logs / pod status when the pod is created or updated.

**Expected behavior**

KubeArmor should treat a malformed or unexpected `container.apparmor.security.beta.kubernetes.io/*` annotation as a non-fatal case (skip it and optionally log a warning), not panic. Malformed annotations can originate from user error, third-party mutating webhooks, or future Kubernetes AppArmor annotation format changes, and should never be able to crash the daemon.

**Suggested fix**

Guard both `strings.Split` results with a length check (`len(parts) == 2`) before indexing, consistent with defensive patterns already used elsewhere in this file, and skip/log the annotation instead of indexing when the shape doesn't match.

bhuvan-somisetty:
**Purpose of PR?**:

`UpdateContainer`'s AppArmor annotation loop indexed into `strings.Split(k, "/")[1]` and `strings.Split(v, "/")[1]` without checking the length of the returned slice. Any pod annotation key without a `/`-separated container name, or a non-`unconfined` value without a `/`-separated profile, made `strings.Split` return a 1-element slice and panicked the goroutine handling pod updates.

This pulls the parsing logic into `parseAppArmorAnnotations`, guards both splits with a length check, and skips + logs a warning for a malformed annotation instead of indexing out of range.

Fixes #2813

No.

- Added `TestParseAppArmorAnnotations` covering: well-formed `unconfined` annotation, well-formed `localhost/<profile>` annotation, a bare annotation key with no container suffix, a value without a `/` separator, unrelated annotations, and nil annotations.
- `go build ./...` and `go test ./...` pass locally.

**Checklist:**
- [x] Bug fix. Fixes #2813
- [x] Commit has unit tests

### DIFF
--- a/KubeArmor/core/kubeUpdate.go
+++ b/KubeArmor/core/kubeUpdate.go
@@ -604,6 +604,43 @@ func (dm *KubeArmorDaemon) HandleUnknownNamespaceNsMap(container *tp.Container) 
 	dm.SystemMonitor.NsMapLock.Unlock()
 }
 
+// parseAppArmorAnnotations extracts container-to-profile mappings from a pod's
+// "container.apparmor.security.beta.kubernetes.io/*" annotations. Malformed
+// annotation keys (missing the "/<containerName>" suffix) or values (missing
+// the "localhost/<profile>" form) are skipped and reported back as warnings
+// instead of causing an index-out-of-range panic.
+func parseAppArmorAnnotations(annotations map[string]string) (map[string]string, []string) {
+	result := map[string]string{}
+	var warnings []string
+
+	for k, v := range annotations {
+		if !strings.HasPrefix(k, "container.apparmor.security.beta.kubernetes.io") {
+			continue
+		}
+
+		keyParts := strings.Split(k, "/")
+		if len(keyParts) != 2 {
+			warnings = append(warnings, fmt.Sprintf("Skipping malformed AppArmor annotation key (%s)", k))
+			continue
+		}
+		containerName := keyParts[1]
+
+		if v == "unconfined" {
+			result[containerName] = v
+			continue
+		}
+
+		valueParts := strings.Split(v, "/")
+		if len(valueParts) != 2 {
+			warnings = append(warnings, fmt.Sprintf("Skipping malformed AppArmor annotation value (%s: %s)", k, v))
+			continue
+		}
+		result[containerName] = valueParts[1]
+	}
+
+	return result, warnings
+}
+
 func (dm *KubeArmorDaemon) handlePodEvent(event string, obj *corev1.Pod) {
 	if event != addEvent && event != updateEvent && event != deleteEvent {
 		return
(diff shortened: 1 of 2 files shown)


bhuvan-somisetty:
Hi @achrefbensaad @Aryan-sharma11 @AryanBakliwal, would appreciate a review on this whenever you get a chance. It fixes a daemon panic in the AppArmor annotation parsing path (issue #2813). Thanks!

PR #2868, opened 2026-09-02
fix(deploy): mount host /etc/machine-id for correct node_id reporting

### LINKED ISSUE #2846: [Docker] KubeArmor reports incorrect machine_id in docker based onboarding
## Bug Report

**General Information**

  - VM (ubuntu 22.04.4 LTS)
  - `Linux worker 5.15.0-187-generic #197-Ubuntu SMP Fri Jul 17 19:17:01 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux` 
- Orchestration system version in use (e.g. `kubectl version`, ...) NA
- Link to relevant artifacts (policies, deployments scripts, ...) NA
- Target containers/pods NA

**To Reproduce**

1. Onboard a VM in docker mode (using `knoxctl onboard vm <remaining-command>`)
2. Check logs of kubearmor container

**Expected behavior**

While starting KubeArmor reports the node_name, node_ip, node_id (machine_id), os image, and kernel version. All of these information are correct and a valid machine_id/node_id is printed (verified using hostnamectl command) when installing in systemd mode. In docker mode the node_name itself is shown as the machine_id/node_id.

This issue is also reproducible in K8s installation as the KubeArmor is running as a container in the K8s environment. 

A description of what you expected to happen.

In both systemd and docker based installation KubeArmor should show the correct machine_id value

**Screenshots**

Image

pujitha24:
**Purpose of PR?**:

KubeArmor reports the wrong `node_id`/`machine_id` when it is not run directly on the
host (Docker/Podman deployment, and any Kubernetes DaemonSet installed via the Helm
chart or the KubeArmorOperator). `KubeArmor/core/kubeArmor.go` reads `/etc/machine-id`
(falling back to `/var/lib/dbus/machine-id`, then to the node name) to populate
`Node.NodeID`, but none of the container-based deployment manifests mounted the host's
`/etc/machine-id` into the KubeArmor container. Inside the container that path either
doesn't exist or belongs to the container's own filesystem, so both reads fail and the
code silently falls back to using the node name as the machine ID — exactly the symptom
described in the issue (systemd installs are unaffected because KubeArmor runs directly
on the host there).

Fixes #

No. This only adds a new read-only hostPath mount of `/etc/machine-id` to existing
deployment manifests; no application code, CLI flags, or existing behavior changes.

This is a manifest-only fix (no cluster/docker available in the sandbox this was
prepared in), so validation was limited to static/template-level checks that mirror
what this repo's own CI runs for these paths:
- `helm lint ./deployments/helm/KubeArmor` and `./deployments/helm/KubeArmorOperator` — pass.
- `.github/workflows/helm-validate-values.sh` (the exact script `ci-test-helm-charts.yml` runs) — passes for docker/crio/k3s/microk8s/minikube/GKE/BottleRocket/EKS/generic.
- Rendered `helm template ... -s templates/daemonset.yaml --set environment.name=<env>` for all 11 supported environment values and confirmed the new `/etc/machine-id` volume + mount is present on the main `kubearmor` container for every environment except `GKE Autopilot` (excluded intentionally, see below), and is absent from the `init` container.
- `deployments/podman/docker-compose.yaml` validated as syntactically-correct YAML (`python3 -c "import yaml; yaml.safe_load(...)"`).
- `cd pkg/KubeArmorOperator && go build ./common/... ./internal/controller/...`, `go vet` on the same packages, and `go test ./common/... ./internal/controller/...` all pass; `gofmt -l` reports no issues on the changed Go file.
- Could not run this repo's live-cluster CI job (`ci-test-controllers.yml`, which installs the operator on a real k3s cluster) — no docker/k3s available here. That job also doesn't assert on `node_id`/`machine-id` values either way, so it wouldn't have specifically exercised this fix even if it were run.

GKE Autopilot (`deployments/helm/KubeArmor`, `environment.name=autopilot`) is deliberately
excluded from the new mount: Autopilot only permits a fixed, Google-controlled allowlist
of hostPath mounts for allowlisted system pods (this chart already opts KubeArmor into
that allowlist via `cloud.google.com/matching-allowlist`), so adding an arbitrary new
hostPath there risks breaking Autopilot installs outright.

Two of the three DaemonSet-generation code paths in this repo are updated:
`deployments/helm/KubeArmor` (used directly, and for GKE Autopilot) and the
KubeArmorOperator (`pkg/KubeArmorOperator/common/defaults.go`, the default path behind
`karmor install` / `helm install kubearmor-operator`). The third, legacy path
(`deployments/get/objects.go`, used only by an opt-in `--legacy` installer flag in the
separate `kubearmor-client` repo and marked deprecated in favor of `karmor install`/Helm
in this repo's development guide) is intentionally left out of scope here to keep this
change small; happy to follow up separately if maintainers would like it covered too.

**Checklist:**
- [ ] Commit has unit tests

Fixes #2846

### DIFF
--- a/deployments/helm/KubeArmor/templates/daemonset.yaml
+++ b/deployments/helm/KubeArmor/templates/daemonset.yaml
@@ -76,6 +76,11 @@ spec:
         terminationMessagePolicy: File
         volumeMounts:
         {{- toYaml .Values.kubearmor.commonMounts | trim | nindent 10 }}
+        {{- if ne .Values.environment.name "autopilot" }}
+          - mountPath: /etc/machine-id
+            name: machine-id
+            readOnly: true
+        {{- end }}
         {{- if .Values.tls.enabled -}}
           {{- toYaml .Values.kubearmor.tls.kubearmorCACertVolumeMount | trim | nindent 10 }}
         {{- end -}}
@@ -148,6 +153,12 @@ spec:
       terminationGracePeriodSeconds: 30
       volumes:
       {{- toYaml .Values.kubearmor.commonVolumes | trim | nindent 8}}
+      {{- if ne .Values.environment.name "autopilot" }}
+        - hostPath:
+            path: /etc/machine-id
+            type: File
+          name: machine-id
+      {{- end }}
       {{- if .Values.tls.enabled -}}
         {{- toYaml .Values.kubearmor.tls.kubearmorCACertVolume | trim | nindent 8 }}
       {{- end -}}
--- a/deployments/podman/docker-compose.yaml
+++ b/deployments/podman/docker-compose.yaml
@@ -59,6 +59,7 @@ services:
       - "/sys/kernel/security:/sys/kernel/security"
       - "/sys/kernel/debug:/sys/kernel/debug"
       - "/etc/apparmor.d:/etc/apparmor.d"
+      - "/etc/machine-id:/etc/machine-id:ro"
       - "/usr/share/kubearmor:/usr/share/kubearmor:rw"
       - "/var/run/kubearmor:/var/run/kubearmor:rw"
     restart: always
--- a/pkg/KubeArmorOperator/common/defaults.go
+++ b/pkg/KubeArmorOperator/common/defaults.go
@@ -357,6 +357,15 @@ var CommonVolumes = []corev1.Volume{
 			},
 		},
 	},
+	{
+		Name: "machine-id",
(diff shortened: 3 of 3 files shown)


vishnusomank (commented):
This assumes `/etc/machine-id` always exists. However, on some systems, the machine ID may be located at `/var/lib/dbus/machine-id`.

Can we check that scenario as well?

PR #2885, opened 2026-09-09
fix(setup): install llvm 12 from apt.llvm.org on bionic

### LINKED ISSUE #2884: Setup script on Ubuntu 18.04
## Bug Report
The /KubeArmor/contribution/self-managed-k8s/setup.sh script is not able to install LLVM and Clang on Ubuntu 18.04 after recent changes from #2706 
```bash
$ sudo apt-get install -y clang-12 llvm-12

Reading package lists... Done
Building dependency tree
Reading state information... Done
E: Unable to locate package clang-12
E: Unable to locate package llvm-12
```

**General Information**

- Environment description (GKE, VM-Kubeadm, vagrant-dev-env, minikube, microk8s, ...) - Ubuntu 18.04
- Kernel version (run `uname -a`) - 5.4.0

**To Reproduce**

Run the /KubeArmor/contribution/self-managed-k8s/setup.sh script on Ubuntu 18.04 machine

**Expected behavior**

The script should install LLVM and Clang

magic-peach:
**Purpose of PR?**:

Fixes #2884

No.

Verified the script logic by hand for the bionic and focal branches (no real Ubuntu 18.04 machine available in this environment): bionic now bootstraps LLVM 12 via the LLVM project's own apt.llvm.org repo before installing, since bionic's own archive never shipped clang/llvm past version 9. Focal already has clang-12/llvm-12 in its own archive and its code path is untouched.

Ubuntu 18.04's official apt archive tops out at clang-9, so `apt-get install -y clang-12 llvm-12` always fails there with package not found, per the reproduction in #2884. The same `clang-12`/`llvm-12` assumption for bionic also exists in `.github/workflows/install-llvm.sh`, but that path is not exercised by this repo's own CI (GitHub Actions runners use jammy/noble), so it was left alone to keep this change scoped to the reported, user-facing bug.

**Checklist:**
- [x] Bug fix. Fixes #2884
- [ ] Commit has unit tests


contribution/self-managed-k8s/setup.sh
```diff
@@ -25,6 +25,13 @@ sudo apt-get -y install build-essential libelf-dev pkg-config net-tools linux-he
...
 sudo apt-get -y install build-essential libelf-dev pkg-config
 if [ "$VERSION_CODENAME" == "focal" ] || [ "$VERSION_CODENAME" == "bionic" ]; then
+    if [ "$VERSION_CODENAME" == "bionic" ]; then
+        # bionic's own apt archive never shipped clang/llvm past version 9,
+        # so pull 12 from the LLVM project's own repo instead
+        wget -O /tmp/build/llvm.sh https://apt.llvm.org/llvm.sh
+        chmod +x /tmp/build/llvm.sh
+        sudo /tmp/build/llvm.sh 12
+    fi
     sudo apt-get install -y clang-12 llvm-12
```
Aryan-sharma11, line 35: does this needs to be in else block then?

magic-peach: No, llvm.sh 12 on its own only installs clang-12/lldb-12/lld-12/clangd-12 (its default PKG list unless you pass all), it does not install the llvm-12 package itself. The tools this script relies on further down (llc, llvm-strip, opt, llvm-dis) come from llvm-12, not clang-12, so the apt-get install line below still needs to run unconditionally to pull that in. Re-installing clang-12 there is a harmless no-op since llvm.sh already installed it.

PR #2900, opened 2026-09-11
fix(networkpolicy): resolve container/host network policy matching outside k8s

### LINKED ISSUE #2891: On VM mode, network policies with bandwidth quota doesn't update nftables
## Bug Report

**General Information**
```
❯ uname -a
Linux e2e-runner 6.18.28-1-lts #1 SMP PREEMPT_DYNAMIC Fri, 08 May 2026 07:04:32 +0000 x86_64 GNU/Linux
```

```
❯ cat /etc/os-release 
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo
```

**To Reproduce**

1. In a VM with the above spec, deploy kubearmor in Systemd mode with this image `docker.io/kubearmor/kubearmor-systemd:1.7.4_linux-amd64`
2. Apply a network policy with a bandwidth quota. e.g., we tried the policies below targeting both the node and the containers running inside the VM.

Policy for the host
```
kind: KubeArmorNetworkPolicy
metadata:
  name: host-egress-quota
spec:
  nodeSelector:
    matchLabels:
      kubearmor.io/hostname: e2e-runner
  action: Audit
  severity: 1
  egress:
  - limit: "4MB"
    duration: "1m"
```
Policy for the nginx container
```apiVersion: security.kubearmor.com/v1
kind: KubeArmorNetworkPolicy
metadata:
  name: quota-test-policy
spec:
  selector:
    matchLabels:
      kubearmor.io/container.name: nginx
  action: Block
  level: Policy # <-- Shared across all selected pods
  egress:
  - limit: "2MB" 
    duration: "20m"
```
None of the policies updated the nftables

**Expected behavior**
Applying all policies should update nftables and enforce the quota.

Ironankit525:
Fixes #2891

**Purpose of PR?**:

Two separate root causes were making `KubeArmorNetworkPolicy` (including bandwidth quotas) silently do nothing outside Kubernetes (Systemd/VM mode):

1. `EndPoint.PodIP` was only ever populated by the K8s pod-watch code. Outside k8s, a pod/container-selector policy had no IP to match against, so `generateRules()` returned zero rules with no error. Fixed by resolving the container's IP via its network namespace (`common.GetContainerIPFromPid`) for containerd/CRI-O (Docker already exposed this through its own inspect API), and backfilling `EndPoint.PodIP` from it in `NetworkPolicyEnforcer`. Policy evaluation also re-runs on container add/destroy, not just on policy change, so nftables stays correct when a container appears after its policy was already applied.

2. `dm.Node.Identities` (used to match `nodeSelector` on any host-scoped policy) was only populated outside k8s when `HostPolicy` was enabled — not when only `NetworkPolicyEnforcer` was on. A host-level `nodeSelector` policy would silently never match in that case. Fixed by including `NetworkPolicyEnforcer` in that gate.

No. K8s-mode behavior is unchanged — a real K8s pod IP still takes precedence over the container-derived one, and that's covered by a test.

- Unit tests for endpoint-IP resolution and the quota lifecycle: create/reuse/delete across container and policy changes, IPv4/IPv6, shared vs. per-pod quotas, and container add/remove triggering re-evaluation.
- A test that creates a real, isolated Linux network namespace with a known non-loopback address and verifies `GetContainerIPFromPid` resolves it correctly and restores the original namespace afterward (not mocked).
- **Not verified here**: an actual containerd/CRI-O daemon, or a live Systemd/VM KubeArmor deployment applying these policies against real nftables and real traffic — i.e. the exact repro in #2891. I don't have that environment available.

Split into two commits, one per root cause: the container-IP resolution fix (the bulk of the change) and the smaller host-identity fix.

`unix.Setns` is the standard technique for entering a container's network namespace (CNI plugins do the same), but it's new to this codebase, so `common.GetContainerIPFromPid` is the piece worth the closest read. It only runs when `NetworkPolicyEnforcer` is enabled, since that's its only consumer.

If someone has a VM handy, confirming the original repro now works — both policies, on whichever runtime that VM uses — would be a big help.

**Checklist:**
- [x] Bug fix. Fixes #2891
- [x] Commit has unit tests
- [ ] Commit has integration tests — a gated nft/traffic test is included at `KubeArmor/networkPolicyEnforcer/nftables_integration_test.go` for manual runs in a disposable namespace, but nothing here exercises a live containerd/CRI-O/systemd deployment

### DIFF
--- a/KubeArmor/common/common.go
+++ b/KubeArmor/common/common.go
@@ -16,6 +16,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"regexp"
+	"runtime"
 	"slices"
 	"strconv"
 	"strings"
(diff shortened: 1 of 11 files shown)


Ironankit525:
Since I couldn't reproduce the original environment (containerd/CRI-O + live nftables in a Systemd/VM deployment), could you or someone verify:

1. The original repro from #2891 — both the `nginx` container quota and the `host-egress-quota` policy — actually resolves against a real deployment.
2. Whichever CRI that environment uses: if it's `containerd` or `CRI-O`, confirm `GetContainerIPFromPid` actually resolves a container's real IP there. The netns-switching approach is new to this codebase.
3. Whether the existing CI matrix (`ci-test-ginkgo` / `ci-test-controllers`, which run against containerd and crio) is enough to catch a regression here, or whether this needs a dedicated `nonk8s_env` test added separately.
4. Any perf or safety concerns with entering a container's network namespace from the daemon's own process on each container add (it only runs when `NetworkPolicyEnforcer` is enabled). Happy to change the approach if there's a lighter-weight option already used elsewhere in KubeArmor that I missed.

Aryan-sharma11:
> Since I couldn't reproduce the original environment (containerd/CRI-O + live nftables in a Systemd/VM deployment), could you or someone verify:

We were using kubearmor in systemd and applied policy on nginx container in docker.
> 
> 1. The original repro from [On VM mode, network policies with bandwidth quota doesn't update nftables #2891](https://github.com/kubearmor/KubeArmor/issues/2891) — both the `nginx` container quota and the `host-egress-quota` policy — actually resolves against a real deployment.
> 2. Whichever CRI that environment uses: if it's `containerd` or `CRI-O`, confirm `GetContainerIPFromPid` actually resolves a container's real IP there. The netns-switching approach is new to this codebase.

kubearmor currently doesn't support cri-o and containerd in vm mode, only docker is supported/

> 3. Whether the existing CI matrix (`ci-test-ginkgo` / `ci-test-controllers`, which run against containerd and crio) is enough to catch a regression here, or whether this needs a dedicated `nonk8s_env` test added separately.

Currently, there are no ginkgo tests that test quota policies.

> 4. Any perf or safety concerns with entering a container's network namespace from the daemon's own process on each container add (it only runs when `NetworkPolicyEnforcer` is enabled). Happy to change the approach if there's a lighter-weight option already used elsewhere in KubeArmor that I missed.

Can you explain more on this?

